PR #30, opened 2024-11-04
Automate installation of overmind and use `bin/dev` for continuity

mjgiarlo:
Per https://github.com/DarthSim/overmind/blob/master/packaging/rubygems/README.md#installation-with-rails


justinlittman (approved):

justinlittman (approved):

justinlittman (approved):
Would you mind adding a note to README pointing out this tweak?

We often make these helpful tweaks, but forget about them when creating new applications.

justinlittman:
Actually, to be honest, I don't love these sorts of changes as they deviate from the gem's README. While they do reduce some keystrokes for us, they do so at the expense of possible future confusion.

PR #45, opened 2024-11-06
Implement capistrano deployment

### LINKED ISSUE #10: Implement application deployment
Add the necessary capistrano gems and configure capistrano for application deployment to our servers.

**Done when**:
- Capistrano is configured following our normal process.
- sdr-deploy?

aaron-collier:
Closes #10 

See 
- https://sul-h3-qa.stanford.edu/ (Not found because we have no root controller), https://sul-h3-qa.stanford.edu/up (shows that the application is responding)
- https://sul-h3-stage.stanford.edu/ (Not found because we have no root controller), https://sul-h3-stage.stanford.edu/up (shows that the application is responding)

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# Load DSL and set up stages
+require 'capistrano/setup'
+
+# Include default deployment tasks
+require 'capistrano/deploy'
+
+require 'capistrano/scm/git'
+install_plugin Capistrano::SCM::Git
+
+require 'capistrano/bundler'
+require 'capistrano/rails'
+require 'capistrano/honeybadger'
+require 'capistrano/passenger'
+require 'capistrano/maintenance'
+require 'dlss/capistrano'
+
+# Load custom tasks from `lib/capistrano/tasks` if you have any defined
+Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
--- a/Capfile
+++ b/Capfile
@@ -10,10 +10,9 @@ require 'capistrano/scm/git'
 install_plugin Capistrano::SCM::Git
 
 require 'capistrano/bundler'
-require 'capistrano/rails'
 require 'capistrano/honeybadger'
 require 'capistrano/passenger'
-require 'capistrano/maintenance'
+require 'capistrano/rails/migrations'
 require 'dlss/capistrano'
 
 # Load custom tasks from `lib/capistrano/tasks` if you have any defined
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,0 +1,2 @@
+//= link_tree ../images
+//= link_tree ../builds
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,2 +1,0 @@
-//= link_tree ../images
-//= link_tree ../builds
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -12,6 +12,6 @@ production:
   adapter: solid_cable
   connects_to:
     database:
-      writing: cable
+      writing: production
   polling_interval: 0.1.seconds
   message_retention: 1.day
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -12,6 +12,6 @@ production:
   adapter: solid_cable
   connects_to:
     database:
-      writing: production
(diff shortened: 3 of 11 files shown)


config/cable.yml
```diff
@@ -12,6 +12,6 @@ production:
   adapter: solid_cable
   connects_to:
     database:
-      writing: cable
+      writing: production
```
aaron-collier, line 15: Setting this to production to allow the deploy to complete and the application start. This should be changed once we determine the multiple database stratagy.

config/cache.yml
```diff
@@ -12,5 +12,5 @@ test:
   <<: *default
 
 production:
-  database: cache
+  database: production
```
aaron-collier, line 15: same as above

config/environments/production.rb
```diff
@@ -51,8 +51,7 @@
 
   # Replace the default in-process and non-durable queuing backend for Active Job.
   config.active_job.queue_adapter = :solid_queue
-  config.solid_queue.connects_to = { database: { writing: :queue } }
-
+  config.solid_queue.connects_to = { database: { writing: :production } }
```
aaron-collier, line 54: same as above

app/assets/config/manifest.js
```diff
@@ -0,0 +1,2 @@
+//= link_tree ../images
```
justinlittman, line 1: This file shouldn't be necessary, I believe?

Capfile
```diff
@@ -0,0 +1,20 @@
...
+require 'capistrano/deploy'
+
+require 'capistrano/scm/git'
+install_plugin Capistrano::SCM::Git
+
+require 'capistrano/bundler'
+require 'capistrano/rails'
+require 'capistrano/honeybadger'
+require 'capistrano/passenger'
+require 'capistrano/maintenance'
```
justinlittman, line 16: Do we need to do more for maintenance to work correctly?

justinlittman (dismissed):

config/environments/production.rb
lwrubel: I thought we were going the path of using a separate database for solid queue. I'm in the process of (trying to) set that up. It would be called queue, the default.

config/cable.yml
lwrubel: I've got a PR up for this, and a question out to Ops.

config/environments/production.rb
aaron-collier: Yeah, this was only to get the deploy to work until we have them. If they will be ready before we want to merge this, I can change it ASAP.

app/assets/config/manifest.js
aaron-collier: The deploy fails on a manifest_backup step without it. I didn't look to deeply yet to see if we can skip that step.

Capfile
aaron-collier: Maybe? I can add more to this PR, I was focused so far on just getting the deployment done. But happy to add that here or in a follow up.

aaron-collier: @justinlittman I don't believe so, The PR when you added to H2 was just the gem (https://github.com/sul-dlss/happy-heron/commit/6c5fc9835cbee95efc42bdece6f49765a44fe4e5) and looking at the readme (https://github.com/capistrano/maintenance) I think we might just have some template work to do when we actually want to put up a maintenance page, but I think that can happen later.

mjgiarlo: Should we yank out the maintenance stuff and add a backlog ticket to set up maintenance for real?

app/assets/config/manifest.js
mjgiarlo: @aaron-collier See the "require what you need manually" section of https://github.com/capistrano/rails#usage

config/deploy.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# config valid for current version and patch releases of Capistrano
+lock '~> 3.19.1'
```
mjgiarlo, line 4: I believe we've usually removed the `lock` from our `config/deploy.rb` files. I can only find one project where we didn't.

mjgiarlo (commented):
Consider removing manual OTK copypasta but otherwise good to go

mjgiarlo (approved):

PR #49, opened 2024-11-07
Authentication

### LINKED ISSUE #41: Explore rails 8 authentication

justinlittman:
closes #41

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ require:
   - rubocop-capybara
   - rubocop-rspec
   - rubocop-rspec_rails
+  - rubocop-factory_bot
 
 AllCops:
   TargetRubyVersion: 3.3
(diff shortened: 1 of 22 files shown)


.rubocop.yml
```diff
@@ -1,10 +1,11 @@
 require:
-  - rubocop-rails
+  - rubocop-rails
```
aaron-collier, line 2: Some trailing whitespace. I guess rubocop doesn't check .rubocop.yml :-)

app/controllers/concerns/authentication.rb
```diff
@@ -0,0 +1,113 @@
...
+  def current_user
+    Current.user
+  end
+
+  private
+
+  def remote_user
+    return DEV_REMOTE_USER if Rails.env.development?
+
+    request.headers['X-Remote-User']
```
mjgiarlo, line 57: Could this use the `REMOTE_USER_HEADER` constant instead of the string?

db/migrate/20241106143736_create_users.rb
```diff
@@ -0,0 +1,12 @@
+class CreateUsers < ActiveRecord::Migration[8.0]
+  def change
+    create_table :users do |t|
+      t.string :email_address, null: false
```
mjgiarlo, line 4: Do we want any other constraints here, like for uniqueness?

spec/support/authentication.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+def authentication_headers_for(user)
+  {
+    Authentication::REMOTE_USER_HEADER => user.email_address,
+    Authentication::NAME_HEADER => user.name,
+    Authentication::FIRST_NAME_HEADER => user.first_name
+  }
+end
```
mjgiarlo, line 9: Worth considering a structure like https://github.com/sul-dlss/happy-heron/blob/main/spec/support/session_helpers.rb so that this method is only in the global namespace during request specs where it's needed, or do you think we'll want this in other sorts of specs?

app/controllers/concerns/authentication.rb
justinlittman: Yes, fixed.

spec/support/authentication.rb
justinlittman: Will probably want it in at least request and system. In this case, seems like unnecessary complexity that might cause future confusion.

mjgiarlo (approved):

config/routes.rb
```diff
@@ -10,7 +13,7 @@
   # get "service-worker" => "rails/pwa#service_worker", as: :pwa_service_worker
 
   # Defines the root path route ("/")
-  # root "posts#index"
+  root to: 'welcome#show'
```
aaron-collier, line 16: I don't see a corresponding welcome_controller.rb...

PR #68, opened 2024-11-09
Adds initial mapping to cocina.

### LINKED ISSUE #62: Add cocina-models to the application
We'll need to be able to create the cocina model to deposit to dor-services-app and display works.

justinlittman:
closes #62

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,9 @@ gem 'config'
 gem 'honeybadger'
 gem 'view_component'
 
+# DLSS/domain-specific dependencies
+gem 'cocina-models'
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem 'debug', platforms: %i[mri windows], require: 'debug/prelude'
--- a/app/services/cocina_description_support.rb
+++ b/app/services/cocina_description_support.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+# Helpers for working with Cocina descriptions
+class CocinaDescriptionSupport
+  # ROLES = {
+  #   AUTHOR: {
+  #     value: 'author',
+  #     code: 'aut',
+  #     uri: 'http://id.loc.gov/vocabulary/relators/aut',
+  #     source: {
+  #       code: 'marcrelator',
+  #       uri: 'http://id.loc.gov/vocabulary/relators/'
+  #     }
+  #   },
+  #   PUBLISHER: {
+  #     value: 'publisher',
+  #     code: 'pbl',
+  #     uri: 'http://id.loc.gov/vocabulary/relators/pbl',
+  #     source: {
+  #       code: 'marcrelator',
+  #       uri: 'http://id.loc.gov/vocabulary/relators/'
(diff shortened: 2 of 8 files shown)


app/services/cocina_description_support.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
+
+# Helpers for working with Cocina descriptions
+class CocinaDescriptionSupport
+  # ROLES = {
```
justinlittman, line 5: I've left a bunch of cruft in some of this code copied from H3 POC and SHROOM with the assumption that we will be needing it shortly. If we don't, it can be cleared out later.

Gemfile
```diff
@@ -43,6 +43,9 @@ gem 'config'
 gem 'honeybadger'
 gem 'view_component'
 
+# DLSS/domain-specific dependencies
+gem 'cocina-models'
```
mjgiarlo, line 47: I might lobby for us to kick around how we want to organize our `Gemfile`s to make sure we want to switch from alphabetical to alphabetical-by-domain or whatever. Do we have preferences, @sul-dlss/infrastructure-team ❓ 

I have a slight preference for straight up alphabetical but I'm open to other suggestions.

spec/services/to_cocina/mapper_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ToCocina::Mapper do
+  subject(:cocina_object) { described_class.call(work_form:, source_id:) }
+
+  let(:source_id) { "h3:object-#{Time.zone.now.iso8601}" }
+
+  context 'when a new work' do
```
mjgiarlo, line 10:
```suggestion
  context 'with a new work' do
```

spec/support/cocina_matchers.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  failure_message do |actual|
+    SuperDiff::EqualityMatchers::Hash.new(
+      expected: expected.to_h.deep_symbolize_keys,
+      actual: actual.to_h.deep_symbolize_keys
+    ).fail
+  rescue StandardError => e
+    "Error in CocinaMatchers: #{e}"
+  end
+end
```
mjgiarlo, line 21: Open to using the matcher we get for free from cocina-models? https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/rspec/matchers.rb#L46-L62

Gemfile
justinlittman: I'm OK with any consistent approach.

spec/support/cocina_matchers.rb
justinlittman: That matcher takes a JSON string and a cocina object. I've always found that confusing and isn't a good fit for tests like these where we are comparing 2 cocina objects.

That being said, this is a slight preference so happy to defer to your judgment.

Gemfile
justinlittman: Rather, any consistent approach that is in a separate PR.

mjgiarlo (approved):

PR #69, opened 2024-11-10
Add okcomputer for initial status check

aaron-collier:
This add the `/status` and `/status/all` paths that we use in most projects. Currently this only checks the application database, a follow up PR will add checks for the `solid_*` db checks.


mjgiarlo (approved):
Were you thinking of setting up OKC in a follow-up PR?

PR #72, opened 2024-11-11
Adds initial deposit job.

### LINKED ISSUE #61: Implement deposit_work job
Implement initial deposit job to deposit directly to dor-services-app

justinlittman:
closes #61

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,8 @@ gem 'view_component'
 
 # DLSS/domain-specific dependencies
 gem 'cocina-models'
+gem 'dor-services-client', '>= 15.2.1'
+gem 'dor-workflow-client', '>= 7.6.1'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/app/jobs/deposit_job.rb
+++ b/app/jobs/deposit_job.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# Performs a deposit (without SDR API).
+class DepositJob < ApplicationJob
+  # @param [WorkForm] work_form
+  # @param [Boolean] deposit if true, deposit the work; otherwise, leave as draft
+  def perform(work_form:, deposit:)
+    cocina_object = ToCocina::Mapper.call(work_form:, source_id: source_id)
+    # Until we're also doing updates:
+    new_cocina_object = Sdr::Repository.register(cocina_object:)
+    # new_cocina_object = if work_form.persisted?
+    #                       Sdr::Repository.open_if_needed(cocina_object:)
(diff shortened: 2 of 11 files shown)


Gemfile
```diff
@@ -45,6 +45,8 @@ gem 'view_component'
 
 # DLSS/domain-specific dependencies
 gem 'cocina-models'
+gem 'dor-services-client', '>= 15.2.1'
+gem 'dor-workflow-client', '>= 7.6.1'
```
mjgiarlo, line 49: Any reason not to be a bit more pessimistic, in case new major versions break things?

app/jobs/deposit_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Performs a deposit (without SDR API).
+class DepositJob < ApplicationJob
+  # @param [WorkForm] work_form
+  # @param [Boolean] deposit if true, deposit the work; otherwise, leave as draft
+  def perform(work_form:, deposit:)
```
mjgiarlo, line 7: Do we want `deposit` to have a default value, or should we expect callers to provide this arg every time so it's more explicit?

spec/serializers/work_form_serializer_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      it 'returns true' do
+        expect(described_class.serialize?(WorkForm.new)).to be true
+      end
+    end
+
+    context 'when not a Work Form' do
+      it 'returns false' do
+        expect(described_class.serialize?(ApplicationForm.new)).to be false
+      end
+    end
```
mjgiarlo, line 26:
```suggestion
    context 'with a Work Form' do
      it 'returns true' do
        expect(described_class.serialize?(WorkForm.new)).to be true
      end
    end

    context 'with something other than a Work Form' do
      it 'returns false' do
        expect(described_class.serialize?(ApplicationForm.new)).to be false
      end
    end
```

mjgiarlo (approved):
Approve pending consideration of comments

Gemfile
justinlittman: The bigger problem we've been having is apps not getting upgraded because they're pinned.

In general, since we control all of the apps and these are all API clients, we tend to just change the APIs, change the clients, and update the apps without worrying about backwards compatibility. Well, at least that is my perspective.

app/jobs/deposit_job.rb
justinlittman: Yes, as of now the deposit job will only be invoked in one place (the work controller) and that controller will need to be explicit in determining the deposit value. (It will depend on which button is pressed.)

Gemfile
justinlittman: @sul-dlss/infrastructure-team happy to discuss this further and if necessary amend with additional PR.

PR #74, opened 2024-11-12
Deposit button

lwrubel:
Screenshot 2024-11-12 at 9 35 33 AM

Note that we need to add the upper limit for filesize at some point.

### DIFF
--- a/app/components/works/edit/deposit_component.html.erb
+++ b/app/components/works/edit/deposit_component.html.erb
@@ -1,6 +1,6 @@
 <p>
   Be sure you have completed all the required sections. Your deposit may take several hours
-  to complete if your filesize is larger than GB.
+  to complete if your filesize is larger than 10GB.
 </p>
 <%= link_to 'Cancel', works_path %>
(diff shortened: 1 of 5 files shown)


app/components/works/edit/deposit_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<p>
+  Be sure you have completed all the required sections. Your deposit may take several hours
+  to complete if your filesize is larger than GB.
```
mjgiarlo, line 3: Intentional to leave the # of GB blank for now?

mjgiarlo (approved):
Looks good! Assuming we're fine leaving the number of GB value blank for now, :shipit:.

app/components/works/edit/deposit_component.html.erb
lwrubel: Wasn't sure what to put there in MVP0, so left it blank. But I could put a placeholder value? I'll put 10GB there for now, following what we have with H2.

PR #80, opened 2024-11-12
Perform deposit.

### LINKED ISSUE #71: Perform deposit

justinlittman:
closes #71


app/controllers/works_controller.rb
```diff
@@ -3,8 +3,11 @@
 # Controller for a Work
 class WorksController < ApplicationController
   def show
+    work = Work.find_by!(druid: params[:druid])
+    return redirect_to wait_works_path(work.id) if work.deposit_job_started?
+
     # Stubbing this out temporarily until we have retrieving from the repository.
```
aaron-collier, line 9: Is this comment still relevant? Not a blocker obviously.

app/jobs/deposit_job.rb
```diff
@@ -3,9 +3,10 @@
 # Performs a deposit (without SDR API).
 class DepositJob < ApplicationJob
   # @param [WorkForm] work_form
+  # @param [Work] work
   # @param [Boolean] deposit if true, deposit the work; otherwise, leave as draft
-  def perform(work_form:, deposit:)
-    cocina_object = ToCocina::Mapper.call(work_form:, source_id: source_id)
+  def perform(work_form:, work:, deposit:)
+    cocina_object = ToCocina::Mapper.call(work_form:, source_id: "h3:object-#{work.id}")
     # Until we're also doing updates:
```
aaron-collier, line 10: Not necessarily for this PR, but I wonder if we want to add some kind of specific text (i.e. "TODO") to comments like this, that are very helpful, to make sure we don't forget about areas of code that we anticipate removing or know are temporary.

config/initializers/sdr.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Configure dor-services-client to use the dor-services URL
+Dor::Services::Client.configure(url: Settings.dor_services.url,
+                                token: Settings.dor_services.token)
```
aaron-collier, line 5: I like the explicitness of this being an initializer, does this mean all calls through `Dor::Services::Client` have this configuration? I like that.

aaron-collier (approved):

config/initializers/sdr.rb
justinlittman: Yes. Many (but not all) of our clients are so configured.

PR #84, opened 2024-11-13
Adds mapping to work form and retrievel of cocina from SDR.

### LINKED ISSUE #83: Map cocina to work form

justinlittman:
closes #83


app/jobs/deposit_job.rb
```diff
@@ -22,6 +22,7 @@ def perform(work_form:, work:, deposit:)
     work.update!(deposit_job_started_at: nil, druid:)
     # Just to be aware for future troubleshooting: There is a possible race condition between the websocket
     # connecting and the following broadcast being sent.
+    sleep 0.5 if Rails.env.test? # Avoids race condition in tests
```
aaron-collier, line 25: I wonder if there is a way to do this without `Rails.env.test?` being in the job code. Not a blocker, just something to think about.

aaron-collier (approved):

PR #90, opened 2024-11-13
Extract fixtures.

justinlittman:


spec/support/cocina_fixtures.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
mjgiarlo, line 1: We should chat about whether we use this approach or factories or both

justinlittman: Yeah, I understand this is not our normal pattern so worth a discussion.

mjgiarlo (approved):

PR #94, opened 2024-11-14
Set up collection model

### LINKED ISSUE #82: Create initial collection model
Associate with user and work

lwrubel:
Resolves #82 to create collection model.


db/migrate/20241114175554_create_collection.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCollection < ActiveRecord::Migration[8.0]
+  def change
+    create_table :collections do |t|
+      t.string :druid
```
justinlittman, line 4: druid and title should not null.

lwrubel: Thank you, fixed.

spec/system/show_work_spec.rb
```diff
@@ -4,7 +4,8 @@
 
 RSpec.describe 'Show a work', :rack_test do
   let(:druid) { druid_fixture }
-  let!(:work) { create(:work, druid: druid, title: title_fixture) }
+  let!(:collection) { create(:collection) }
+  let!(:work) { create(:work, druid: druid, title: title_fixture, collection:) }
```
aaron-collier, line 8: Are the `let!` necessary over a normal `let` here? Not a blocker, mostly curious if you found this was required.

aaron-collier (approved):
Approved with non-blocking consideration for the comment above.

app/controllers/works_controller.rb
```diff
@@ -14,19 +14,24 @@ def new
     @work_form = WorkForm.new
   end
 
+  # rubocop:disable Metrics/AbcSize
   def create
     @work_form = WorkForm.new(work_params)
     # The deposit param determines whether extra validations for deposits are applied.
     if @work_form.valid?
+      # TODO: Once we have a path from the dashboard, remove this step to create a collection
+      collection = Collection.create!(title: 'Temp Collection', druid: 'druid:cc234dd5678', user: current_user)
```
justinlittman, line 23: Can you do this as a `create_or_find_by` to avoid dupes?

db/migrate/20241114231056_create_collection.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCollection < ActiveRecord::Migration[8.0]
+  def change
+    create_table :collections do |t|
+      t.string :druid, null: false
+      t.string :title, null: false
+      t.references :user, null: false, foreign_key: true
+
+      t.timestamps
```
justinlittman, line 8: Can you put a unique index on druid?

justinlittman (approved):
Approve pending addressing comments.

spec/system/show_work_spec.rb
lwrubel: Hmm, I'm not sure that it is, although the `let!` was there in the code before for instantiating `work`, so perhaps we were anticipating a situation where it would be helpful to not create lazily? I removed it for now.

justinlittman (approved):

PR #95, opened 2024-11-14
Initial Dashboard

aaron-collier:
Super basic dashboard that currently only lists the current users works by title and links to them by druid.

Screenshot 2024-11-15 at 7 18 25 AM


app/models/work.rb
```diff
@@ -14,4 +14,10 @@ def deposit_job_started?
   def deposit_job_finished?
     deposit_job_started_at.nil?
   end
+
+  def status
```
justinlittman, line 18: Can we not put this on the model?

app/components/dashboard/show/works_list_component.html.erb
```diff
@@ -0,0 +1,22 @@
+<h3 class="h3 mb-4"><%= label %></h3>
+<%= render Elements::Tables::TableComponent.new(
+      id: 'worksTable',
```
justinlittman, line 3: Can we do ids with dashes, e.g., `works-table`?

app/components/elements/tables/table_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= tag.table class: classes, id: do %>
+  <% if label.present? %>
+    <caption><%= label %></caption>
+  <% end %>
+  <%= render Elements::Tables::TableHeaderComponent.new(classes: header_classes, headers: headers) if headers %>
```
justinlittman, line 5: Instead of `if headers` add a `renders?` method to the TableHeaderComponent.

app/components/elements/tables/table_header_component.rb
```diff
@@ -0,0 +1,17 @@
...
+    # Component for rendering a table header section.
+    class TableHeaderComponent < ApplicationComponent
+      def initialize(classes:, headers: nil)
+        # If value is not provided, content block will be rendered instead.
+        @classes = classes
+        @headers = headers
+        super()
+      end
+
+      attr_reader :classes, :headers
```
justinlittman, line 14: There should be a `merge_classes` performed on `@classes`.

spec/components/dashboard/show/works_list_component_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Dashboard::Show::WorksListComponent, type: :component do
+  let(:work) { create(:work) }
```
justinlittman, line 6: Doesn't this work need to be assigned to the user?

app/controllers/dashboard_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Controller for the user dashboard
+class DashboardController < ApplicationController
+  def show; end
```
justinlittman, line 5: Can you add a request or system spec for this page?

justinlittman (changes requested):

app/components/dashboard/show/works_list_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+           current_user.name,
+           I18n.l(work.updated_at, format: '%b %d, %Y'),
+           link_to(nil, Sdr::Purl.from_druid(druid: work.druid))
+         ]
+       ) %>
+  <% end %>
+<% end %>
+
+<%= link_to new_work_path, class: 'btn btn-primary' do %>
+  <div class="align-self-center d-flex justify-content-center"><%= icon %> Add new work</div>
```
justinlittman, line 22: How about just putting the plus icon here? Putting in the VC class adds an unnecessary layer of abstraction.

justinlittman (approved):
Approved pending consideration of minor comment.

app/components/dashboard/show/works_list_component.html.erb
aaron-collier: Done. We may want to chat (maybe part of retro) how and when we abstract into the helper(s)...

PR #96, opened 2024-11-14
Adds editing a work.

### LINKED ISSUE #91: Implement editing

justinlittman:
closes #91

https://github.com/user-attachments/assets/1911fad5-676d-4cc5-9b9c-99f4eb6111ef


spec/support/authentication_helpers.rb
```diff
@@ -15,7 +15,14 @@ def sign_in(user, groups: [])
...
   end
 
+  def clear_authentication
+    TestShibbolethHeaders.user = nil
+    TestShibbolethHeaders.groups = nil
+  end
+
   RSpec.configure do |config|
+    config.before(:each, type: :request) { clear_authentication }
+    config.before(:each, type: :system) { clear_authentication }
```
mjgiarlo, line 25: Can you say why we need to sign out the user before every request and system spec?

spec/support/authentication_helpers.rb
```diff
@@ -15,7 +15,14 @@ def sign_in(user, groups: [])
     TestShibbolethHeaders.groups = groups
   end
 
+  def clear_authentication
```
mjgiarlo, line 18: Would you consider renaming this method `sign_out` or something similar?

spec/support/authentication_helpers.rb
justinlittman: Added a comment to explain.

mjgiarlo (approved):
Even if we don't have a specific need for the automatic signing out (since I don't think we did this in H2?), it's a good idea to eliminate sticky state between specs. I'm 💯 in favor of preventing confusion down the line.

PR #108, opened 2024-11-18
Extract locales

justinlittman:


app/views/works/form.html.erb
```diff
@@ -1,29 +1,23 @@
 <h1 class="h2 mb-3"><%= @work_form.title || 'Untitled deposit' %></h1>
  <div class="d-flex tabbable-panes">
     <div class="nav nav-pills flex-column px-2 py-1 me-5 col-3" role="tablist" aria-orientation="vertical">
-        <%= render Works::Edit::TabComponent.new(label: 'Title', tab_name: :title, selected: true) %>
-        <%= render Works::Edit::TabComponent.new(label: 'Abstract', tab_name: :abstract) %>
-        <%= render Works::Edit::TabComponent.new(label: 'Deposit', tab_name: :deposit) %>
+        <%= render Works::Edit::TabComponent.new(label: I18n.t('works.edit.panes.title.tab_label'), tab_name: :title, selected: true) %>
+        <%= render Works::Edit::TabComponent.new(label: I18n.t('works.edit.panes.abstract.tab_label'), tab_name: :abstract) %>
+        <%= render Works::Edit::TabComponent.new(label: I18n.t('works.edit.panes.deposit.tab_label'), tab_name: :deposit) %>
```
mjgiarlo, line 6: Not a big deal, but in the view context, you can drop the explicit `I18n.` module namespace (see the various view snippets [in the Rails i18n docs](https://guides.rubyonrails.org/i18n.html))

mjgiarlo: (Actually, it may be needed given this is a component instantiation, not a string interpolation in a view. You probably already considered this!)

mjgiarlo (approved):

PR #120, opened 2024-11-18
Add has_many collections relationship to user

aaron-collier:
Small tweak to the user model to allow user.collections list (useful for the dashboard).


app/models/user.rb
```diff
@@ -3,4 +3,5 @@
 # Model for a user.
 class User < ApplicationRecord
   has_many :works, dependent: :destroy
+  has_many :collections
```
justinlittman, line 6: `dependent: :destroy`?

justinlittman (approved):

PR #123, opened 2024-11-19
Update dashboard to list collections and works

### LINKED ISSUE #119: Add user collections to the user dashboard

aaron-collier:
Update the dashboard to list work by collection:

Screenshot 2024-11-18 at 4 22 31 PM

Fixes #119


app/components/elements/header_component.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Elements
+  # Applies an h# component with the expected styling.
+  class HeaderComponent < ApplicationComponent
+    def initialize(variant, value:, classes: [])
+      raise ArgumentError, 'Invalid variant' unless %w[h1 h2 h3 h4 h5 h6].include?(variant)
```
justinlittman, line 7: This should also allow for a symbol to be passed as variant, e.g., `:h1`.

app/components/elements/header_component.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Elements
+  # Applies an h# component with the expected styling.
+  class HeaderComponent < ApplicationComponent
+    def initialize(variant, value:, classes: [])
```
justinlittman, line 6: I'd recommend calling the initial arg "level" and making it a kwarg for consistency: `def initialize(level:, value:, classes: [])`

app/components/elements/header_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= content_tag variant, value, class: classes %>
```
justinlittman, line 1: This is simple enough to use `call()` in the component instead of a separate erb.

app/components/dashboard/show/collections_list_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render Elements::HeaderComponent.new('h3', classes: ['mb-4'], value: label) %>
```
justinlittman, line 1: Will `h3` lead to gaps in the heading levels (accessibility problem)?

Alternative is something like `<h2 class="h3">`.

Also, FYI -- you can just do `classes: 'mb-4'` and the class merger will sort it out.

app/components/dashboard/show/collections_list_component.html.erb
```diff
@@ -0,0 +1,10 @@
...
+
+<% collections.each do |collection| %>
+  <%= render Elements::HeaderComponent.new('h4', value: link_to(collection.title, "/collections/#{collection.druid}")) %>
```
justinlittman, line 4: Please use a path helper instead of hardcoding. (If there isn't a path yet then don't link.)

app/components/dashboard/show/collections_list_component.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <%= render Dashboard::Show::WorksListComponent.new(collection:) %>
+
+  <%= link_to new_work_path, class: 'btn btn-primary' do %>
```
justinlittman, line 7: This should be an element component.

app/components/dashboard/show/collections_list_component.html.erb
```diff
@@ -0,0 +1,10 @@
...
+    <div class="align-self-center d-flex justify-content-center">Deposit to this collection</div>
```
justinlittman, line 8: Why does the label need to be in a div?

spec/components/dashboard/show/collections_list_component_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Dashboard::Show::CollectionsListComponent, type: :component do
+  let!(:top_collection) { create(:collection, user:, druid: 'druid:ab234dd5678') }
+  let!(:botom_collection) { create(:collection, user:, druid: 'druid:bc234dd5678') }
```
justinlittman, line 7: Typo

justinlittman (changes requested):

app/components/dashboard/show/collections_list_component.html.erb
aaron-collier: I noticed we do this a lot in H2.. `<h1 class="h1...` and `<h3 class="h4...`, I was thinking those _should_ align, but probably not, so I went with `level` as requested, and kept `variant` as optional to add to the class.

app/components/dashboard/show/collections_list_component.html.erb
aaron-collier: Actually, I realized in this case we don't have a collection show page yet, so I'm making this just the text of the title. I'll write up a ticket for the collection show page and updating this link

app/components/dashboard/show/collections_list_component.html.erb
aaron-collier: Cruft from H2 I think, this goes away when using the `ButtonLinkComponent`

justinlittman (approved):

PR #125, opened 2024-11-19
Adds form validation.

### LINKED ISSUE #93: Validation - display errors on the work form
- Show banner alert
- Highlighting the tab with the error
- Display the error under the field
- Style field with red/error indication
- Take user to first tab with error

justinlittman:
closes #93

image

^^ need a design for when a tab is selected and has an error.

### DIFF
--- a/app/views/works/form.html.erb
+++ b/app/views/works/form.html.erb
@@ -1,5 +1,6 @@
 <h1 class="h2 mb-3"><%= @work_form.title || 'Untitled deposit' %></h1>
- <div class="d-flex tabbable-panes">
+ <%= render Elements::AlertComponent.new(variant: :danger, value: I18n.t('works.edit.errors.validation')) if @work_form.errors.present? %>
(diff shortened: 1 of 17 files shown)


app/views/works/form.html.erb
```diff
@@ -1,5 +1,6 @@
 <h1 class="h2 mb-3"><%= @work_form.title || 'Untitled deposit' %></h1>
- <div class="d-flex tabbable-panes">
+ <%= render Elements::AlertComponent.new(variant: :danger, value: I18n.t('works.edit.errors.validation')) if @work_form.errors.present? %>
```
aaron-collier, line 2: I wonder if it would be possible to use a `render?` in the component as we have elsewhere so we don't use the inline conditional here. Not a blocker, and could easily be a later refactor if we decide that pattern would work here later.

justinlittman: The alert component is generic; it doesn't know anything about work forms or errors. (I had the same thought too.)

aaron-collier (approved):

PR #153, opened 2024-11-21
Adds Next button to progress through work form tabs

### LINKED ISSUE #111: Add next button to tab panes
image

For reference, see https://github.com/sul-dlss-labs/hungry-hungry-hippo-poc/blob/main/app/components/tab/next_button_component.rb and https://github.com/sul-dlss-labs/hungry-hungry-hippo-poc/blob/main/app/javascript/controllers/tab_next_controller.js

lwrubel:
Closes #111 to add a next button to panes (except for the last pane).

Screenshot 2024-11-21 at 3 25 58 PM

### DIFF
--- a/app/components/elements/button_component.rb
+++ b/app/components/elements/button_component.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Elements
+  # Generic button component
+  class ButtonComponent < ViewComponent::Base
+    def initialize(label:, classes: [], variant: nil, size: nil, **options)
+      @label = label
+      @classes = classes
+      @variant = variant
+      @size = size
+      @options = options
+      super()
+    end
+
+    attr_reader :options, :label
+
+    def call
+      tag.button(
+        class: Elements::ButtonSupport.classes(variant: @variant, size: @size, classes: @classes),
+        type: 'button',
+        **options
+      ) do
+        label
+      end
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/javascript/controllers/tab_next_controller.js
```diff
@@ -0,0 +1,16 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { selector: String }
+
+  show(event) {
+    event.preventDefault();
+
+    const tabsOrder = ["title", "abstract", "related_content", "deposit"]
```
justinlittman, line 9: I really don't like having this hard-coded in here. It is inevitable that we will changes tabs and forget to update.

justinlittman (changes requested):

app/javascript/controllers/tab_next_controller.js
lwrubel: Yeah, that was a compromise since the code for the next button is shared, unlike the POC. I thought that the system test, by using the tabs, would catch any missed updates for future tab changes. I'll try to do this another way.

lwrubel: next tab selection is now a param in each pane

app/components/elements/button_component.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Elements
```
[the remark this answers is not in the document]
lwrubel: Added a spec

PR #157, opened 2024-11-21
Wire up validation of nested models with appropriate error routing

mjgiarlo:


lib/active_model/nested_attributes.rb
```diff
@@ -71,6 +73,32 @@ def allow_serialization_of_nested_models(models)
...
+          self
+            .class
+            .nested_attributes_hash
+            .keys
+            .flat_map { |nested_attr| public_send(nested_attr) }
+        end
+      end
+
+      def define_validation_methods
+        define_method(:valid?) do |*args, **kwargs, &block|
```
mjgiarlo, line 89: Anonymous block parameter forwarding is not a thing in Ruby 3.3, so we have to handle args the old-fashioned way. This gets better in Ruby 3.4.

justinlittman (approved):

PR #158, opened 2024-11-21
Adds authorization for Content and Content Files controllers.

### LINKED ISSUE #152: Add authorization to Content and Content Files controllers
Currently, they have `skip_verify_authorized`

justinlittman:
closes #152

### DIFF
--- a/app/controllers/content_files_controller.rb
+++ b/app/controllers/content_files_controller.rb
@@ -2,15 +2,19 @@
 
 # Controller for a Work content file
 class ContentFilesController < ApplicationController
-  # Skipping authorization for now
-  skip_verify_authorized only: %i[show edit update destroy]
   before_action :set_content_file
 
-  def show; end
+  def show
+    authorize! @content_file
+  end
 
-  def edit; end
+  def edit
+    authorize! @content_file
+  end
 
   def update
+    authorize! @content_file
+
     if @content_file.update(content_file_params)
       redirect_to content_file_path(@content_file)
     else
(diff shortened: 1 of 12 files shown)


app/controllers/content_files_controller.rb
```diff
@@ -2,15 +2,19 @@
 
 # Controller for a Work content file
 class ContentFilesController < ApplicationController
-  # Skipping authorization for now
-  skip_verify_authorized only: %i[show edit update destroy]
```
mjgiarlo, line 6: Can you replace the changes to the action methods below with:

```
  verify_authorized only: %i[show edit update destroy]
```

❓

app/controllers/contents_controller.rb
```diff
@@ -2,17 +2,21 @@
 
 # Controller for a Work contents (files)
 class ContentsController < ApplicationController
-  # Skipping authorization for now
-  skip_verify_authorized only: %i[show edit update]
```
mjgiarlo, line 6: Same question as other controller.

mjgiarlo (approved):
Approve pending consideration of AP question in controllers

app/controllers/content_files_controller.rb
justinlittman: We already do that here: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/controllers/application_controller.rb#L11

`verify_authorized` makes sure that `authorize!` is called; it doesn't do the calling itself.

mjgiarlo: Ah, thank you!

PR #159, opened 2024-11-22
Adjust the validate work deposit spec now that files are at the top

aaron-collier:
This might be a temporary fix, but this test is currently failing because the page always defaults the active to the files tab (at the moment)


mjgiarlo (commented):
Check out https://github.com/sul-dlss/hungry-hungry-hippo/pull/160

PR #161, opened 2024-11-22
Fix bug that prevented multiple form errors from showing in the UI at the same time

mjgiarlo:


app/forms/application_form.rb
```diff
@@ -23,7 +23,7 @@ def self.immutable_attributes
   # @param deposit [Boolean] whether the form is being used for a deposit
   def valid?(deposit: false)
     @deposit = deposit
-    super()
```
mjgiarlo, line 26: It's fine to let the args bubble up, apparently.

lib/active_model/nested_attributes.rb
```diff
@@ -96,7 +96,11 @@ def define_validation_methods
             end
 
             false
-          end && super(*args, **kwargs, &block)
```
mjgiarlo, line 99: Before, if the nested forms did not validate, it short-circuited the validation of the main form. Now we run both validations, storing results, and returning the `&&` of the results

justinlittman (approved):

PR #163, opened 2024-11-22
Adds pagination to work form.

justinlittman:
refs #162

https://github.com/user-attachments/assets/16955ac5-6429-4f65-b12a-d5adcb9b53f9


config/initializers/kaminari_config.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+Kaminari.configure do |config|
+  config.default_per_page = 2
```
lwrubel, line 4: This is set low for development/demo purposes?

lwrubel (approved):

lwrubel:
The pagination component is "dev ready" but not complete. Although component-izing it would be a separate effort, would it make sense to generally align with in the meantime?  

Screenshot 2024-11-25 at 10 14 24 AM

https://docs.google.com/presentation/d/1M4pBdijLciVsRXCJKIC4TEG2xVXNw96SiODnSDQ1pjw/edit#slide=id.g36c8395e53bfd5b9_45

"First" and "Last" are `<<` `>>`  instead of words and there is no bracket on Previous and Next. 

Or could go in another ticket.

config/initializers/kaminari_config.rb
justinlittman: Yes, and apparently I forgot to undo it.

justinlittman:
I opted to ticket separately: https://github.com/sul-dlss/hungry-hungry-hippo/issues/176

PR #178, opened 2024-11-25
Rename mappers, fixtures, and deposit job for clarity

aaron-collier:
This moves the work related cocina mappers into `cocina/work/...`, renames the `DepositJob` to `DepositWorkJob`, and the fixtures from `MappingFixtures` to `WorkMappingFixtures` for clarity before adding the `Collection*` components.

Doing this separately from adding the collection now to avoid complicated rebases for any in flight work.


.rubocop.yml
```diff
@@ -24,7 +24,7 @@ AllCops:
     - 'vendor/**/*'
 
 Layout/LineLength:
-  Max: 120
+  Max: 130
```
justinlittman, line 27: Why is this being changed?

https://github.com/sul-dlss/DeveloperPlaybook/blob/main/style/infra_template.rubocop.yml#L34-L35

justinlittman (approved):
Approve pending addressing comment.

.rubocop.yml
aaron-collier: These 2 lines are 2 characters too long:
```
spec/jobs/deposit_work_job_spec.rb:26:121: C: [Correctable] Layout/LineLength: Line is too long. [122/120]
      expect(ToCocina::Work::Mapper).to have_received(:call).with(work_form:, content:, source_id: "h3:object-#{work.id}")
                                                                                                                        ^^
spec/jobs/deposit_work_job_spec.rb:46:121: C: [Correctable] Layout/LineLength: Line is too long. [122/120]
      expect(ToCocina::Work::Mapper).to have_received(:call).with(work_form:, content:, source_id: "h3:object-#{work.id}")
```

Initially I felt like it reads better without breaking the line. But being that this is in the playbook, I'll revert and stick to the standard. Good call.

PR #188, opened 2024-11-26
Spike on automagic nested component rendering for simple forms (all strings)

mjgiarlo:


app/components/related_links/edit_component.rb
```diff
@@ -1,13 +0,0 @@
...
-  # Component for editing related links
-  class EditComponent < ApplicationComponent
-    def initialize(form:)
-      @form = form
-      super()
-    end
-
-    attr_reader :form
-  end
-end
```
mjgiarlo, line 13: I asked myself yesterday, "what does this RelatedLinks::EditComponent add, really? could/should we handle 'simple' nested forms more automagically?" This is a spike showing one way how to do that.

PR #185, opened 2024-11-26
Adds license selection with one license.

lwrubel:
Resolves #58, with one license. 

Screenshot 2024-11-26 at 1 41 59 PM
Screenshot 2024-11-26 at 1 42 07 PM
Screenshot 2024-11-26 at 1 41 39 PM

PR #190, opened 2024-11-26
Tweaks to license field

lwrubel:
Improvements to license fields based on comments in #185.

### DIFF
--- a/app/components/works/edit/license_component.rb
+++ b/app/components/works/edit/license_component.rb
@@ -4,23 +4,27 @@ module Works
   module Edit
     # Select field for license
     class LicenseComponent < ApplicationComponent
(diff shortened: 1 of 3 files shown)


app/components/works/edit/license_component.rb
```diff
@@ -4,23 +4,27 @@ module Works
...
         :license
       end
 
+      def label
+        I18n.t('works.edit.fields.license.label')
+      end
+
+      def help_text
+        I18n.t('works.edit.fields.license.help_text')
+      end
```
lwrubel, line 24: Moved these from `.html.erb` to the component.

justinlittman (approved):

PR #191, opened 2024-11-26
Handles content missing when redirecting to show page.

justinlittman:
image

Explanation below.

### DIFF
--- a/app/controllers/contents_controller.rb
+++ b/app/controllers/contents_controller.rb
@@ -2,20 +2,17 @@
 
 # Controller for a Work contents (files)
 class ContentsController < ApplicationController
-  before_action :set_content, only: %i[edit update show]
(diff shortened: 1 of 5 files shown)


app/controllers/contents_controller.rb
```diff
@@ -2,20 +2,17 @@
...
+  before_action :set_content_files, only: %i[show show_table]
 
+  # Called from work edit/update form.
   def show
     authorize! @content
+  end
 
-    @content_files = @content.content_files.order(filename: :asc).page(params[:page])
-
-    # This can be called from the work edit/update page or the work show page.
```
justinlittman, line 12: When roundtripping failed, the user is redirected here. However, in this case the referrer test below fails and the wrong template was being rendered.

mjgiarlo (approved):

PR #193, opened 2024-11-26
Fix button color on related conent pane.

justinlittman:
image

### DIFF
--- a/app/components/elements/forms/nested_component.html.erb
+++ b/app/components/elements/forms/nested_component.html.erb
@@ -5,30 +5,32 @@
   <%= tag.div(class: 'mb-3 container', data: { nested_form_target: 'container' }) do %>
     <template data-nested-form-target="template">
       <%= form.fields_for field_name, model_class.new, child_index: 'NEW_RECORD' do |nested_form| %>
+        <%# Note that this div is duplicated below. If making a change here, also make the change below. %>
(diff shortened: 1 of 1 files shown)


app/components/elements/forms/nested_component.html.erb
```diff
@@ -2,6 +2,11 @@
   <legend class="h5">
     <%= header_label %>
   </legend>
+  <% delete_button = capture do %>
```
justinlittman, line 5: Why DRY this? I initially missed that there were 2 identical copies of the button.

app/components/elements/forms/nested_component.html.erb
```diff
@@ -2,6 +2,11 @@
...
+    <%= render Elements::ButtonComponent.new(classes: 'border border-0 float-end', variant: :'outline-primary', size: :lg, data: { action: 'click->nested-form#delete' }) do %>
+      <i class="bi bi-trash"></i>
```
mjgiarlo, line 7:
```suggestion
      <%= helpers.delete_icon %>
```

app/components/elements/forms/nested_component.html.erb
justinlittman: I un-DRYed (wet?) it and added a comment to help clarify.

mjgiarlo (approved):
Appreciate the comment, and the CSS fix!

PR #196, opened 2024-11-27
Extract Tabbed form into Element::Layout components for reuse

aaron-collier:
Extracts the tabbing out of `Work::Edit` into `Elements::TabForm` in order to be reused for the Collection form (in follow up)

This includes the approach suggested in #205


app/components/works/edit/form_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: 'works/form_fields', locals: { work_form: @work_form, content: @files } %>
```
aaron-collier, line 1: I found this partial rendering approach was required mainly due to the turbo stream component.

app/components/collections/edit/form_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: 'collections/form_fields', locals: { collection_form: @collection_form } %>
```
aaron-collier, line 1: This is matching the pattern for works where this was required for turbo streams.

app/components/elements/layout/next_button_component.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-module Works
-  module Edit
+module Elements
+  module Layout
     # Next button to progress through tabbed form
```
aaron-collier, line 5: These are all just moving elements from `Works::Edit` to `Elements::Layout` for reuse with the collection form.

app/components/collections/edit/form_component.html.erb
justinlittman: To the best I can tell, there's no reason for this to be a component.

app/components/elements/layout/tabbed_form_component.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Elements
+  module Layout
```
justinlittman, line 4: "Layout" doesn't seem like an appropriately descriptive namespace.

app/components/works/edit/form_component.html.erb
justinlittman: Wrapping all of this in a component doesn't seem necessary. Can't you just pass in the content to the slot instead of using a component?

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+  before_action :set_collection_form_from_cocina, only: %i[show edit]
+  before_action :set_status, only: %i[show edit]
+
+  def show
+    authorize! @collection
+    @status_presenter = StatusPresenter.new(status: @status)
+  end
+
+  def new
+    # Once collection is being passed, should authorize that the user can create a work in that collection.
```
justinlittman, line 16: new can be authorized when creating a collection.

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    @collection_form = CollectionForm.new
+
+    render :form
+  end
+
+  def edit
+    authorize! @collection
+
+    unless editable?
+      flash[:danger] = I18n.t('collection.edit.errors.cannot_be_edited')
```
justinlittman, line 28: This needs a test.

app/controllers/collections_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+      return redirect_to collection_path(druid: params[:druid])
+    end
+    render :form
+  end
+
+  def create
+    # Once collection is being passed, should authorize that the user can create a work in that collection.
+    skip_verify_authorized!
```
justinlittman, line 36: This can be authorized.

app/views/collections/form.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Elements::HeaderComponent.new(level: :h1, variant: :h2, text: @collection_form.title || 'Untitled collection', classes: 'mb-3') %>
+<%= render Elements::AlertComponent.new(variant: :danger, value: I18n.t('collections.edit.errors.validation')) if @collection_form.errors.present? %>
+<%= render Elements::Layout::TabbedFormComponent.new do |component| %>
+  <% component.with_tab(label: I18n.t('collections.edit.panes.title.tab_label'), tab_name: :title, selected: true) %>
+  <% component.with_tab(label: I18n.t('collections.edit.panes.description.tab_label'), tab_name: :description) %>
+  <% component.with_tab(label: I18n.t('collections.edit.panes.related_content.tab_label'), tab_name: :related_content) %>
+  <% component.with_tab(label: I18n.t('collections.edit.panes.deposit.tab_label'), tab_name: :deposit) %>
+  <% component.with_form_collection(collection_form: @collection_form) %>
```
justinlittman, line 8: It seems like the tabs and the panes should be defined in the same place instead of 2 different places.

app/components/works/edit/form_component.html.erb
aaron-collier: I'll see. I ran into some issues earlier, but I may have worked those out and not come back to this.

app/components/elements/layout/tabbed_form_component.rb
aaron-collier: Yeah, I was flip flopping on name. Open to suggestion, or bringing it up a level just into Elements, or higher... Or in it's own namespace...

app/components/works/edit/form_component.html.erb
aaron-collier: Done

justinlittman (approved):

PR #202, opened 2024-11-27
Increases maxFileSize for dropzone. Tweaks progress bar.

justinlittman:
https://github.com/user-attachments/assets/6a749e68-58bd-4e98-ae4b-728cc9356462

### DIFF
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -5,24 +5,25 @@ export default class extends Controller {
     static targets = ["progress"]
 
     connect() {
-        console.log("Dropzone connected", this.element.id);
         this.dropzone = new Dropzone(this.element, {
             paramName: "content[files]",
             uploadMultiple: true,
             createImageThumbnails: false,
             addRemoveLinks: false,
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/dropzone_controller.js
```diff
@@ -5,24 +5,25 @@ export default class extends Controller {
...
             addRemoveLinks: false,
             disablePreviews: true,
+            maxFilesize: 10000, // 10GB
         });
+        this.progress = 0;
         this.dropzone.on("addedfiles", () => {
-            console.log('addedfiles');
             this.progressTarget.classList.remove("d-none");
+            // This shows the user that something is going on since there may be a paused before first progress.
+            this.updateProgress(2);
```
mjgiarlo, line 20: Why `2`?

mjgiarlo (approved):
👍🏻

app/javascript/controllers/dropzone_controller.js
justinlittman: Arbitrary. Large enough to be noticed, but small enough not to be deceptive.

PR #212, opened 2024-12-02
Remove unnecessary explicit namespaces in components

mjgiarlo:


justinlittman (approved):
Can this be rubocopped? Otherwise I will forget and continue to add.

mjgiarlo:
@justinlittman I looked for it and didn't find it, alas.

PR #217, opened 2024-12-03
Adds publication date field.

### LINKED ISSUE #114: Add the dates tab and fields to the work deposit form

justinlittman:
closes #114

https://github.com/user-attachments/assets/4e96ea54-335f-4c49-9033-c42cd408c57f

### DIFF
--- a/app/components/elements/forms/field_component.rb
+++ b/app/components/elements/forms/field_component.rb
@@ -4,7 +4,8 @@ module Elements
   module Forms
     # Base component for all form fields.
     class FieldComponent < ApplicationComponent
-      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, disabled: false, hidden: false, data: {}) # rubocop:disable Metrics/ParameterLists, Layout/LineLength
+      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists
+                     disabled: false, hidden: false, data: {}, placeholder: nil)
         @form = form
         @field_name = field_name
         @required = required
@@ -14,10 +15,12 @@ module Elements
         @hidden = hidden
         @disabled = disabled
         @data = data
+        @placeholder = placeholder
         super()
       end
 
-      attr_reader :form, :field_name, :required, :help_text, :hidden_label, :label, :hidden, :disabled, :data
+      attr_reader :form, :field_name, :required, :help_text, :hidden_label, :label, :hidden, :disabled, :data,
+                  :placeholder
 
       def help_text_id
         @help_text_id ||= form.field_id(field_name, 'help')
--- a/app/components/elements/forms/select_day_field_component.rb
+++ b/app/components/elements/forms/select_day_field_component.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Component for form day of the month fields
+    # Note that this uses select_day, which is not called on a form builder instance.
+    # Thus, some of the machinery of the form builder is provided below.
+    class SelectDayFieldComponent < FieldComponent
+      def value
+        form.object.public_send(field_name)
+      end
+
+      # Prefix for the field name, e.g., work[publication_date_attributes]
+      def prefix
+        form.object_name
+      end
+
+      def classes
+        merge_classes('form-select', form.object.errors[field_name].present? ? 'is-invalid' : nil)
+      end
+    end
+  end
+end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -87,9 +87,7 @@ class WorksController < ApplicationController
   private
 
   def work_params
-    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
-    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes] +
-    [{ publication_date_attributes: %i[year month day] }])
+    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes])
   end
 
   def update_work_params
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -87,7 +87,9 @@ class WorksController < ApplicationController
   private
 
   def work_params
-    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes])
+    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
+    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes] +
+    [{ publication_date_attributes: %i[year month day] }])
   end
 
   def update_work_params
--- a/app/forms/date_form.rb
+++ b/app/forms/date_form.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# Form object for handling date input
+class DateForm < ApplicationForm
+  attribute :year, :integer
+  validates :year,
+            numericality: { greater_than_or_equal_to: 1000, less_than_or_equal_to: Time.zone.now.year, allow_nil: true }
+  validates :year, presence: true, if: -> { month.present? }
+  attribute :month, :integer
+  validates :month, numericality: { greater_than_or_equal_to: 1, less_than_or_equal_to: 12, allow_nil: true }
+  validates :month, presence: true, if: -> { day.present? }
+  attribute :day, :integer
+  validates :day, numericality: { greater_than_or_equal_to: 1, less_than_or_equal_to: 31, allow_nil: true }
+
+  def to_edtf_s
+    EdtfSupport.to_edtf_s(year:, month:, day:)
+  end
+end
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -4,6 +4,12 @@
 class WorkForm < ApplicationForm
   accepts_nested_attributes_for :related_links, :related_works
 
+  def initialize(params = {})
+    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
+    @publication_date = DateForm.new
+    super
+  end
+
   def self.immutable_attributes
     ['druid']
   end
@@ -41,5 +47,22 @@ class WorkForm < ApplicationForm
 
     selected_license = WorkForm.licenses.select { |_, v| v['uri'] == license }
     selected_license.keys.first
+  end
+
+  # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
+  attr_accessor :publication_date
+
+  validate :publication_date_is_valid
+
+  def publication_date_attributes=(attributes)
+    @publication_date = DateForm.new(attributes)
+  end
+
+  def publication_date_is_valid
+    return if publication_date.valid?
+
+    publication_date.errors.each do |error|
+      errors.add("publication_date.#{error.attribute}", error.message)
+    end
   end
 end
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -3,12 +3,7 @@
 # Form for a Work
 class WorkForm < ApplicationForm
   accepts_nested_attributes_for :related_links, :related_works
-
-  def initialize(params = {})
-    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
-    @publication_date = DateForm.new
-    super
-  end
+  accepts_nested_attributes_for :publication_date, form_class: DateForm, array: false
 
   def self.immutable_attributes
     ['druid']
@@ -47,22 +42,5 @@ class WorkForm < ApplicationForm
 
     selected_license = WorkForm.licenses.select { |_, v| v['uri'] == license }
     selected_license.keys.first
-  end
-
-  # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
-  attr_accessor :publication_date
-
-  validate :publication_date_is_valid
-
-  def publication_date_attributes=(attributes)
-    @publication_date = DateForm.new(attributes)
-  end
-
-  def publication_date_is_valid
-    return if publication_date.valid?
-
-    publication_date.errors.each do |error|
-      errors.add("publication_date.#{error.attribute}", error.message)
-    end
   end
 end
--- a/app/javascript/controllers/form_date_controller.js
+++ b/app/javascript/controllers/form_date_controller.js
@@ -1,0 +1,29 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['year', 'month', 'day']
+  
+  connect() {
+    this.update();
+  }
+
+  update() {
+    // Disable month and day fields if year is empty
+    // Disable day field if month is empty
+    this.monthTarget.disabled = false
+    this.dayTarget.disabled = false
+    if (this.yearTarget.value === '') {
+      this.monthTarget.disabled = true
+      this.dayTarget.disabled = true
+    } else if (this.monthTarget.value === '') {
+      this.dayTarget.disabled = true
+    }
+  }
+
+  reset() {
+    this.yearTarget.value = ''
+    this.monthTarget.value = ''
+    this.dayTarget.value = ''
+    this.update()
+  }
+}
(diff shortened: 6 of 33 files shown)


app/forms/work_form.rb
```diff
@@ -4,6 +4,12 @@
 class WorkForm < ApplicationForm
   accepts_nested_attributes_for :related_links
 
+  def initialize(params = {})
+    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
```
mjgiarlo, line 8: Itd be helpful to know how the current nested attrs impl is *not* doing the right thing for publication dates. What all is broken?

app/components/elements/forms/field_component.rb
```diff
@@ -4,18 +4,20 @@ module Elements
   module Forms
     # Base component for all form fields.
     class FieldComponent < ApplicationComponent
-      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, data: {}) # rubocop:disable Metrics/ParameterLists
+      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists
+                     data: {}, **extra_args)
```
mjgiarlo, line 8: This component is looking like it might be getting ripe for refactoring, judging by the growing param list. Does not need to be part of this PR, though.

app/components/elements/forms/select_day_field_component.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Component for form day of the month fields
+    class SelectDayFieldComponent < Forms::FieldComponent
+      def value
+        form.object.public_send(field_name)
```
mjgiarlo, line 8: Does the Rails form magic not do this automagically?

app/components/elements/forms/select_day_field_component.rb
```diff
@@ -0,0 +1,20 @@
...
+      end
+
+      def prefix
+        form.object_name
+      end
```
mjgiarlo, line 13: Can you say what this is doing? This machinery is new to me.

app/components/elements/forms/select_day_field_component.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      def classes
+        merge_classes('form-select', form.object.errors[field_name].present? ? 'is-invalid' : nil)
+      end
```
mjgiarlo, line 17: It looks like currently on `main`, all the `is-invalid` jazz is handled by the invalid feedback component and the tab error controller. Can you say more about why the current invalidity infrastructure doesn't/shouldn't handle this?

app/components/elements/forms/field_component.rb
```diff
@@ -4,18 +4,20 @@ module Elements
...
         @hidden_label = hidden_label
         @label = label
         @help_text = help_text
         @data = data
+        @extra_args = extra_args
         super()
       end
 
-      attr_reader :form, :field_name, :required, :help_text, :hidden_label, :label, :data
+      attr_reader :form, :field_name, :required, :help_text, :hidden_label, :label, :data, :extra_args
```
mjgiarlo, line 20: Can you give examples of what's being slurped into `extra_args` in this PR?

app/controllers/works_controller.rb
```diff
@@ -87,7 +87,9 @@ def wait
   private
 
   def work_params
-    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes])
+    # TODO: Extract nested form handling. See https://github.com/sul-dlss/hungry-hungry-hippo/issues/218
+    params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes] +
+    [{ publication_date_attributes: %i[year month day] }])
```
mjgiarlo, line 92: Depending on your answer to my question above about what's currently broken w/ 1-to-1 nested attrs, I might want to have a go at making those changes before we move forward w/ this PR (*i.e.*, if the problem is small enough that I can hack it out in an hour or two). Reasonable?

app/forms/date_form.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Form object for handling date input
+class DateForm < ApplicationForm
+  attribute :year, :integer
+  validates :year,
+            numericality: { greater_than_or_equal_to: 1000, less_than_or_equal_to: Time.zone.now.year, allow_nil: true }
```
mjgiarlo, line 7: Should we consider constantizing this to give it a name?

app/javascript/controllers/form_date_controller.js
```diff
@@ -0,0 +1,29 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
```
mjgiarlo, line 3: It looks like our other stimulus controllers use two-space indents and I'm seeing four-space indents here. Editor config issue? Or maybe it's time to add JS linting so we don't need to care as much? (I want to add JS linting after I get related works merged.)

app/services/cocina_support.rb
```diff
@@ -57,4 +57,26 @@ def self.update_version_and_lock(cocina_object:, version:, lock:)
   def self.collection_druid_for(cocina_object:)
     cocina_object.structural.isMemberOf.first
   end
+
+  # Returns the event date parsed from an EDTF date value.
+  # @param [Cocina::Models::DRO] cocina_object
+  # @param [String] type, e.g., 'publication'
+  # @return [Hash] with keys for year, month, and day
```
mjgiarlo, line 64: Can also return `NilClass`, it looks like

app/services/to_cocina/work/description_mapper.rb
```diff
@@ -26,11 +26,14 @@ def call
 
       attr_reader :work_form
 
+      delegate :publication_date, to: :work_form
```
mjgiarlo, line 29: Why delegate this method to the work form but not others referenced in this class? Wondering if this inconsistency will cause confusion. At least comment-worthy?

app/services/to_cocina/work/description_mapper.rb
```diff
@@ -54,6 +57,15 @@ def note_params
...
         end.presence
       end
+
+      def event_params
+        [].tap do |params|
+          if publication_date.year.present?
+            params << CocinaDescriptionSupport.event_date(type: 'publication', primary: true,
+                                                          date: publication_date.to_edtf_s)
+          end
+        end.presence
```
mjgiarlo, line 67: I'm wondering if we should prefer the `tap`-based approach or something different, like: 

```suggestion
        return if publication_date.year.blank?

        Array(
          CocinaDescriptionSupport.event_date(type: 'publication', primary: true,
                                              date: publication_date.to_edtf_s)
        ).presence
```

app/views/works/form.html.erb
```diff
@@ -4,6 +4,7 @@
   <% component.with_tab(label: t('works.edit.panes.files.tab_label'), tab_name: :files, selected: true) %>
   <% component.with_tab(label: t('works.edit.panes.title.tab_label'), tab_name: :title) %>
   <% component.with_tab(label: t('works.edit.panes.abstract.tab_label'), tab_name: :abstract) %>
+  <% component.with_tab(label: I18n.t('works.edit.panes.dates.tab_label'), tab_name: :dates) %>
```
mjgiarlo, line 7: I don't think we want to invoke `I18n.t` anywhere in H3 if we can avoid it, because it cuts out the Rails i18n niceness (such as HTML formatting depending on key name). We should use `t()` in views and otherwise use `helpers.t` outside of views (controllers, components).

app/components/elements/forms/select_day_field_component.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Component for form day of the month fields
+    class SelectDayFieldComponent < Forms::FieldComponent
```
mjgiarlo, line 6: We can drop the `Forms::` prefix on components in `app/components/elements/forms/`

app/views/works/form.html.erb
```diff
@@ -35,6 +36,31 @@
       <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :abstract, label: t('works.edit.fields.abstract.label')) %>
     <% end %>
 
+    <% component.with_pane(tab_name: :dates, label: t('works.edit.panes.dates.label'), form:) do %>
+      <% form.fields_for :publication_date do |date_form| %>
+        <fieldset id="publication-date">
+          <legend class="form-label"><%= I18n.t('works.edit.fields.publication_date.label') %></legend>
```
mjgiarlo, line 42: See above re: `I18n.t`

app/views/works/form.html.erb
```diff
@@ -35,6 +36,31 @@
...
+          <div class="row" data-controller="form-date">
+            <div class="col-2">
+              <%= render Elements::Forms::TextFieldComponent.new(form: date_form, field_name: :year, label: I18n.t('works.edit.fields.publication_date.year.label'), hidden_label: true, placeholder: I18n.t('works.edit.fields.publication_date.year.label'), data: { form_date_target: 'year', action: 'form-date#update' }) %>
+            </div>
+            <div class="col-2">
+              <%= render Elements::Forms::SelectMonthFieldComponent.new(form: date_form, field_name: :month, label: I18n.t('works.edit.fields.publication_date.month.label'), hidden_label: true, data: { form_date_target: 'month', action: 'form-date#update' }) %>
+            </div>
+            <div class="col-2">
+              <%= render Elements::Forms::SelectDayFieldComponent.new(form: date_form, field_name: :day, label: I18n.t('works.edit.fields.publication_date.day.label'), hidden_label: true, data: { form_date_target: 'day', action: 'form-date#update' }) %>
```
mjgiarlo, line 51: Same re: `I18n.t`

spec/support/work_mapping_fixtures.rb
```diff
@@ -10,7 +10,8 @@ def new_work_form_fixture
       abstract: abstract_fixture,
       related_links_attributes: related_links_fixture,
       license: license_fixture,
-      collection_druid: collection_druid_fixture
+      collection_druid: collection_druid_fixture,
+      publication_date: DateForm.new(year: '2024', month: '12')
```
mjgiarlo, line 14: Should this be `publication_date_attributes` like the other nested fields?

spec/support/work_mapping_fixtures.rb
```diff
@@ -121,6 +123,7 @@ def request_dro_with_structural_fixture
                             })
   end
 
+  # rubocop:disable Metrics/AbcSize
```
mjgiarlo, line 126: Consider moving to line 127 and dropping the explicit disable on line 152, but this is minor.

spec/system/create_work_deposit_spec.rb
```diff
@@ -65,6 +65,31 @@
     # Filling in abstract
     fill_in('work_abstract', with: abstract_fixture)
 
+    # Clicking on Next to go to dates tab
+    click_link_or_button('Next')
+    expect(page).to have_css('.nav-link.active', text: 'Dates (optional)')
+    expect(page).to have_text('Enter dates related to your deposit')
+
+    # Filling in dates
+    within('fieldset#publication-date') do
```
mjgiarlo, line 74: Would you consider:

```suggestion
    within_fieldset('publication-date') do
```

❓

app/components/elements/forms/field_component.rb
justinlittman: Yeah, I've been stewing over that.

app/components/elements/forms/select_day_field_component.rb
justinlittman: No, `select_day` isn't hooked into the form. See https://guides.rubyonrails.org/form_helpers.html#select-boxes-for-time-or-date-components

mjgiarlo: Form-related "bare methods," TIL 🍻

app/components/elements/forms/select_day_field_component.rb
justinlittman: This is the prefix of the field name, e.g., `work[publication_date_attributes]`. I added a comment.

app/components/elements/forms/select_day_field_component.rb
justinlittman: Yes, because `select_day` isn't bound to a form builder instance. I've added a comment to the top of the class.

app/components/elements/forms/field_component.rb
mjgiarlo: An idea I considered earlier but abandoned (because other things were more important): making the signature `initialize` look like `(form:, field_name:, **options)` and then manually writing getters for options, such as `def hidden_label = @options.fetch(:hidden_label, false)` Not sure if that's better or just different.

app/components/elements/forms/field_component.rb
justinlittman: placeholder

I could be convinced to just add this as an argument.

app/forms/date_form.rb
justinlittman: Can you say more? I'm not sure what you're suggesting.

app/components/elements/forms/field_component.rb
mjgiarlo: Oh, yeah, that strikes me as a reasonable named arg for a form field component.

app/services/cocina_support.rb
justinlittman: Fixed

app/services/cocina_support.rb
```diff
@@ -57,4 +57,26 @@ def self.update_version_and_lock(cocina_object:, version:, lock:)
...
+    cocina_date = event.date.first
+    return unless cocina_date.encoding&.code == 'edtf'
+
+    date = Date.edtf!(cocina_date.value)
+    # Using count of dashes to determine the date parts present
+    dash_count = cocina_date.value.count('-')
+    { year: date.year }.tap do |date_params|
+      date_params[:month] = date.month if dash_count >= 1
+      date_params[:day] = date.day if dash_count == 2
+    end
```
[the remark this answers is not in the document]
justinlittman: I don't think so. I'm fairly sure that would be invalid cocina.

app/services/to_cocina/work/description_mapper.rb
justinlittman: The reasons for which I did it do not justify the inconsistency, so removing delegation.

app/services/to_cocina/work/description_mapper.rb
justinlittman: In the future, we will be adding additional events, some of which are multiples (e.g., multiple deposit dates), for which I don't think that `Array()` will work as well.

app/views/works/form.html.erb
justinlittman: Agree. That was a merge screw up.

app/components/elements/forms/field_component.rb
justinlittman: Named it.

spec/support/work_mapping_fixtures.rb
justinlittman: Could be. I changed it.

app/forms/work_form.rb
mjgiarlo: Nevermind, since you're already doing the extraction! 🎉

app/components/elements/forms/select_day_field_component.rb
mjgiarlo: Nice.

app/controllers/works_controller.rb
mjgiarlo: Once you get past the thick of the one-to-one work, I expect you won't need this anymore because `WorkForm.nested_attributes` will handle it for you.

app/forms/date_form.rb
mjgiarlo: Oh, my bad. I meant:

```
class DateForm < ApplicationForm
  EARLIEST_YEAR = 1000
  # ...
  validates :year,
            numericality: { greater_than_or_equal_to: EARLIEST_YEAR, less_than_or_equal_to: Time.zone.now.year, allow_nil: true }
```

app/services/cocina_support.rb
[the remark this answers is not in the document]
mjgiarlo: Good call.

app/forms/date_form.rb
justinlittman: I'm not a big fan of constantizing things that only appear once. To me it feels like the indirection adds cognitive overhead. (But I don't feel that strongly about this.)

PR #224, opened 2024-12-03
Add merge_classes to NestedComponent to allow stayling differences pe…

aaron-collier:
…r design

The figma designs show that while the contact email (on the title pane) is nested, it does not have the box styling included the same way as the related works content does. This add `hide_label` and `merge_classes` to allow broader reuse of the component.

PR #225, opened 2024-12-04
Add nested contact emails to the work form title tab

### LINKED ISSUE #112: Add contact email field to the title tab of the deposit work form
Both of our early beta users utilize the contact email field (particularly in Acquisitions Serials) as part of their workflow.

aaron-collier:
Closes #112 

Note - this can be merged as is (pending review) but a styling update should happen after #224 to remove the border to more closely match the design.

Nested form:

Screenshot 2024-12-03 at 7 59 32 PM

Show page:

Screenshot 2024-12-03 at 8 04 00 PM

### DIFF
--- a/app/forms/contact_email_form.rb
+++ b/app/forms/contact_email_form.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+# Form for contact emails
+class ContactEmailForm < ApplicationForm
+  attribute :email, :string
+  validates :email, presence: true, format: {
+    with: /\A[^@]+@([a-z0-9-]+\.)+[a-z]{2,4}\z/
+  }
+end
--- a/app/services/cocina_description_support.rb
+++ b/app/services/cocina_description_support.rb
@@ -27,6 +27,21 @@ class CocinaDescriptionSupport
     [{ value: title }]
   end
 
+  def self.contact_emails(contact_emails:)
+    contact_emails.map do |contact_email|
+      # Since contact_email is either a Hash or an instance of ContactEmailForm,
+      # we need to make it a hash of the attributes if it's a ContactEmailForm
+      contact_email = contact_email.attributes if contact_email.respond_to?(:attributes)
+      next if contact_email['email'].blank?
+
+      {
+        value: contact_email['email'],
+        type: 'email',
+        displayLabel: 'Contact'
+      }
+    end.compact_blank
+  end
+
   def self.related_links(related_links:)
     related_links.map do |related_link|
       # NOTE: Sometimes this is an array of hashes and sometimes it's an array of RelatedLinkForm instances
--- a/app/views/works/show.html.erb
+++ b/app/views/works/show.html.erb
@@ -6,8 +6,9 @@
   <%= render Elements::SpinnerComponent.new %>
 <% end %>
 
-<%= render Elements::Tables::TableComponent.new(id: 'title-table', classes: 'table-show mb-5', label: 'Title') do |component| %>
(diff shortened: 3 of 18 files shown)


justinlittman (changes requested):

app/forms/contact_email_form.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Form for contact emails
+class ContactEmailForm < ApplicationForm
+  attribute :email, :string
+  validates :email, presence: true, format: {
+    with: /\A[^@]+@([a-z0-9-]+\.)+[a-z]{2,4}\z/
+  }, if: :deposit?
```
justinlittman, line 8: I wouldn't make this for deposit only. It should be an error for draft as well.

app/services/cocina_description_support.rb
```diff
@@ -27,6 +27,19 @@ def self.title(title:)
     [{ value: title }]
   end
 
+  def self.contact_emails(contact_emails:)
+    contact_emails.map do |contact_email|
+      contact_email = contact_email.attributes if contact_email.respond_to?(:attributes)
```
justinlittman, line 32: Why is this necessary? (Can you put explanation in a code comment.)

app/views/works/show.html.erb
```diff
@@ -6,8 +6,9 @@
   <%= render Elements::SpinnerComponent.new %>
 <% end %>
 
-<%= render Elements::Tables::TableComponent.new(id: 'title-table', classes: 'table-show mb-5', label: 'Title') do |component| %>
+<%= render Elements::Tables::TableComponent.new(id: 'title-table', classes: 'table-work mb-5', label: 'Title and contact') do |component| %>
```
justinlittman, line 9: I don't think the table-show class should have changed.

db/structure.sql
```diff
@@ -158,6 +158,39 @@ CREATE SEQUENCE public.collections_id_seq
 ALTER SEQUENCE public.collections_id_seq OWNED BY public.collections.id;
 
 
+--
```
justinlittman, line 161: Is this db change cruft?

spec/system/show_work_spec.rb
```diff
@@ -145,9 +145,11 @@
 
     # Title table
     within('table#title-table') do
-      expect(page).to have_css('caption', text: 'Title')
+      expect(page).to have_css('caption', text: 'Title and contact')
```
justinlittman, line 148: Is the discrepancy between "Title and contact" and "Title & contact" deliberate?

justinlittman (changes requested):

spec/system/show_work_spec.rb
aaron-collier: I assume you mean between the tab label and the show page? Yes, they are different in the figma designs.

db/structure.sql
aaron-collier: Must be, I'll remove.

app/views/works/show.html.erb
aaron-collier: I don't think that was deliberate, maybe part of rebasing. I'll fix. Thanks for catching this.

app/services/cocina_description_support.rb
aaron-collier: Hmm, this basically directly from the related_links/related_works nested components. I'll see if I can determine if it's necessary or not.

aaron-collier: done.

justinlittman (approved):

PR #232, opened 2024-12-04
[REFACTOR] Support non-repeatable nested components

mjgiarlo:


app/components/elements/forms/select_date_field_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div>
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, label_text: label, hidden_label:) %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= select_day(value, { prompt: label, field_name:, prefix: }, { class: classes, aria: field_aria, data: }) %>
+  <%= select_unit(value, { prompt: label, field_name:, prefix: }, { class: classes, aria: field_aria, data: }) %>
```
mjgiarlo, line 4: Collapse the select day/month stuff into a single reusable date helpers component.

app/components/publication_date/edit_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <%= render Elements::Forms::SelectDateFieldComponent.new(:day, form:, field_name: :day, label: helpers.t('publication_date.edit.fields.day.label'), hidden_label: true,
+                                                                   data: { form_date_target: 'day', action: 'form-date#update' }) %>
+  </div>
+  <div class="col-1"></div>
+  <div class="col-1">
+    <%= render Elements::ButtonComponent.new(classes: 'border border-0', variant: :'outline-primary', size: :lg, data: { action: 'form-date#reset' }) do %>
+      <%= helpers.delete_icon %><span class="visually-hidden">Clear</span>
+    <% end %>
+  </div>
+</div>
```
mjgiarlo, line 20: Trying to keep nested form in dedicated components to keep the form view lean.

app/views/works/form.html.erb
```diff
@@ -37,33 +37,12 @@
     <% end %>
 
     <% component.with_pane(tab_name: :dates, label: t('works.edit.panes.dates.label'), form:) do %>
-      <% form.fields_for :publication_date do |date_form| %>
```
mjgiarlo, line 40: Pull out of the form view. Now uses the non-repeatable nested component HTML with the publication date edit component HTML, with the former being a reusable component.

config/locales/en.yml
```diff
@@ -37,6 +37,16 @@ en:
...
+  publication_date:
+    edit:
+      legend: 'Previous publication date'
+      fields:
+        year:
+          label: 'Year'
+        month:
+          label: 'Month'
+        day:
+          label: 'Day'
```
mjgiarlo, line 49: Trying to insulate nested fields from needing to care about where they're nested.

app/components/elements/forms/select_date_field_component.rb
```diff
@@ -2,10 +2,23 @@
...
-    class SelectMonthFieldComponent < FieldComponent
+    class SelectDateFieldComponent < FieldComponent
+      def initialize(unit, **)
+        @unit = unit
+
+        super(**)
+      end
+
+      def select_unit(...)
+        helpers.public_send(:"select_#{@unit}", ...)
```
justinlittman, line 19: I don't love this. I would never guess or get from the name that there is a single component that can render a month or day.

Would it be clearer to DRY this up with component inheritance, but leave separate components for month and day?

app/services/nested_component_renderer.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Instantiates a nested component based on the arity of the form field
+class NestedComponentRenderer
```
justinlittman, line 4: Presenter? I don't like this as a Service.

justinlittman (commented):
I really like "Support non-repeatable nested attrs in the nested component"

I don't like "Refactor date helper-based select components into a single component with a variable unit." While elegant, I think it would surprise our future selves.

app/components/elements/forms/select_date_field_component.rb
mjgiarlo: @justinlittman 💬 
> I don't love this. I would never guess or get from the name that there is a single component that can render a month or day.

Yeah, I buy that. I'll ditch that refactoring.

> Would it be clearer to DRY this up with component inheritance, but leave separate components for month and day?

Maybe? I don't think it's egregiously repetitive. I'm fine with leaving it be and seeing if a better refactoring is clearer later on.

app/components/elements/forms/non_repeatable_nested_component.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Encapsulates a non-repeatable nested form.
```
justinlittman, line 5: Can you add a comment like: "Don't call this directly, use NestedComponentPresenter"?

app/components/elements/forms/repeatable_nested_component.rb
```diff
@@ -2,8 +2,8 @@
 
 module Elements
   module Forms
-    # Encapsulates a nested form, including adding and removing nested models.
-    class NestedComponent < ApplicationComponent
+    # Encapsulates a repeatable nested form, including adding and removing nested models.
+    class RepeatableNestedComponent < ApplicationComponent
```
justinlittman, line 6: Same here.

app/presenters/nested_component_presenter.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Presents a nested component based on the arity of the form field
```
justinlittman, line 3: Typo

justinlittman (approved):

app/components/elements/forms/non_repeatable_nested_component.rb
mjgiarlo: Good call, yes, done.

PR #240, opened 2024-12-04
Refactor repeatability in nested attributes

mjgiarlo:


app/forms/publication_date_form.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Form object for handling publication date
+class PublicationDateForm < DateForm
+end
```
mjgiarlo, line 5: We can revisit this decision later on. For now, I felt like having inherited forms was less noisy than adding an option to `ActiveModel::NestedAttributes` that isn't present in `ActiveRecord::NestedAttributes` (`form_class:`).

app/forms/work_form.rb
```diff
@@ -2,8 +2,7 @@
 
 # Form for a Work
 class WorkForm < ApplicationForm
-  accepts_nested_attributes_for :related_links, :related_works
-  accepts_nested_attributes_for :publication_date, form_class: DateForm, array: false
+  accepts_nested_attributes_for :related_links, :related_works, :publication_date
```
mjgiarlo, line 5: We can handle publication_date correctly after two changes:

1. Subclassing `DateForm` as `PublicationDateForm`, which obviates the need to specify a form class; and
2. Sniffing the nested attr to determine if it's plural or singular, and deducing from that whether the attr is repeatable or not.

lib/active_model/nested_attributes.rb
```diff
@@ -11,83 +11,84 @@ module NestedAttributes
...
-          # For a :related_links attr, define a custom RelatedLinksType to be used in the `attribute` call below
-          const_set(
-            (options[:form_class] || model).to_s.classify.concat('Type'),
-            Class.new(ActiveModel::Type::Value) do
-              # Defining `#cast` on an ActiveModel type tells it how to deserialize,
-              # e.g., from a hash or an array of hashes to RelatedLinkForm instances
-              define_method(:cast) do |attributes_hash_or_list|
-                form_class = options[:form_class] || model.to_s.classify.concat('Form').constantize
+      def determine_repeatability_of(model)
+        model.to_s.pluralize == model.to_s
```
mjgiarlo, line 36: We can configure `ActiveSupport::Inflector` if this ever becomes a problem.

justinlittman (approved):
Nice. Squeezed out some of the complexity that I added in.

PR #242, opened 2024-12-04
Show status on dashboard page. Refactor table components.

### LINKED ISSUE #241: Show actual status for works on dashboard
As in the status from querying SDR.

justinlittman:
closes #241

image


app/components/component_support/css_classes.rb
```diff
@@ -10,7 +10,7 @@ class CssClasses
     def self.merge(*args)
       args.map do |arg|
         Array(arg&.split)
-      end.flatten.compact.uniq.join(' ')
+      end.flatten.compact.uniq.join(' ').presence
```
justinlittman, line 13: Without this, empty classes were getting rendered: `class=""`

app/components/dashboard/show/works_list_component.html.erb
```diff
@@ -1,19 +1,11 @@
 <%= render Elements::Tables::TableComponent.new(
-      id: 'works-table',
+      id:,
```
justinlittman, line 2: This avoids dupe ids for tables.

app/components/dashboard/show/works_list_component.html.erb
```diff
@@ -1,19 +1,11 @@
...
-    <% component.with_row(
-         values: [
-           link_to(work.title, "/works/#{work.druid}"),
-           work.deposit_job_started? ? 'Depositing' : 'Deposited',
-           work.user.name,
-           I18n.l(work.updated_at, format: '%b %d, %Y'),
-           link_to(nil, Sdr::Purl.from_druid(druid: work.druid))
-         ]
-       ) %>
+    <% component.with_row(values: values_for(work), id: id_for(work)) %>
```
justinlittman, line 9: Moved to within the component class.

app/components/dashboard/show/works_list_component.rb
```diff
@@ -4,14 +4,33 @@ module Dashboard
...
+      end
+
+      def id_for(work)
+        dom_id(work, id)
+      end
+
+      def values_for(work)
+        [
+          link_to(work.title, work_path(work)),
+          @status_map[work.id].status_message,
```
justinlittman, line 28: This is what gets the correct status message.

app/components/dashboard/show/works_list_component.rb
```diff
@@ -4,14 +4,33 @@ module Dashboard
...
+          work.user.name,
+          I18n.l(work.updated_at, format: '%b %d, %Y'),
+          work.druid ? link_to(nil, Sdr::Purl.from_druid(druid: work.druid)) : nil
```
justinlittman, line 31: Also, fixed the Purl to not render when there was no druid yet.

app/components/elements/tables/table_body_component.html.erb
```diff
@@ -1,7 +0,0 @@
-<%= tag.tbody class: classes do %>
```
justinlittman, line 1: Having a separate table body component added unnecessary complexity, especially for testing correctly (since it was passed rows that had been rendered by the table component).

app/components/elements/tables/table_component.html.erb
```diff
@@ -3,5 +3,9 @@
     <caption><%= label %></caption>
   <% end %>
   <%= header %>
-  <%= render Elements::Tables::TableBodyComponent.new(rows: rows) %>
+  <%= tag.tbody class: classes do %>
```
justinlittman, line 6: See, table body component wasn't really that helpful.

app/controllers/dashboard_controller.rb
```diff
@@ -4,5 +4,13 @@
 class DashboardController < ApplicationController
   skip_verify_authorized only: :show
 
-  def show; end
+  def show
+    # This is going to be slow because it makes a call to DSA for each work.
+    # It is also going to block rendering since it is synchronous.
+    # Also, it processes all of a user's works, but not all may be rendered on the dashboard.
```
justinlittman, line 10: This is where the status is retrieved from SDR.

app/presenters/status_presenter.rb
```diff
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
-# Presenter for the status of a work.
+# Presenter for the status of a work or collection.
 # Wraps the status returned from DSA.
 class StatusPresenter
```
justinlittman, line 5: Sweet. We already had a status presenter.

spec/rails_helper.rb
```diff
@@ -92,3 +92,13 @@
 
 # This has to be after FactoryBot include
 require 'cocina/rspec'
+
+# View Components silently ignore within calls.
+# See https://github.com/ViewComponent/view_component/issues/1910
+module ViewComponent
```
justinlittman, line 98: Surprise!

app/components/dashboard/show/works_list_component.rb
mjgiarlo, line 28: I'm fuzzy on this so take it with a grain of salt: could we use ViewComponent's "collections" (or slots) handle these values? Each would be its own component, I guess. I'm still trying to get a handle on what the sweet spot is for structuring our components.

mjgiarlo (approved):

app/components/dashboard/show/works_list_component.rb
justinlittman: Currently we are providing a hash of values for each row to a slot which is used to construct a TableRowComponent.

Alternatively, we could provide an array of hashes of values that would be used by to construct TableRowComponents. This is show here: https://viewcomponent.org/guide/slots.html#rendering-collections

I would consider this a slight stylistic difference, but not really substantive. (I have a minor preference for the way it is in this component because providing an array of hashes seems sorta non-intuitive.)

PR #245, opened 2024-12-05
Use object modified date for last modified.

### LINKED ISSUE #244: Use object update for work last modified on dashboard
That is, when was the object updated in SDR instead of updated_at for Work record.

justinlittman:
closes #244

### DIFF
--- a/db/migrate/20241205013747_add_object_updated_at_to_work.rb
+++ b/db/migrate/20241205013747_add_object_updated_at_to_work.rb
@@ -1,0 +1,8 @@
+class AddObjectUpdatedAtToWork < ActiveRecord::Migration[8.0]
+  def change
+    add_column :works, :object_updated_at, :datetime
+    add_index :works, :object_updated_at
+
+    Work.all.find_each {|work| work.update!(object_updated_at: work.updated_at) }
+  end
+end
--- a/app/components/dashboard/show/works_list_component.rb
+++ b/app/components/dashboard/show/works_list_component.rb
@@ -27,7 +27,7 @@ module Dashboard
           link_to(work.title, link_for(work)),
           @status_map[work.id].status_message,
           work.user.name,
-          I18n.l(work.updated_at, format: '%b %d, %Y'),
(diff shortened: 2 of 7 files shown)


db/migrate/20241205013747_add_object_updated_at_to_work.rb
```diff
@@ -0,0 +1,8 @@
+class AddObjectUpdatedAtToWork < ActiveRecord::Migration[8.0]
+  def change
+    add_column :works, :object_updated_at, :datetime
+    add_index :works, :object_updated_at
+
+    Work.all.find_each {|work| work.update!(object_updated_at: work.updated_at) }
```
mjgiarlo, line 6: Is this a reversible migration with this line in it? We might want to tweak this migration. (I know we seldom roll back migrations in general, so maybe it's no biggie?) https://guides.rubyonrails.org/active_record_migrations.html#using-the-change-method

justinlittman: Rollback works fine:
```
bin/rails db:rollback:primary

== 20241205013747 AddObjectUpdatedAtToWork: reverting =========================
-- remove_index(:works, :object_updated_at)
   -> 0.0112s
-- remove_column(:works, :object_updated_at, :datetime)
   -> 0.0025s
== 20241205013747 AddObjectUpdatedAtToWork: reverted (0.0828s) ================
```

mjgiarlo: @justinlittman If the `:object_updated_at` column is removed, won't the `update!` call raise? Assuming you have works in the DB, of course.

justinlittman: I don't have detailed knowledge about how Rails migrations works; however, it worked locally with works in the database.

mjgiarlo: OK, excellent.

mjgiarlo (approved):

PR #248, opened 2024-12-05
Extract CSS from table components to stylesheet.

justinlittman:
Dashboard before:
image

Dashboard after:
image

Show before:
image

Show after:
image


mjgiarlo (approved):
Good call!

PR #252, opened 2024-12-05
Adds auth for creating a work.

### LINKED ISSUE #156: Add authorization for WorksController new and create
https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/controllers/works_controller.rb#L17
and
https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/controllers/works_controller.rb#L39

justinlittman:
closes #156

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -18,11 +18,11 @@ class WorksController < ApplicationController
   end
 
   def new
-    # Once collection is being passed, should authorize that the user can create a work in that collection.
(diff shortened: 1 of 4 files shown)


mjgiarlo (approved):
What's here looks good. Question: it looks like neither request spec is testing the case where the user is not an admin but is authorized to new/create collections. Is that a thing we want to cover?

justinlittman:
It is covered by the system specs (as it is the happy path).

PR #258, opened 2024-12-05
Use URI::MailTo::EMAIL_REGEX instead of bespoke regex

aaron-collier:
I didn't initially see any helpers like this on the initial PR, but found this used in [H2](https://github.com/sul-dlss/happy-heron/blob/main/app/models/user.rb#L17C15-L17C46) for email validation. We should probably avoid bespoke regex in cases like this.


justinlittman (approved):
Nice!

Issue #211, opened 2024-12-02
Implement work type tab


aaron-collier:

amyehodge:
For MVP1 we don't need all the options available, since she'll be depositing all the same kind of content. But we need to be able label these as 
work type: text
work subtype: poster

PR #261, opened 2024-12-06
Adds components for radio and checkbox collections. Adds work type model.

justinlittman:
refs #211

### DIFF
--- a/app/models/work_type.rb
+++ b/app/models/work_type.rb
@@ -1,0 +1,125 @@
+# frozen_string_literal: true
+
+# Represents the list of valid work types
+class WorkType
+  class InvalidType < StandardError; end
+
+  MINIMUM_REQUIRED_MUSIC_SUBTYPES = 1
+  MINIMUM_REQUIRED_MIXED_MATERIAL_SUBTYPES = 2
+  MIXED_MATERIAL = 'mixed material'
+  MUSIC = 'music'
+  OTHER = 'other'
+
+  DATA_TYPES = [
+    '3D model', 'Database', 'Documentation', 'Geospatial data', 'Image',
+    'Tabular data', 'Text corpus'
+  ].freeze
+
+  VIDEO_TYPES = [
+    'Conference session', 'Documentary', 'Event', 'Oral history', 'Performance'
+  ].freeze
+
+  SOUND_TYPES = ['Interview', 'Oral history', 'Podcast', 'Speech'].freeze
+
+  TEXT_TYPES = [
+    'Article', 'Capstone', 'Government document', 'Policy brief', 'Preprint', 'Report',
+    'Technical report', 'Thesis', 'Working paper'
+  ].freeze
+
+  SOFTWARE_TYPES = %w[Code Documentation Game].freeze
+
+  IMAGE_TYPES = ['CAD', 'Map', 'Photograph', 'Poster', 'Presentation slides'].freeze
+
+  MUSIC_TYPES = [
+    'Data',
+    'Image',
+    'MIDI',
+    'Musical transcription',
+    'Notated music',
+    'Piano roll',
+    'Software/Code',
+    'Sound',
+    'Text',
+    'Video'
+  ].freeze
+
+  # These types appear below the fold and may be expanded
+  MORE_TYPES = [
+    '3D model', 'Animation', 'Article', 'Book', 'Book chapter', 'Broadcast', 'CAD', 'Capstone',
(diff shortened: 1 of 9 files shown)


app/models/work_type.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
```
justinlittman, line 1: This and spec are copied from H2.

app/components/elements/forms/base_collection_component.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Base component for a set of inputs for a collection, e.g. checkboxes or radio buttons
+    class BaseCollectionComponent < ApplicationComponent
```
mjgiarlo, line 6: I hate to ask this, but what is a collection in this context? I ask because `collection` already means something in the H3 domain model. If this means a form element that can have multiple values, I wonder if we might use a different term.

app/models/work_type.rb
```diff
@@ -0,0 +1,127 @@
...
+  def self.purl_reservation_type
+    new(id: 'purl_reservation', label: 'PURL reservation', html_label: 'PURL reservation', icon: '', subtypes: [],
+        cocina_type: Cocina::Models::ObjectType.object)
+  end
+
+  def self.find(id)
+    (all + [purl_reservation_type]).find { |work| work.id == id } || raise(InvalidType, "Unknown worktype #{id}")
+  end
+
+  # rubocop:disable Metrics/AbcSize
```
mjgiarlo, line 86: Consider moving to line 89 and dumping the explicit `rubocop:enable` below

app/models/work_type.rb
```diff
@@ -0,0 +1,127 @@
...
+  # id is a value acceptable for MODS typeOfResource
```
mjgiarlo, line 87: Is this comment still meaningful in the H3 context?

app/components/elements/forms/collection_radio_buttons_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form.collection_radio_buttons field_name, collection, value_method, text_method do |builder| %>
```
mjgiarlo, line 1: Since nothing in this branch is using the new components, how do we know what the useful params are? As a reviewer, it's unclear to me why e.g. `value_method` and `text_method` or any of this is here, because nothing's using any of it. I assume this is because it's lifted from H2 and the follow-up PR will make it clear.

mjgiarlo, line 1: I appreciate referring to form builder instances as `builder` and not `form` to keep that clear! I suspect we have not been consistent about that (I'm likely a guilty party).

app/components/elements/forms/base_collection_component.rb
justinlittman: Open to discussion on this. This is the term from the Rails side of things.

app/models/work_type.rb
justinlittman: Yes, I think it will be. We should be mapping the work types / subtypes in the same way.

app/components/elements/forms/collection_radio_buttons_component.html.erb
justinlittman: Actually, this isn't lifted from H2; it is extracted from my H3 code. (I'm trying the keep the size of the next PR smaller.)

justinlittman: I was just following the rails code (which used "b" but I expanded to "builder").

app/components/elements/forms/base_collection_component.rb
mjgiarlo: @justinlittman Do you have a long-story-short sense of e.g. how `select` and `collection_select` are different?

justinlittman: I'm not sure they are significantly different for selects. For checkboxes and radios, they make building sets of them much easier.

mjgiarlo: @justinlittman OK. How about we use the term `InputCollection` (`input_collection`)? Would that be a net plus, net minus, or net zero in terms of clarity?

justinlittman: I like that.

mjgiarlo (approved):

PR #262, opened 2024-12-06
Limit solargraph to Linux installs

justinlittman:
Erb formatting has been broken in VSCode for weeks, which has been driving me crazy. Finally tracked it down: https://github.com/Shopify/ruby-lsp/issues/2282

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,6 +74,11 @@ end
 
 group :development do
   gem 'overmind'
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
I tossed another commit at this branch that works for me. Can you try it out on your side and see how VSCode looks?

mjgiarlo (approved):

PR #279, opened 2024-12-06
Add and update related work options

### LINKED ISSUE #237: Related works drop-down should have 14 options and "this related work" text needs to be included.
I'm only seeing these, which is about half of them:
Image

For the "this related work" text, in the original mockup we placed this over to the right of the selection box and left it out off the end of the actual selections, but we've decided that it should be included in each of the entries instead. 

This is the mockup:

Image

This is the full list of choices. I've written them with the "this related work" text appended to each entry:
It's a supplement to this related work.
It's supplemented by this related work.
It's referenced or cited by this related work.
It references or cites this related work.
It's derived from this related work.
It's the source of this related work.
The version of record / publisher's version of my deposit is this related work.
It's identical to this related work.
It has specific versions or instances, one of which is this related work.
It's a specific version or instance of this related work.
It succeeds or is a continuation of this related work.
It precedes or is continued by this related work.
It's one part of this related work.
It consists of parts, one of which is this related work.

aaron-collier:
Fixes #237 

This just updates the text and options for the related work options drop down:

Screenshot 2024-12-06 at 2 16 51 PM

### DIFF
--- a/app/forms/related_work_form.rb
+++ b/app/forms/related_work_form.rb
@@ -3,10 +3,16 @@
 # Form for a related work
 class RelatedWorkForm < ApplicationForm
   RELATIONSHIP_TYPES = [
+    'supplement to',
+    'supplemented by',
     'referenced by',
     'references',
-    'has original version',
+    'derived from',
+    'source of',
+    'version of record',
+    'identical to',
     'has version',
+    'has original version',
     'preceded by',
     'succeeded by',
     'part of',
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
I originally left these out because I thought they might not yet have mappings to/from Cocina, but in retrospect, this may have been before I clarified with Arcadia what the mappings should be. Should be good to go.

PR #291, opened 2024-12-09
Adds work type and work subtype to form.

### LINKED ISSUE #211: Implement work type tab

justinlittman:
closes #211


app/forms/work_form.rb
```diff
@@ -42,4 +42,24 @@ def license_label
...
     selected_license.keys.first
   end
+
+  attribute :work_type, :string
+  validates :work_type, presence: true, if: :deposit?
+
+  attribute :work_subtypes, array: true, default: -> { [] }
+  # Music and mixed materials have a minimum number of subtypes required
+  validates :work_subtypes, length: { minimum: WorkType::MINIMUM_REQUIRED_MUSIC_SUBTYPES,
+                                      too_short: '%<count>s term is the minimum allowed' },
```
aaron-collier, line 52: Would this end up being something like `2s term is the minimum allowed` instead of something like `2 terms is the minimum allowed`?

app/forms/work_form.rb
```diff
@@ -42,4 +42,24 @@ def license_label
...
+                            if: lambda {
+                              work_type == WorkType::MUSIC
+                            }
+  validates :work_subtypes, length: { minimum: WorkType::MINIMUM_REQUIRED_MIXED_MATERIAL_SUBTYPES,
+                                      too_short: '%<count>s terms is the minimum allowed' },
```
aaron-collier, line 57: Same question as above.

app/components/elements/forms/base_input_collection_component.rb
```diff
@@ -5,21 +5,26 @@ module Forms
     # Base component for a set of inputs for a collection, e.g. checkboxes or radio buttons
     class BaseInputCollectionComponent < ApplicationComponent
       def initialize(form:, field_name:, input_collection:, value_method: :to_s, text_method: :to_s, # rubocop:disable Metrics/ParameterLists
-                     div_classes: [])
+                     div_classes: [], input_data: {})
```
aaron-collier, line 8: Perhaps nitpicky and not a blocker, would it make sense to move the `rubocop:disable` onto this line or is there a good reason for it to stay on 7?

app/models/work_type.rb
```diff
@@ -63,63 +63,47 @@ class InvalidType < StandardError; end
...
+      new(label: 'Video', subtypes: VIDEO_TYPES, cocina_type: Cocina::Models::ObjectType.object),
+      new(label: 'Music', subtypes: MUSIC_TYPES, cocina_type: Cocina::Models::ObjectType.object),
+      new(label: 'Mixed Materials', subtypes: MIXED_TYPES, cocina_type: Cocina::Models::ObjectType.object),
+      new(label: 'Other', subtypes: [], cocina_type: Cocina::Models::ObjectType.object)
     ]
   end
 
   def self.type_list
-    all.map(&:id).sort
+    all.map(&:label).sort
```
aaron-collier, line 99: not blocker, but would be nice to make sure a test covers this line.

justinlittman: Added.

aaron-collier (approved):

PR #303, opened 2024-12-10
[WIP] Authors tab

lwrubel:


lwrubel:
Will need to be squashed before merging.

astridu:
This looks really good! I am ok with dropping 'term', but will defer to Amy in case there is a specific reason to keep it. 

My only comment is on the trash icon next to the 'Contact email' text field. Could we move it closer and align with with the text field? TY!

app/components/contributors/edit_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+                                                  options: role_type_options) %>
+  <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :person_role,
+                                                       options: PERSON_ROLES, label: t('works.edit.fields.contributors.role.label'),
+                                                       data: { contributors_target: 'selectPersonRole' }) %>
+  <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :organization_role,
+                                                       options: ORGANIZATION_ROLES, label: t('works.edit.fields.contributors.role.label'),
+                                                       data: { contributors_target: 'selectOrganizationRole' }) %>
+
+  <!-- Contributor person name/orcid fields -->
+  <%= tag.div data: { contributors_target: 'personName' } do %>
```
justinlittman, line 14: I personally find using `<%= tag.div data: { contributors_target: 'personName' } do %>` harder to read than `<div data-contribors-target="personName">` when HTML only (and no variable). It makes it clearer where the erb code is.

app/components/contributors/edit_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+                                                  label: t('works.edit.fields.contributors.role_type.label'),
+                                                  options: role_type_options) %>
+  <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :person_role,
+                                                       options: PERSON_ROLES, label: t('works.edit.fields.contributors.role.label'),
+                                                       data: { contributors_target: 'selectPersonRole' }) %>
+  <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :organization_role,
+                                                       options: ORGANIZATION_ROLES, label: t('works.edit.fields.contributors.role.label'),
+                                                       data: { contributors_target: 'selectOrganizationRole' }) %>
+
+  <!-- Contributor person name/orcid fields -->
```
justinlittman, line 13: Use erb comments unless you want this to appear in the HTML.

app/components/elements/forms/toggle_component.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Elements
+  module Forms
+    # Component for a toggle-like radio button group field
+    class ToggleComponent < FieldComponent
+      def initialize(options:, **)
+        @options = options
```
justinlittman, line 8: I don't love building this off options. I'd recommend slots.

app/components/works/show/authors_show_component.rb
```diff
@@ -0,0 +1,57 @@
...
+    # Component for rendering a table of authors on the work show page.
+    class AuthorsShowComponent < ApplicationComponent
+      def initialize(work:)
+        @work = work
+        super()
+      end
+
+      attr_reader :work
+
+      def id
```
justinlittman, line 14: This method isn't used.

app/services/cocina_description_support.rb
```diff
@@ -99,25 +251,72 @@ def self.related_works(related_works:) # rubocop:disable Metrics/AbcSize
     end.compact_blank
   end
 
-  # @param forename [String] the forename of the person
+  def self.contributors(contributors:) # rubocop:disable Metrics/AbcSize
+    count = 0
```
justinlittman, line 255: Use `.with_index` instead of count?

app/services/cocina_support.rb
```diff
@@ -146,4 +169,11 @@ def self.event_date_for(cocina_object:, type:) # rubocop:disable Metrics/AbcSize
       date_params[:day] = date.day if dash_count == 2
     end
   end
+
+  def self.orcid_for(contributor:)
+    orcid = contributor.identifier&.find { |id| id.type == 'ORCID' }&.value
+    return unless orcid.presence
+
+    "https://orcid.org/#{orcid}"
```
justinlittman, line 177: "https://orcid.org" shouldn't be hard-coded all over the place.

app/forms/author_form.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Form for an author contributor
+class AuthorForm < ContributorForm
```
justinlittman, line 4: Is there validation that there must be at least 1 author to deposit?

justinlittman (commented):
Some of these comments can be addressed in follow-up PRs if you'd prefer.

amyehodge:
@lwrubel I also prefer this without "term," so looks good to me.

app/components/contributors/edit_component.html.erb
lwrubel: Done.

app/components/elements/forms/toggle_component.rb
lwrubel: Going to address this in a follow-up PR.

app/services/cocina_support.rb
lwrubel: Added to settings.

justinlittman (approved):

PR #317, opened 2024-12-11
Back keywords with FAST API

mjgiarlo:
- **Add controller and route for proxying FAST API for keywords**
- **Back keywords implementation with FAST API**


.rubocop.yml
```diff
@@ -20,7 +20,6 @@ AllCops:
   Exclude:
     - 'bin/**/*'
     - 'config/environments/*.rb'
-    - 'config/routes.rb'
```
mjgiarlo, line 23: It's not an auto-generated file managed by Rails, so let's lint it.

Gemfile
```diff
@@ -41,20 +41,19 @@ gem 'bootsnap', require: false
...
 gem 'kaminari' # For pagination
 gem 'okcomputer'
 gem 'view_component'
 gem 'whenever', require: false
 
-# DLSS/domain-specific dependencies
-gem 'cocina-models'
-gem 'dor-services-client', '>= 15.2.1'
-gem 'dor-workflow-client', '>= 7.6.1'
-gem 'druid-tools'
```
mjgiarlo, line 56: Use alphabetical order.

app/components/elements/forms/field_component.rb
```diff
@@ -28,7 +28,7 @@ def help_text_id
       end
 
       def field_aria
-        return if @help_text.blank?
+        return if help_text.blank?
```
mjgiarlo, line 31: Already have an `attr_reader` for this property, so use it instead of relying on the instance variable.

app/components/keywords/edit_component.html.erb
```diff
@@ -1 +1,11 @@
-<%= render Elements::Forms::TextFieldComponent.new(form:, field_name: :text, label: helpers.t('keyword.edit.fields.text.label')) %>
```
mjgiarlo, line 1: Moved away from the text field because there's no seam currently to omit the `mb-4` class on the containing `<div>`, which creates a gap between the keyword input field and the dropdown with FAST values.

app/forms/keyword_form.rb
```diff
@@ -4,4 +4,7 @@
 class KeywordForm < ApplicationForm
   attribute :text, :string
   validates :text, presence: true, if: :deposit?
+  attribute :uri, :string
+  validates :uri, url: true, allow_blank: true
```
mjgiarlo, line 8: Verify the URI is a valid URI using a custom validator defined below.

spec/forms/work_form_spec.rb
```diff
@@ -17,7 +17,7 @@
 
     context 'when saving draft with blank work type' do
       it 'is valid' do
-        expect(form.valid?).to be true
+        expect(form).to be_valid
```
mjgiarlo, line 20: Use predicate matchers when possible.

spec/support/stub_fast_connection.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+RSpec.shared_context('with FAST connection') do
```
mjgiarlo, line 3: Put the FAST testing machinery in a shared context so it can be used by the FAST request spec and feature specs.

app/components/keywords/edit_component.html.erb
justinlittman: Can we do a separate ticket / PR to (1) move `mb-4` outside the component and (2) make this a `TextFieldComponent`?

I think we discussed this previously and decided the `mb-4` was probably a bad idea.

app/controllers/fast_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+  def show
+    result = lookups(params.require(:q))
+    @suggestions = result.value_or([])
+    if result.success?
+      return render status: :no_content if @suggestions.empty?
+
+      render status: :ok, layout: false
+    else
+      logger.warn(result.failure)
+      render status: :internal_server_error, html: ''
```
justinlittman, line 18: Are we sure this is what we want to do in this case? Should be we treat as if no content and HB alert?

app/controllers/fast_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  end
+
+  private
+
+  def lookups(query)
```
justinlittman, line 24: What is this method doing? Is it adding value?

app/controllers/fast_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    result = lookup(query)
+    return result if result.failure?
+
+    Success(result.value!)
+  end
+
+  def lookup(query) # rubocop:disable Metrics/AbcSize
```
justinlittman, line 31: Should this be moved into a service?

app/forms/keyword_form.rb
justinlittman: Is there a path by which we would not get a valid URI? What would happen if this failed?

(My general philosophy has been to use validations for validating form input by the user only.)

app/views/fast/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% @suggestions.map do |suggestion| %>
+  <li class="list-group-item" role="option" data-autocomplete-value="<%= suggestion.values.first %>" data-autocomplete-label="<%= suggestion.keys.first %>"><%= html_escape(suggestion.keys.first) %></li>
```
justinlittman, line 2: Can this be `<%= tag.li ...` instead of inlining variables into html attributes?

package.json
```diff
@@ -10,7 +10,8 @@
     "postcss": "^8.4.47",
     "postcss-cli": "^11.0.0",
     "sass": "^1.80.6",
-    "standard": "^17.1.2"
+    "standard": "^17.1.2",
+    "stimulus-autocomplete": "^3.1.0"
```
justinlittman, line 14: Why is this going in package.json?

spec/requests/fast_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    it 'returns status 200 and html with suggestions' do
+      get '/fast', params: { q: query }
+      expect(response).to have_http_status :ok
+
+      suggestions.each do |suggestion|
+        actual_suggestion = suggestion.keys.first
+        uri = suggestion.values.first
+        li_element_html = '<li class="list-group-item" role="option" ' \
+                          "data-autocomplete-value=\"#{uri}\" data-autocomplete-label=\"#{actual_suggestion}\">" \
+                          "#{ERB::Util.html_escape(actual_suggestion)}</li>"
```
justinlittman, line 44: Does this need the html_escape?

spec/services/cocina_support_spec.rb
```diff
@@ -13,6 +13,47 @@
     end
   end
 
+  describe '#keywords_for' do
+    context 'when object has keywords' do
+      let(:cocina_object) do
+        # NOTE: the :dro factory in the cocina-models gem does not have a seam
+        #       for injecting related links, so we do it manually here
```
justinlittman, line 20: "for injecting keywords"?

spec/support/work_fixtures.rb
```diff
@@ -60,10 +60,14 @@ def contact_emails_fixture
...
   [
     {
-      'text' => 'Biology'
+      'text' => 'Biology',
+      'uri' => 'http://id.worldcat.org/fast/832383/',
+      'cocina_type' => 'topic'
     },
     {
-      'text' => 'Chemistry'
+      'text' => 'Chemistry',
```
justinlittman, line 68: Can we create non-FAST keywords? (I think we can.) Does that need to be tested?

app/components/keywords/edit_component.html.erb
mjgiarlo: @justinlittman Good call: #323

app/controllers/fast_controller.rb
mjgiarlo: @justinlittman It's what H2 does 🤷🏻‍♂️

mjgiarlo: Oh, and I think we're logging the OCLC errors to HB in `lookups` below. So I think we're good.

app/forms/keyword_form.rb
mjgiarlo: OCLC data error, I suppose? If it failed, I don't think the user would have any recourse. I'll yank the validator, because I agree w/ your philosophy.

package.json
mjgiarlo: Good question. We don't need it there w/ importmap goodness, yeah?

app/views/fast/show.html.erb
mjgiarlo: You bet.

spec/requests/fast_spec.rb
mjgiarlo: Copypasta from H2. I'll try removing.

spec/services/cocina_support_spec.rb
mjgiarlo: Yes, thank you!

spec/support/work_fixtures.rb
mjgiarlo: Like, free-text ones? If so, yes we can. And sure, I'll change the second keyword here to a free-text one.

justinlittman (approved):

PR #318, opened 2024-12-11
Fix null related link in show page

mjgiarlo:
~~HOLD until #303 is merged~~

This fixes the bug @lwrubel pointed out the other day where the works show view is showing a bogus related link that isn't in the Cocina.

Includes a presenter refactoring and some test tidying.


app/controllers/works_controller.rb
```diff
@@ -6,12 +6,11 @@ class WorksController < ApplicationController
   before_action :check_deposit_job_started, only: %i[show edit]
   before_action :set_work_form_from_cocina, only: %i[show edit]
   before_action :set_content, only: %i[show edit]
-  before_action :set_status, only: %i[show edit]
+  before_action :set_status, only: %i[edit]
```
mjgiarlo, line 9: No longer needed because the new `WorkPresenter` is driving the works show view now.

app/forms/work_form.rb
```diff
@@ -37,14 +37,6 @@ def persisted?
   attribute :license, :string
   validates :license, presence: true, if: :deposit?
 
-  # Shorter license label for the show page
-  def license_label
-    return unless license
-
-    selected_license = WorkForm.licenses.select { |_, v| v['uri'] == license }
-    selected_license.keys.first
```
mjgiarlo, line 45: Move to `WorkPresenter` since it drives the show page.

app/controllers/works_controller.rb
```diff
@@ -6,12 +6,11 @@ class WorksController < ApplicationController
...
 
   def show
     authorize! @work
-    @status_presenter = StatusPresenter.new(status: @status)
-    @purl_link_presenter = PurlLinkPresenter.new(druid: @work.druid)
+    @work_presenter = WorkPresenter.new(work: @work, work_form: @work_form)
```
mjgiarlo, line 13: Now the entire works show page is driven by a single presenter. This is how we use presenters in Argo, IIRC.

app/presenters/work_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+  delegate :collection, :created_at, :druid, :user, to: :work
+  delegate :abstract, :contact_emails_attributes, :keywords_attributes, :license, :other_work_subtype,
+           :related_links_attributes, :related_works_attributes, :title, :version, :work_subtypes, :work_type,
+           to: :work_form
+
+  def purl_link
+    # No druid yet, so there's no PURL link yet either. Work is likely still depositing.
+    return if druid.blank?
+
+    link_to(nil, Sdr::Purl.from_druid(druid:))
```
mjgiarlo, line 23: The first arg used to be `@label` which was never set anywhere. If we need an optional label later, let's add it when we need it.

app/views/works/show.html.erb
```diff
@@ -1,45 +1,45 @@
...
-  <% component.with_row(label: 'Deposit created', values: [l(@work.created_at, format: :long)]) %>
+  <% component.with_row(label: 'Collection', values: [@work_presenter.collection_link]) %>
+  <% component.with_row(label: 'Depositor', values: [@work_presenter.depositor]) %>
+  <% component.with_row(label: 'Version details', values: [@work_presenter.version]) %>
+  <% component.with_row(label: 'Deposit created', values: [@work_presenter.deposited_at]) %>
 <% end %>
 
 <%= render Elements::Tables::TableComponent.new(id: 'title-table', classes: 'table-show mb-5', label: 'Title and contact') do |component| %>
-  <% component.with_row(label: 'Title', values: [@work_form.title]) %>
-  <% component.with_row(label: 'Contact emails', values: [@work_form.contact_emails_attributes.map(&:email).join(', ')]) %>
```
mjgiarlo, line 20: Get logic out of the view.

spec/forms/contact_email_form_spec.rb
```diff
@@ -9,7 +9,7 @@
 
     context 'when nil' do
       it 'is valid' do
-        expect(form.valid?).to be true
+        expect(form).to be_valid
```
mjgiarlo, line 12: Use predicate matchers when available.

app/presenters/work_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def contact_emails
+    contact_emails_attributes.map(&:email).join(', ')
+  end
+
+  def collection_link
+    link_to(collection.title, Rails.application.routes.url_helpers.collection_path(collection))
```
justinlittman, line 31: Do you need "Rails.application.routes.url_helpers" since it is included in ApplicationPresenter?

app/presenters/work_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+    @work = work
+    @work_form = work_form
+    super()
+  end
+
+  attr_reader :work, :work_form
+
+  delegate :status_message, :editable?, to: :status_presenter
+  delegate :collection, :created_at, :druid, :user, to: :work
+  delegate :abstract, :contact_emails_attributes, :keywords_attributes, :license, :other_work_subtype,
```
justinlittman, line 15: Do we want to expose the `*_attributes` outside the presenter?

app/presenters/work_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def publication_date
+    work_form.publication_date.to_s
+  end
+
+  private
+
+  def status
+    Sdr::Repository.status(druid:)
```
justinlittman, line 75: Can we move the API call outside the presenter?

justinlittman (commented):
I'm wondering if instead of creating a presenter that is used for a single page, it wouldn't be more consistent to just create a view component to encapsulate the presentation logic for the page.

app/presenters/work_presenter.rb
mjgiarlo: I was surprised to find that, yes, I did.

app/presenters/work_presenter.rb
mjgiarlo: I don't think so. Most of this can be private.

app/presenters/work_presenter.rb
mjgiarlo: Sure. Where do you think it ought to live?

app/presenters/work_presenter.rb
mjgiarlo: Done.

app/presenters/work_presenter.rb
mjgiarlo: Nevermind. On rebasing, this is no longer necessary.

app/components/works/show/authors_show_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= render Elements::Tables::TableComponent.new(id: 'authors-table', classes: 'table-work mb-5', label: 'Authors') do |component| %>
-  <% component.with_header(headers: headers) %>
-  <% work.authors.each do |author| %>
+  <% component.with_header(headers:) %>
+  <% presenter.authors.each do |author| %>
```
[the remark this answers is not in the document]
mjgiarlo: I waffled on that, myself. I'll make it clearer.

app/components/works/show/authors_show_component.rb
```diff
@@ -4,12 +4,12 @@ module Works
   module Show
     # Component for rendering a table of authors on the work show page.
     class AuthorsShowComponent < ApplicationComponent
-      def initialize(work:)
-        @work = work
+      def initialize(presenter:)
```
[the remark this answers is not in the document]
mjgiarlo: Fixed

justinlittman (approved):

PR #320, opened 2024-12-11
Adjust collection policy.

justinlittman:
So that a collection's user can view, edit, etc. the collection.

### DIFF
--- a/app/policies/collection_policy.rb
+++ b/app/policies/collection_policy.rb
@@ -1,13 +1,26 @@
 # frozen_string_literal: true
 
 class CollectionPolicy < ApplicationPolicy
-  pre_check :collection_creator?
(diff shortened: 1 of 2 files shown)


lwrubel (approved):

app/policies/collection_policy.rb
```diff
@@ -1,13 +1,26 @@
...
+    record.user_id == user.id
+  end
+
+  def new?
+    collection_creator?
+  end
+
+  def create?
+    collection_creator?
+  end
```
mjgiarlo, line 16: @justinlittman I know this was merged already, but I have a question for my own ActionPolicy learning: is it customary to have a policy expressed as both an alias and in its own method? It looks like both `new?` and `create?` are in both spots here. Naïvely, I might expect the explicit `def` methods to take precedence over whatever `alias_rule` does but maybe AP is smarter than the average bear.

PR #321, opened 2024-12-11
Support discarding a draft.

### LINKED ISSUE #306: Discard draft for works

justinlittman:
closes #306

https://github.com/user-attachments/assets/b750a815-e6cc-4582-a74e-b53b41413857


app/controllers/collections_controller.rb
```diff
@@ -9,7 +9,6 @@ class CollectionsController < ApplicationController
 
   def show
     authorize! @collection
-    @status_presenter = StatusPresenter.new(status: @status)
```
justinlittman, line 12: StatusPresenter is being changed to VersionStatus model. VersionStatus is being returned directly be Sdr::Repository.status().

This is many of the changes below.

app/controllers/dashboard_controller.rb
```diff
@@ -10,9 +10,9 @@ def show
     # Also, it processes all of a user's works, but not all may be rendered on the dashboard.
     @draft_works = []
     @status_map = current_user.works.to_h do |work|
-      status = work.druid ? Sdr::Repository.status(druid: work.druid) : nil
-      @draft_works << work if status&.open?
-      [work.id, StatusPresenter.new(status: status)]
+      version_status = work.druid ? Sdr::Repository.status(druid: work.druid) : VersionStatus::NilStatus.new
```
justinlittman, line 13: Using a special NilStatus object when the object is not yet registered instead of handling that special case within VersionStatus.

mjgiarlo: @justinlittman What would you think about having `Sdr::Repository` being responsible for handing back the `NilStatus` instead of the controller switching on what it gets back?

mjgiarlo (approved):
Looks good. Have a question about who's responsible for `NilStatus`.

app/controllers/dashboard_controller.rb
justinlittman: That's what I originally did, but realized that in every case but this one, no druid is an error (and Sdr::Repository.status detects and raises on).

That being said, open to discussion.

PR #336, opened 2024-12-12
Add Citation tab & entry

### LINKED ISSUE #115: Add citation tab and field to the work deposit form
[See figma design](https://www.figma.com/proto/g46Nfl5EHG7DfwqupK3G4X/H2-2024?page-id=1092%3A2227&node-id=1116-4747&node-type=frame&viewport=-5408%2C-4503%2C1&t=qNDDrcYDWKGfvw1F-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=1114%3A442)

Do not auto-generate citation, the option for custom citation will enable to entry text area and any content entered into the form will be stored in cocina, however we will NOT store the citation in cocina when using the auto-generated citation.

aaron-collier:
Fixes #115 

I did not extract the radio buttons into components in this PR as the use case is specific to citation, but could be happy to refactor an extraction as a follow up PR.

Note: The screen recording is about 4-5 minutes, but you can ignore the last minute of me trying to figure out how to stop the recording.

https://github.com/user-attachments/assets/b9a13db9-98fa-4734-84d9-705aed659955


app/components/works/edit/citation_component.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <p><%= t('works.edit.fields.citation.help_text') %></p>
+
+  <%= tag.div class: 'form-check mb-4' do %>
+    <%= form.radio_button :auto_generate_citation, true, class: 'form-check-input', checked: true, data: { citation_target: 'autoGenerateCitation', action: 'click->citation#autoGenerateCitationSelected' } %>
+    <%= form.label :auto_generate_citation_true, t('works.edit.fields.citation.auto_citation'), class: 'form-check-label' %>
+  <% end %>
+
+  <%= tag.div class: 'form-check mb-4' do %>
+    <%= form.radio_button :auto_generate_citation, false, class: 'form-check-input', data: { citation_target: 'customCitation', action: 'click->citation#customCitationSelected' } %>
+    <%= form.label :auto_generate_citation_false, t('works.edit.fields.citation.custom_citation') %>
```
mjgiarlo, line 11: Does this not get the `form-label` class before its associated input isn't checked?

app/components/works/edit/citation_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= tag.div data: { controller: 'citation' } do %>
+  <p><%= t('works.edit.fields.citation.help_text') %></p>
+
+  <%= tag.div class: 'form-check mb-4' do %>
+    <%= form.radio_button :auto_generate_citation, true, class: 'form-check-input', checked: true, data: { citation_target: 'autoGenerateCitation', action: 'click->citation#autoGenerateCitationSelected' } %>
```
mjgiarlo, line 5: For the data-actions here, I might prefer name that is more descriptive of what they do (vs. some reflection of form state). E.g., `disableCustomCitation` & `enableCustomCitation` or some such?

app/javascript/controllers/citation_controller.js
```diff
@@ -0,0 +1,21 @@
...
+export default class extends Controller {
+  static targets = ['autoGenerateCitation', 'customCitation', 'citation']
+
+  connect () {
+    if (this.autoGenerateCitationTarget.checked) {
+      this.autoGenerateCitationSelected()
+    } else {
+      this.customCitationSelected()
+    }
+  }
```
mjgiarlo, line 12: Do you think we need the `customCitation` target? It doesn't seem to be used.

app/services/cocina_support.rb
```diff
@@ -88,6 +88,10 @@ def self.abstract_for(cocina_object:)
     cocina_object.description.note.find { |note| note.type == 'abstract' }&.value
   end
 
+  def self.citation_for(cocina_object:)
+    cocina_object.description.note.find { |note| note.type == 'preferred citation' }&.value
```
mjgiarlo, line 92: Do we not use the label (from the `CocinaDescriptionSupport` change above) when we map back from Cocina?

app/services/to_cocina/work/description_mapper.rb
```diff
@@ -45,14 +45,21 @@ def params
       end
       # rubocop:enable Metrics/AbcSize
 
+      # rubocop:disable Metrics/AbcSize
```
mjgiarlo, line 48: If you move this after `def note_params` on the next line, you can remove the explicit re-enable on line 62.

mjgiarlo (approved):

PR #342, opened 2024-12-12
Update related links and content sections to match style of design

### LINKED ISSUE #284: Related content page - viz design refinements
1. Text field, drop down, check box border color is 40% black
2. Text field/drop down width is 550px
3. Trash icon should be 16px.

aaron-collier:
Fixes #284 

Screenshot 2024-12-12 at 3 05 18 PM

### DIFF
--- a/app/components/elements/forms/checkbox_component.html.erb
+++ b/app/components/elements/forms/checkbox_component.html.erb
@@ -1,6 +1,6 @@
 <div class="form-check mb-4">
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, label_text: label, hidden_label:, default_label_class: 'form-check-label') %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.checkbox field_name, class: 'form-check-input', required:, aria: field_aria, data: %>
+  <%= form.checkbox field_name, class: 'form-check-input', required:, aria: field_aria, data:, style: styles %>
   <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name:) %>
 </div>
--- a/app/components/elements/forms/select_field_component.html.erb
+++ b/app/components/elements/forms/select_field_component.html.erb
@@ -1,6 +1,6 @@
 <%= tag.div class: 'mb-4', data: do %>
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, label_text: label, hidden_label:) %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.select field_name, options, { prompt: }, { class: 'form-select', aria: field_aria } %>
+  <%= form.select field_name, options, { prompt: }, { class: 'form-select', aria: field_aria, style: styles } %>
   <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name:) %>
 <% end %>
--- a/app/components/elements/forms/select_field_component.rb
+++ b/app/components/elements/forms/select_field_component.rb
@@ -3,32 +3,14 @@
 module Elements
   module Forms
     # Component for form select fields
-    class SelectFieldComponent < ApplicationComponent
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -98,3 +98,7 @@ table.table-show {
 .form-control {
   --bs-border-color: #979797;
 }
+
+.form-check-input {
+  padding: 0
```
aaron-collier, line 103: This was necessary in order to use the `form-control` class to get the correct border on the checkbox component but not increase it's padding.

app/components/elements/forms/select_field_component.rb
```diff
@@ -3,32 +3,14 @@
 module Elements
   module Forms
     # Component for form select fields
-    class SelectFieldComponent < ApplicationComponent
-      def initialize(form:, field_name:, options:, required: false, hidden_label: false, # rubocop:disable Metrics/ParameterLists
-                     label: nil, help_text: nil, prompt: false, data: {})
-        @form = form
-        @field_name = field_name
+    class SelectFieldComponent < FieldComponent
```
aaron-collier, line 6: The removes the need for duplicate logic in this class.

app/assets/stylesheets/application.bootstrap.scss
mjgiarlo: If this change is only for the checkbox component, I wonder if adding the `p-0` class there would work OK without having side effects.

app/components/elements/forms/select_field_component.rb
```diff
@@ -3,32 +3,14 @@
...
+      def initialize(options:, prompt: nil, **args)
```
mjgiarlo, line 7: Could we use anonymous forwarding (`...`) instead of `**args` on lines 7 and 10?

mjgiarlo (approved):
Approve pending consideration of two fairly minor points

app/components/elements/forms/checkbox_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="form-check mb-4">
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, label_text: label, hidden_label:, default_label_class: 'form-check-label') %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.checkbox field_name, class: 'form-check-input', required:, aria: field_aria, data: %>
+  <%= form.checkbox field_name, class: 'form-control form-check-input', required:, aria: field_aria, data:, style: styles %>
```
justinlittman, line 4: Using both form-control and form-check-input doesn't seem like correct BS.

app/components/elements/forms/select_field_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= tag.div class: 'mb-4', data: do %>
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, label_text: label, hidden_label:) %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.select field_name, options, { prompt: }, { class: 'form-select', aria: field_aria } %>
+  <%= form.select field_name, options, { prompt: }, { class: 'form-control form-select', aria: field_aria, style: styles } %>
```
justinlittman, line 4: Using both form-control and form-select doesn't seem like correct BS.

app/components/elements/forms/checkbox_component.html.erb
aaron-collier: Yeah, I was thinking about this and realized it was a bit of an anti-pattern and that I should be setting the border on the `form-check-input` instead of using `form-control` to get the border but then have to override padding. Fixed.

app/assets/stylesheets/application.bootstrap.scss
aaron-collier: See comment below. This was about the border, so an odd changed... I updated it.

app/components/elements/forms/select_field_component.html.erb
aaron-collier: Same as above, fixed.

PR #343, opened 2024-12-12
Fixes styling of abstract entry per designs

### LINKED ISSUE #282: Abstract and keywords page - viz design refinements
1. missing info icon next to field label
2. text field width should be 700px
3. text field border color is 40% black
4. width of the instructions under the page title should also be 700px (makes it easier to read if the width isn't too wide)

aaron-collier:
Fixes #282 

Screenshot 2024-12-12 at 3 27 00 PM

### DIFF
--- a/app/views/works/form.html.erb
+++ b/app/views/works/form.html.erb
@@ -41,8 +41,8 @@
     <% end %>
 
     <% component.with_pane(tab_name: :abstract, label: t('works.edit.panes.abstract.label'), form:) do %>
-      <p><%= t('works.edit.fields.abstract.help_text') %></p>
-      <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :abstract, label: t('works.edit.fields.abstract.label')) %>
+      <p style="width: 700px;"><%= t('works.edit.fields.abstract.help_text') %></p>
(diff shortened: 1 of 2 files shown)


app/views/works/form.html.erb
```diff
@@ -41,8 +41,8 @@
     <% end %>
 
     <% component.with_pane(tab_name: :abstract, label: t('works.edit.panes.abstract.label'), form:) do %>
-      <p><%= t('works.edit.fields.abstract.help_text') %></p>
-      <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :abstract, label: t('works.edit.fields.abstract.label')) %>
+      <p style="width: 700px;"><%= t('works.edit.fields.abstract.help_text') %></p>
```
mjgiarlo, line 44: If all help texts will use the same width, I wonder if we might want, e.g., a `help-text` class? (If not, ignore!)

mjgiarlo (approved):
Question about generalizing help text styles but otherwise good to go

app/views/works/form.html.erb
justinlittman: I don't understand why we're hard coding widths all over the place. If we're concerned about the readability, we should limit the width of the page, not individual text.

aaron-collier: I can back out the widths and we can discuss.

PR #345, opened 2024-12-13
Update styling of the left nav to match figma designs

### LINKED ISSUE #277: Left nav visual design refinements
1. width should be 246 pixels
2. There should be about twice as much padding as there is between the selected dark box and the light gray background
On stage: 
image

Mockup:
image

3. Font size is 18
4. There should be a white line inbetween each of the links (215px wide). No white line appears above or below the selected link.
5. There should be a hover state on the unselected links. An underline should appear when user hovers over them.

aaron-collier:
Fixes #277 

Applies a few styling updates to match the requirements of the design (noted in the ticket). This screenshot shows the underline over a hover as well as improved spacing of the invalid error indicator.

Screenshot 2024-12-12 at 4 08 35 PM

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -17,41 +17,48 @@ $container-max-widths: (
   .nav-pills {
     background-color: var(--stanford-10-black);
     border-radius: var(--bs-nav-pills-border-radius);
+    width: 246px;
+    font-size: 18px;
 
-    .nav-link:not(.active) {
+    .nav-link {
       color: var(--stanford-black);
-      border-bottom: 1px solid var(--stanford-10-black);
-      border-radius: unset;
-    }
-
-    .nav-link.active {
-      font-weight: 700;
-      color: var(--stanford-10-black);
-      background-color: var(--stanford-black);
-
-      &::before {
-        display: inline-block;
-        margin-right: 6px;
-        font-family: bootstrap-icons;
-        vertical-align: -10%;
-        content: "\F138"; // right arrow
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -17,41 +17,48 @@ $container-max-widths: (
   .nav-pills {
     background-color: var(--stanford-10-black);
     border-radius: var(--bs-nav-pills-border-radius);
+    width: 246px;
```
mjgiarlo, line 20: Will using hard-coded widths cause issues for smaller viewports? I don't think too many users will be depositing files via their smartphones, but I guess the question, generally, is how much do we care about responsive design?

mjgiarlo (approved):
Looks good as long as the hard-coded widths isn't going to be problematic (see questions).

app/assets/stylesheets/application.bootstrap.scss
justinlittman: @astridu ?

astridu: yes, we care about responsive designs. I don't think it's a primary use case, but I'm sure someone is going to try to deposit on their tablet or phone. Is there a way to specify a max width for desktop without causing problems for the responsive designs?

justinlittman: The responsive design is borked.

image

I think what you would want to do is pick an appropriate number of columns at each breakpoint that gives you an appropriately sized left nav. However, the left nav doesn't make sense once it has gotten small enough to stack, so we'll need some other sort of a nav (say, a hamburger).

PR #365, opened 2024-12-16
Broadcast file updates only when completed.

justinlittman:
This makes things zippier when uploading large number of files.


app/javascript/controllers/dropzone_controller.js
```diff
@@ -37,12 +37,13 @@ export default class extends Controller {
         const path = file.fullPath ? file.fullPath : file.name
         data.append(`content[paths][${i}]`, path)
       }
+      data.append('content[completed]', this.dropzone.getActiveFiles().length === files.length)
     })
   }
 
-  updateProgress (progress) {
+  updateProgress (progress, reset = false) {
```
justinlittman, line 44: This fixes a bug in which the progress wasn't being reset for >= 2nd uploads.

mjgiarlo (approved):

Issue #364, opened 2024-12-16
Spike on refactoring CocinaSupport and CocinaDescriptionSupport


lwrubel:
Make an initial attempt to refactor cocina-handling code into smaller classes. This can be a partial refactoring to test out the approach. See discussion notes:

https://docs.google.com/document/d/1GpdojGnYZJmwHCDF9V7qnBKvgFeNqRfSOSI5jo9EoT0/edit?tab=t.0#heading=h.4clo4vrszgzd

PR #377, opened 2024-12-17
Collapse ContributorForm into AuthorForm

mjgiarlo:
Connects to #364


app/components/works/show/authors_show_component.rb
```diff
@@ -4,12 +4,12 @@ module Works
   module Show
     # Component for rendering a table of authors on the work show page.
     class AuthorsShowComponent < ApplicationComponent
-      def initialize(work_presenter:)
-        @work_presenter = work_presenter
+      def initialize(authors:)
```
mjgiarlo, line 7: Small tweak: I think the authors show component cares more about the authors (from the presenter) than the presenter itself, so inject them directly and limit knowledge of the presenter to the works controller & show template.

app/forms/author_form.rb
```diff
@@ -1,19 +1,56 @@
...
+
   attribute :role_type, :string
 
-  def organization?
-    role_type == 'organization'
+  attribute :with_orcid, :boolean, default: false
+  attribute :orcid, :string, default: nil
+  validates :orcid, format: { with: /\A\d{4}-\d{4}-\d{4}-\d{3}[0-9X]\z/ },
+                    allow_blank: true,
+                    if: :person?
```
mjgiarlo, line 28: This is the bulk of the work in this PR: collapsing the contributor form into the author form. My reasoning is that 1) it makes working with the author form (e.g. mapping it) easier because it's self-contained, and 2) the contributor form isn't used yet so the boundary between these two forms is fuzzy. If we can defer doing something until when we know more or know better, I'd advocate for taking that just-in-time vs. just-in-case approach.

app/forms/author_form.rb
```diff
@@ -1,19 +1,56 @@
...
+
+    first_name.present? && last_name.present?
+  end
+
+  def organization?(with_names: false)
+    return false if role_type != 'organization'
+    return true unless with_names
+
+    organization_name.present?
+  end
```
mjgiarlo, line 42: I thought it was reasonable to ask the form to take on more responsibility for itself. This means the contributors mapper needs less smarts about this form.

justinlittman (approved):

PR #381, opened 2024-12-18
Add support for discarding collections.

### LINKED ISSUE #86: Add ability to discard drafts for collections

justinlittman:
closes #86

https://github.com/user-attachments/assets/c1ee43cc-86c2-4a89-be26-ad02ee8e4639

### DIFF
--- a/app/components/show/header_component.html.erb
+++ b/app/components/show/header_component.html.erb
@@ -1,0 +1,10 @@
+<div class="d-flex justify-content-between">
+  <div>
+    <%= render Elements::HeaderComponent.new(level: :h1, text: title, classes: 'mb-4 d-inline-block') %>
+    <span class="ms-4 fst-italic status"><%= status_message %></span>
+  </div>
+  <div class="d-flex align-items-start">
+    <%= render Elements::ButtonLinkComponent.new(link: edit_path, label: 'Edit or deposit', variant: :'outline-primary') if editable? %>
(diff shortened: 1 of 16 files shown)


app/components/show/header_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="d-flex justify-content-between">
```
justinlittman, line 1: Extracted show page header into component.

config/locales/en.yml
```diff
@@ -224,8 +224,9 @@ en:
             User agrees that, where applicable, content will not be used to identify or to otherwise infringe the privacy or
             confidentiality rights of individuals. Content distributed via the Stanford Digital Repository may be subject to
             additional license and use restrictions applied by the depositor.
-      errors:
+      messages:
```
justinlittman, line 227: Renamed to be consistent with works. (The messages are not all errors.)

app/components/show/header_component.html.erb
mjgiarlo: @justinlittman Is the idea that this header component will be used for both work show and collection show?

mjgiarlo (approved):

app/components/show/header_component.html.erb
justinlittman: Yes, I plugged it into both. (I decided to extract after realizing I was copying from work show to collection show ...)

PR #385, opened 2024-12-18
Implement collection participants (Managers and Depositors)

### LINKED ISSUE #210: Implement different kinds of users associated with a collection
- Depositors
- Managers
- Reviewers

### LINKED ISSUE #221: Add collection participants tab and input
This will include adding the `managers` and `depositors` fields on the form, as well as active record relationships for collection_managers and collection_depositors (see the H2 join tables, etc., for the implementation there.)

aaron-collier:
Fixes #210
Fixes #221 

https://github.com/user-attachments/assets/8e25d6c4-ab2e-4d58-b18d-23ff19225db7


app/services/cocina_support.rb
```diff
@@ -141,10 +141,12 @@ def self.clean(obj)
   def self.update_version_and_lock(cocina_object:, version:, lock:)
     params = Cocina::Models.without_metadata(cocina_object).to_h
     params['version'] = version
-    params[:structural][:contains].each do |file_set_params|
-      file_set_params[:version] = version
-      file_set_params[:structural][:contains].each do |file_params|
-        file_params[:version] = version
+    if params[:structural].present?
```
aaron-collier, line 144: This was necessary as editing a collection fails (at least locally) with an error related to not having structural cocina.

db/migrate/20241206021849_drop_not_null_from_users.rb
```diff
@@ -0,0 +1,6 @@
+class DropNotNullFromUsers < ActiveRecord::Migration[8.0]
+  def change
+    change_column_null :users, :name, true
```
aaron-collier, line 3: Since we might be creating a user record with only an email address if they are added as a collection participant before they've ever logged in, their names will be null.

db/migrate/20241205222413_add_depositors_and_managers_join_tables.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_join_table :collections, :users, table_name: :depositors do |t|
+      t.index [:collection_id, :user_id], unique: true
+    end
+
+    create_join_table :collections, :users, table_name: :managers do |t|
+      t.index [:collection_id, :user_id], unique: true
+    end
+
+    create_join_table :collections, :users, table_name: :reviewers do |t|
```
aaron-collier, line 11: Just a note that I felt like it made sense to make the reviewers join table now so that it's in place when we add the review workflow/reviewers, but I can remove if we don't want to wait on including this.

app/components/collections/contributor_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Collections
+  # Component for editing collection contributors
+  class ContributorComponent < ApplicationComponent
```
justinlittman, line 5: These don't seem necessary to componentize. And if you do, they can just be be subclasses of TextFieldComponent instead of rendering a TextFieldComponent.

app/services/to_collection_form/mapper.rb
```diff
@@ -27,8 +28,26 @@ def params
...
+        depositors_attributes:,
         version: cocina_object.version
       }
     end
+
+    def managers_attributes
+      return [] if collection&.managers.blank?
+
+      collection.managers.map do |manager|
+        { sunetid: manager.email_address.delete_suffix('@stanford.edu') }
```
justinlittman, line 41: Add sunetid generation to model / form / other?

app/forms/manager_form.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Form for collection managers
+class ManagerForm < ApplicationForm
```
justinlittman, line 4: If this form is exactly the same as the DepitorForm can we use only one form?

app/jobs/deposit_collection_job.rb
```diff
@@ -33,4 +35,22 @@ def perform_persist
       Sdr::Repository.register(cocina_object:)
     end
   end
+
+  def set_managers
+    collection.managers.clear
+    collection_form.managers_attributes.each do |manager|
+      manager = manager.attributes if manager.respond_to?(:attributes)
+      user = User.find_or_create_by(email_address: "#{manager['sunetid']}@stanford.edu")
+      collection.managers.append(user)
```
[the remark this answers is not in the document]
aaron-collier: Yes. I verified this by stopping my local `bin/dev` and restarting the docker stack and all previous assignments persisted.

app/jobs/deposit_collection_job.rb
```diff
@@ -11,6 +11,8 @@ def perform(collection_form:, collection:, deposit:)
 
     new_cocina_object = perform_persist
     druid = new_cocina_object.externalIdentifier
+    set_managers
```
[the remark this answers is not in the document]
aaron-collier: This should now be covered.

app/forms/depositor_form.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Form for collection depositors
+class DepositorForm < CollectionParticipantForm
```
aaron-collier, line 4: Note that the nested attribute components require a form that matches the field name, so the parent `CollectionContributorForm` is used to avoid duplication in `DepositorForm` and `ManagerForm`. This will likely also be used in a future `ReviewerForm`.

app/views/dashboard/show.html.erb
```diff
@@ -19,3 +19,19 @@
     <%= render Elements::ButtonLinkComponent.new(link: new_work_path(collection_druid: collection.druid), label: 'Deposit to this collection', variant: :primary) %>
   </div>
 <% end %>
+
+<% current_user.manages_collections.each do |collection| %>
```
aaron-collier, line 23: In H2, the current interface includes managed and deposit into collections on the dashboard with no distinction between them and the current users owned collections. This is simply copying that format to our current dashboard until any design refinement is complete.

app/forms/depositor_form.rb
justinlittman: I'd be inclined to duplicate since DepositorForm isn't logically a child of CollectionParticipantForm.

app/jobs/deposit_collection_job.rb
```diff
@@ -33,4 +35,32 @@ def perform_persist
...
+      # Sometimes this is a hash and sometimes it is a CollectionParticipantForm object.
+      manager = manager.attributes if manager.respond_to?(:attributes)
+      next if manager['sunetid'].blank?
+
+      user = User.find_or_create_by(email_address: sunetid_to_email_address(manager['sunetid']))
+      collection.managers.append(user)
+    end
+  end
+
+  def set_depositors
```
justinlittman, line 51: Wondering if set_depositors and set_managers could be DRYed.

app/policies/work_policy.rb
```diff
@@ -1,9 +1,33 @@
...
+  def collection_manager?
+    return record.managers.include?(user) if record.is_a? Collection
+
+    record.collection.managers.include?(user)
+  end
+
+  def collection_depositor?
+    return record.depositors.include?(user) if record.is_a? Collection
+
+    record.collection.depositors.include?(user)
```
justinlittman, line 27: Untested

app/services/to_collection_form/mapper.rb
```diff
@@ -27,8 +28,26 @@ def params
...
         contact_emails_attributes: CocinaSupport.contact_emails_for(cocina_object:),
         related_links_attributes: CocinaSupport.related_links_for(cocina_object:),
+        managers_attributes:,
+        depositors_attributes:,
         version: cocina_object.version
       }
     end
+
+    def managers_attributes
+      return [] if collection&.managers.blank?
```
justinlittman, line 38: Is this protection necessary? Is collection ever nil? Could this just be `return [] if collection.nil?`?

app/services/to_collection_form/mapper.rb
```diff
@@ -27,8 +28,26 @@ def params
...
+
+      collection.managers.map do |manager|
+        { sunetid: manager.sunetid }
+      end
+    end
+
+    def depositors_attributes
+      return [] if collection&.depositors.blank?
```
justinlittman, line 46: Same here.

app/views/collections/show.html.erb
```diff
@@ -15,6 +15,11 @@
   <% component.with_row(label: 'Contact emails', values: [@collection_presenter.contact_emails]) %>
 <% end %>
 
+<%= render Elements::Tables::TableComponent.new(id: 'participants-table', classes: 'table-show mb-5', label: 'Collection participants') do |component| %>
+  <% component.with_row(label: 'Managers', values: [@collection.managers.pluck(:email_address).join(', ')]) %>
```
justinlittman, line 19: Should `@collection.managers.pluck(:email_address).join(', ')` be in presenter?

db/migrate/20241206021849_drop_not_null_from_users.rb
justinlittman: Should we just put the sunetid in as the name so there is some value?

spec/system/create_work_deposit_spec.rb
```diff
@@ -29,140 +29,183 @@
...
-      select('10', from: 'Day')
-      # Clear
-      click_link_or_button('Clear')
-      expect(page).to have_field('Year', disabled: false, text: '')
-      expect(page).to have_field('Month', disabled: true)
-      expect(page).to have_field('Day', disabled: true)
-      fill_in('Year', with: '2024')
-      select('June', from: 'Month')
-      select('10', from: 'Day')
+  context 'when the user is a collection manager' do
```
justinlittman, line 176: Can this be done as a request spec?

spec/system/edit_collection_spec.rb
```diff
@@ -83,13 +83,22 @@
       find('button[data-action="click->nested-form#delete"]').click
     end
 
+    # Clicking on Next to go to Participants tab
+    click_link_or_button('Next')
+    expect(page).to have_css('.nav-link.active', text: 'Participants')
+    expect(page).to have_text('Managers')
+    fill_in('collection_managers_attributes_0_sunetid', with: 'stepking')
```
justinlittman, line 90: I know we have done this elsewhere, but can we not use the id here?

db/migrate/20241206021849_drop_not_null_from_users.rb
aaron-collier: That seems a bit confusing to me, but I'm happy to do so if we'd feel better with a value in there rather than allowing null.

app/services/to_collection_form/mapper.rb
aaron-collier: Actually, I don't think so. This appears to be another case where the collection being nil was only a test artifact and easily fixed there.

db/migrate/20241206021849_drop_not_null_from_users.rb
justinlittman: You can image going to render a list of users and the blanks causing problems.

app/jobs/deposit_collection_job.rb
```diff
@@ -33,4 +36,19 @@ def perform_persist
       Sdr::Repository.register(cocina_object:)
     end
   end
+
+  def assign_participants(role)
```
justinlittman, line 40: Can you add some code comments to this method?

app/models/user.rb
```diff
@@ -4,4 +4,13 @@
 class User < ApplicationRecord
   has_many :works, dependent: :destroy
   has_many :collections, dependent: :destroy
+
+  has_and_belongs_to_many :manages_collections, class_name: 'Collection', join_table: 'managers'
```
justinlittman, line 8: Just `manages` instead of `manages_collections`?

app/models/user.rb
```diff
@@ -4,4 +4,13 @@
...
+  has_and_belongs_to_many :deposits_into, class_name: 'Collection', join_table: 'depositors'
```
justinlittman, line 9: `depositor_for`?

app/presenters/collection_presenter.rb
```diff
@@ -8,4 +8,8 @@ def initialize(collection:, collection_form:, version_status:)
   end
 
   attr_reader :collection
+
+  def participants(role)
```
justinlittman, line 12: Can you comment this code?

app/services/to_collection_form/mapper.rb
```diff
@@ -9,6 +9,7 @@ def self.call(...)
 
     def initialize(cocina_object:)
       @cocina_object = cocina_object
+      @collection = Collection.find_by(druid: cocina_object.externalIdentifier)
```
justinlittman, line 12: Can this be moved to the caller instead of in the mapper?

app/services/to_collection_form/mapper.rb
```diff
@@ -27,8 +28,16 @@ def params
         description: CocinaSupport.abstract_for(cocina_object:), # Cocina abstract maps to Collection description
         contact_emails_attributes: CocinaSupport.contact_emails_for(cocina_object:),
         related_links_attributes: CocinaSupport.related_links_for(cocina_object:),
+        managers_attributes: participant_attributes(:managers),
+        depositors_attributes: participant_attributes(:depositors),
         version: cocina_object.version
       }
     end
+
+    def participant_attributes(role)
```
justinlittman, line 37: Can you comment the code? (For me, using `.send` is magic, so always begs a comment.)

spec/requests/destroy_collection_spec.rb
```diff
@@ -5,12 +5,18 @@
...
+  before do
+    allow(Sdr::Repository).to receive(:find).with(druid:).and_return(cocina_object)
+    allow(Sdr::Repository).to receive(:discard_draft)
+  end
 
   context 'when the user is not authorized' do
     before do
       create(:collection, druid:)
-      allow(Sdr::Repository).to receive(:find).with(druid:).and_return(collection_with_metadata_fixture)
+      # allow(Sdr::Repository).to receive(:find).with(druid:).and_return(collection_with_metadata_fixture)
```
justinlittman, line 19: Cruft?

app/jobs/deposit_collection_job.rb
justinlittman: Also, where does this get tested? <-- This is my only critical question / comment.

justinlittman (changes requested):
Only one critical below, otherwise minor suggestions / requests.

app/jobs/deposit_collection_job.rb
aaron-collier: I was thinking this was tested in the create_collection system test, but I'll expand the job spec as well.

spec/system/edit_collection_spec.rb
aaron-collier: I think we're stuck using IDs here, especially on form elements that have matching strings, in this case 'SUNet ID'.

I'll continue to look for a solution, but don't think it should hold this up and can be improved in a follow up PR.

justinlittman: Maybe put it in a `within` to disambiguate?

See https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/spec/system/manage_authors_spec.rb#L51-L61

aaron-collier: Done. Awesome, this was the example I needed as my prior attempt to use `within` was not fruitful.

justinlittman (approved):

PR #392, opened 2024-12-18
[SPIKE] Replace ActionPolicy with Pundit

mjgiarlo:
Long story short: not much of a difference at all, so sticking with AP seems defensible?


justinlittman (commented):
That wasn't very edifying.

PR #393, opened 2024-12-18
Breadcrumbs for works and collections

### LINKED ISSUE #330: Implement breadcrumbs as described in the designs
Review how the breadcrumbs are displayed through the figma designs https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1092-2227&node-type=canvas&t=ycXCLrdQjVVOrGwq-0 in order to implement in H3

lwrubel:
Resolves #330. Adds breadcrumbs for collections and works. There are some small deviations from the design based on conversation in ticket with Astrid: Does not handle auto-updating the breadcrumb when someone enters a title (@astridu does this still make sense-- to change the breadcrumbs while the user is working?)

Creating a work:
Screenshot 2024-12-18 at 5 01 28 PM

Editing a work:
Screenshot 2024-12-18 at 5 01 42 PM

Showing a work:
Screenshot 2024-12-18 at 5 05 27 PM

Creating a collection:
Screenshot 2024-12-18 at 5 02 09 PM

Editing a collection:
Screenshot 2024-12-18 at 5 01 54 PM

Showing a collection:
Screenshot 2024-12-18 at 5 01 48 PM


app/presenters/work_presenter.rb
```diff
@@ -27,7 +27,7 @@ def contact_emails
   end
 
   def collection_link
-    link_to(collection.title, Rails.application.routes.url_helpers.collection_path(collection))
+    link_to(collection.title, Rails.application.routes.url_helpers.collection_path(druid: collection.druid))
```
lwrubel, line 30: Fix for link unrelated to breadcrumbs.

app/components/elements/breadcrumb_component.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Elements
+  # Displays an item in the top bread crumb navigation
```
mjgiarlo, line 4: Trivial:

```suggestion
  # Displays an item in the top breadcrumb navigation
```

app/components/elements/breadcrumb_component.rb
```diff
@@ -0,0 +1,25 @@
...
+    attr_reader :text, :link
+
+    def truncated_title
+      truncate(text, length: 150, separator: ' ')
+    end
+
+    def classes
+      return 'breadcrumb-item active' if link.nil? || @active
+
+      'breadcrumb-item'
```
mjgiarlo, line 22: I'm fuzzy on where we want to stick to `merge_classes`, but maybe @justinlittman can say more about whether we should be using that here. My two cents: ☝🏻 is perfectly readable.

app/components/elements/breadcrumb_nav_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Elements
+  # Displays the top bread crumb navigation
```
mjgiarlo, line 4: Trivial:

```suggestion
  # Displays the top breadcrumb navigation
```

app/components/elements/breadcrumb_nav_component.rb
```diff
@@ -0,0 +1,16 @@
...
+  class BreadcrumbNavComponent < ApplicationComponent
+    renders_many :breadcrumbs, BreadcrumbComponent
+
+    def initialize(dashboard: true, admin: false)
+      @dashboard = dashboard
+      @admin = admin
+      super()
+    end
+
+    attr_reader :dashboard, :admin
```
mjgiarlo, line 14: I'm not sure we've used this pattern consistently, but I have seen some spots where we do this instead:

```suggestion
    def dashboard?
      @dashboard
    end
    
    def admin?
      @admin
    end
```

app/views/collections/form.html.erb
```diff
@@ -1,3 +1,16 @@
+<% content_for :breadcrumbs do %>
+  <%= render Elements::BreadcrumbNavComponent.new(dashboard: true, admin: false) do |component| %>
+    <% if (current_page?(action: :new) || (controller.action_name == 'create')) %>
+      <% component.with_breadcrumb(text: I18n.t('collections.edit.no_title')) %>
```
mjgiarlo, line 4: Since we're in a view here, I think you can invoke `t()` without any scoping, and that makes sure we can use Rails' I18n extensions.

```suggestion
      <% component.with_breadcrumb(text: t('collections.edit.no_title')) %>
```

app/views/collections/show.html.erb
```diff
@@ -1,3 +1,12 @@
+<% content_for :breadcrumbs do %>
+  <%= render Elements::BreadcrumbNavComponent.new(dashboard: true, admin: false) do |component| %>
+    <% if current_page?(action: :new) %>
+      <% component.with_breadcrumb(text: I18n.t('collections.edit.no_title')) %>
```
mjgiarlo, line 4: Same i18n scoping comment as above

app/components/elements/breadcrumb_nav_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Elements
+  # Displays the top bread crumb navigation
+  class BreadcrumbNavComponent < ApplicationComponent
+    renders_many :breadcrumbs, BreadcrumbComponent
+
+    def initialize(dashboard: true, admin: false)
```
mjgiarlo, line 8: When would we use `dashboard: false`? Or `admin: true`? I assume this is a seam for something coming soon.

app/views/works/show.html.erb
```diff
@@ -1,3 +1,9 @@
+<% content_for :breadcrumbs do %>
+  <%= render Elements::BreadcrumbNavComponent.new(dashboard: true, admin: false) do |component| %>
+    <% component.with_breadcrumb(text: @work_presenter.collection.title, link: collection_path(druid: @work_presenter.collection.druid)) %>
+    <% component.with_breadcrumb(text: @work_form.title) %>
```
mjgiarlo, line 4:
```suggestion
    <% component.with_breadcrumb(text: @work_presenter.title) %>
```

app/controllers/works_controller.rb
```diff
@@ -20,6 +20,8 @@ def show
 
   def new
     collection = Collection.find_by!(druid: params.expect(:collection_druid))
+    @collection_title = collection.title
```
mjgiarlo, line 23: The changes in here make me wonder if we want to add `collection_title` as an attr to the `WorkForm`, the way we're handling the `collection_druid`, and then we'd access it via the form instance instead of via a new instance variable. What's currently here is a-ok. (If you have thoughts on this, I'm all ears, but if you don't, you can also disregard!)

spec/components/elements/breadcrumb_nav_component_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Elements::BreadcrumbNavComponent, type: :component do
+  let(:component) { described_class.new(dashboard: true) }
```
mjgiarlo, line 6: Since we have `dashboard` declared as a `let` in this scope, I think we could consider:

```suggestion
  let(:component) { described_class.new(dashboard:) }
```

And that would let us make this other change below...

spec/components/elements/breadcrumb_nav_component_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(page).to have_css('nav#breadcrumbs')
+      expect(page).to have_css('li.breadcrumb-item', count: 3)
+      expect(page).to have_link('Dashboard', href: '/')
+      expect(page).to have_link('Breadcrumb 1', href: '/example')
+      expect(page).to have_css('li.breadcrumb-item.active', text: 'Breadcrumb 2')
+    end
+  end
+
+  context 'without dashboard' do
+    let(:component) { described_class.new(dashboard: false) }
```
mjgiarlo, line 26:
```suggestion
```

astridu:
> Resolves #330. Adds breadcrumbs for collections and works. There are some small deviations from the design based on conversation in ticket with Astrid: Does not handle auto-updating the breadcrumb when someone enters a title (@astridu does this still make sense-- to change the breadcrumbs while the user is working?)

@lwrubel Yes, I think it makes sense so that they can see the title.

app/components/elements/breadcrumb_nav_component.rb
lwrubel: @astridu do you think there will be pages that don't have breadcrumbs that start with Dashboard?

app/components/elements/breadcrumb_nav_component.rb
lwrubel: Fixed

app/components/elements/breadcrumb_nav_component.rb
lwrubel: Thanks, made that change and it helps readability in the erb, I think.

app/controllers/works_controller.rb
lwrubel: I spent a little time trying this just now, since we have to create the instance variable in three different actions. However, where I got stuck was the `edit` action. It uses `ModelSync::Work.call` and the collection title is not in the work or the cocina. I wasn't sure how to get it into the form in that case. If you have an idea how to do that, let me know.

app/components/elements/breadcrumb_component.rb
lwrubel: Breadcrumbs should be styled the same wherever they are, so I don't think we need to pass different classes to the component when calling it. It doesn't seem like we need merge_classes here (so far at least) since this is the only logic related to classes. But I could be missing something.

app/components/elements/breadcrumb_nav_component.rb
mjgiarlo: @lwrubel I don't think it's overkill. I mostly wanted to be sure I wasn't missing something already in place.

app/controllers/works_controller.rb
mjgiarlo: OK, thanks for giving that a whirl. We could probably make a case for grabbing the collection title in `ModelSync::Work` but this is far enough from breadcrumbs that it may not be worth more of our time.

mjgiarlo (approved):

app/components/elements/breadcrumb_nav_component.rb
astridu: > @astridu do you think there will be pages that don't have breadcrumbs with Dashboard as the root?

I also looked through the collections and admin pages, and I don't see any that would not start at the Dashboard. So Dashboard should be the root for all pages. Thank you!

PR #405, opened 2024-12-19
Error handling for large files and too many files.

### LINKED ISSUE #371: Handle upload corner cases
Upload should be prevented and the user should be notified when:
* Uploading an empty folder
* Uploading > 10GB in total
* Uploading > 25K files in total

justinlittman:
closes #371

https://github.com/user-attachments/assets/ffcef9f5-115b-4fe2-b92a-dfcd04a1c3bf


mjgiarlo (approved):
LGTM. I wonder if we have a design for file upload errors and if that might be good follow-up work.

justinlittman:
I think that these sorts of errors will be few and far between, so wasn't worrying about design.

PR #416, opened 2024-12-20
Untangle forms in the tabbed interface.

justinlittman:
At the end of the day, the correct approach was to avoid nested forms altogether.


app/components/elements/forms/submit_component.rb
```diff
@@ -4,8 +4,8 @@ module Elements
   module Forms
     # Component for a form submit button
     class SubmitComponent < ApplicationComponent
-      def initialize(form:, label: nil, variant: :primary, classes: [], **options)
-        @form = form
+      def initialize(form_id: nil, label: nil, variant: :primary, classes: [], **options)
```
justinlittman, line 7: You really don't need a form to create a submit button. Setting the `form` attribute can be used to tie it to its form even if it isn't contained in the form.

app/components/elements/tab_form/tab_list_component.html.erb
```diff
@@ -7,15 +7,23 @@
     <% end %>
   </div>
   <div class="col-md-9">
-    <%= form_with(model:) do |form| %>
-      <% hidden_fields.each do |field| %>
-        <%= form.hidden_field field %>
+    <div class="tab-content h-100">
+      <% before_form_panes.each do |pane| %>
```
justinlittman, line 11: Splitting out panes that shouldn't be contained in the form.

app/controllers/contents_controller.rb
```diff
@@ -15,10 +15,6 @@ def show_table
     authorize! @content
   end
```
justinlittman, line 17: No longer used since can inline the edit form.

app/helpers/application_helper.rb
```diff
@@ -4,21 +4,4 @@ module ApplicationHelper
...
     boolean ? 'Yes' : 'No'
   end
-
-  # Form with an id and for which all input tags have the form attribute set.
-  # This is useful for forms that are not directly inside the form tag or
-  # for forms that are inside other forms. (When a form is inside another form,
-  # the browser will not submit the inner form.)
-  # Unfortunately, Rails doesn't provide a clean seam for adding the form attribute
-  # to the autogenerated hidden fields so here we are.
-  def ided_form_with(**args, &)
```
justinlittman, line 14: Since the forms are untangled, we don't need this hack.

app/components/elements/tab_form/tab_list_component.rb
```diff
@@ -5,26 +5,29 @@ module TabForm
...
-      attr_reader :model, :hidden_fields
+      attr_reader :model, :hidden_fields, :form_id
 
       def classes
         # Provides d-flex, tabbable-panes as the static default classes
         # merged with any additional classes passed in.
         merge_classes(%w[row tabbable-panes gx-4 gy-4 mb-5], @classes)
       end
-
-      def render?
```
justinlittman, line 25: There's no use case for this so getting rid of it.

lwrubel:
So this un-nests the files form (via its pane) and the deposit button's pane, which is no longer in a form?

justinlittman:
Yes, that sounds right.

app/views/collections/form.html.erb
```diff
@@ -13,28 +13,29 @@
 <% end %>
 <%= render Elements::HeaderComponent.new(level: :h1, variant: :h2, text: @collection_form.title || 'Untitled collection', classes: 'mb-3') %>
 <%= render Elements::AlertComponent.new(variant: :danger, value: I18n.t('collections.edit.messages.validation')) if @collection_form.errors.present? %>
-<%= render Elements::TabForm::TabListComponent.new(model: @collection_form, hidden_fields: %i[lock version]) do |component| %>
+<% form_id = dom_id(@collection_form, 'tabbed_form') %>
```
mjgiarlo, line 16: I pondered whether we might expect this to be set as an instance variable in the controller instead of a local variable, but I don't think it matters too much?

mjgiarlo (approved):
Dropped a couple quick suggestions but overall this LGTM. The one piece I'm struggling to remember at the moment is what forms are present in the before and after panes that cause tangling? Might be worth a small comment amendment, maybe.

app/views/collections/form.html.erb
justinlittman: Yeah, my thought it that it made it a tad clearer to put it closer to where it was used at the expense of slightly more complicated erb.

justinlittman:
@mjgiarlo Added additional comments per suggestion.

PR #426, opened 2025-01-06
Remove duplicate `license` from path to license label field in i18n

mjgiarlo:
Follows #425


lwrubel (approved):

lwrubel:
This ends up being a different path than the other fields, but it gives us something shorter, so that's an improvement.

PR #428, opened 2025-01-06
Switches to looking up version status in batch from dashboard.

### LINKED ISSUE #427: Dashboard: Retrieve version statuses in batch
Instead of retrieving one at a time.

justinlittman:
closes #427

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -5,14 +5,20 @@ class DashboardController < ApplicationController
   skip_verify_authorized only: :show
 
   def show
-    # This is going to be slow because it makes a call to DSA for each work.
-    # It is also going to block rendering since it is synchronous.
+    # This is going to be slow because it is going to block rendering since it is synchronous.
     # Also, it processes all of a user's works, but not all may be rendered on the dashboard.
+    druid_to_status_map = Sdr::Repository.statuses(druids: druids_for_works_for_current_user)
     @draft_works = []
     @status_map = current_user.works.to_h do |work|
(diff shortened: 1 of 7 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -5,14 +5,20 @@ class DashboardController < ApplicationController
   skip_verify_authorized only: :show
 
   def show
-    # This is going to be slow because it makes a call to DSA for each work.
-    # It is also going to block rendering since it is synchronous.
+    # This is going to be slow because it is going to block rendering since it is synchronous.
     # Also, it processes all of a user's works, but not all may be rendered on the dashboard.
+    @druid_to_status_map = Sdr::Repository.statuses(druids: druids_for_works_for_current_user)
```
mjgiarlo, line 10: If this instance variable isn't used in any views, it could be a local variable so it doesn't leak.

app/controllers/dashboard_controller.rb
```diff
@@ -5,14 +5,20 @@ class DashboardController < ApplicationController
...
+      version_status = @druid_to_status_map[work.druid] || VersionStatus::NilStatus.new
       @draft_works << work if version_status.draft?
       [work.id, version_status]
     end
   end
+
+  private
+
+  def druids_for_works_for_current_user
+    current_user.works.where.not(druid: nil).pluck(:druid)
```
mjgiarlo, line 22: Maybe we'll eventually replace this with a policy scope? (But this is fine for now, I think.)

app/controllers/dashboard_controller.rb
justinlittman: Agree. `@` was cruft.

mjgiarlo (approved):

PR #434, opened 2025-01-07
Use :on instead of :if for validations to separate data concerns from context concerns

mjgiarlo:
I stumbled on https://guides.rubyonrails.org/active_record_validations.html#on over the winter break and thought we might be able to dump some custom validation code in H3 to be more Rails-y.

### DIFF
--- a/app/forms/application_form.rb
+++ b/app/forms/application_form.rb
@@ -19,16 +19,4 @@ class ApplicationForm
   def self.immutable_attributes
     []
   end
-
-  # @param deposit [Boolean] whether the form is being used for a deposit
-  def valid?(deposit: false)
-    @deposit = deposit
-    super
-  end
-
-  # This can be used to control validations specific to deposits.
-  # For example: validates :authors, presence: { message: 'requires at least one author' }, if: :deposit?
-  def deposit?
-    @deposit
-  end
 end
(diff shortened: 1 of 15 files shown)


app/forms/application_form.rb
```diff
@@ -19,16 +19,4 @@ def self.user_editable_attributes
   def self.immutable_attributes
     []
   end
-
-  # @param deposit [Boolean] whether the form is being used for a deposit
-  def valid?(deposit: false)
-    @deposit = deposit
-    super
-  end
```
mjgiarlo, line 27: Neither ActiveRecord nor ActiveModel objects support kwargs in `#valid?`, and we want our forms to be Rails-y.

app/forms/author_form.rb
```diff
@@ -3,15 +3,15 @@
 # Form for an author
 class AuthorForm < ApplicationForm
   attribute :first_name, :string
-  validates :first_name, presence: true, if: -> { person? && deposit? }
+  validates :first_name, presence: true, on: :deposit, if: -> { person? }
```
mjgiarlo, line 6: This batch of changes separate data concerns ("Is the form a person-like author or an org-like author?") from context concerns ("Are we validating a draft or a deposit?").

justinlittman (approved):
Oh, I like this!

lwrubel:
Helpful to learn more about how to use validation contexts, thank you!

PR #461, opened 2025-01-09
Applies collection settings for licenses to works.

### LINKED ISSUE #460: Additional license functionality for work form
* When the collection allows the user to choose a license, but has a default, select that default.
* When the collection requires a license, use that license.

justinlittman:
closes #460

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -35,7 +35,6 @@ class WorkForm < ApplicationForm
   validates :citation, presence: true, if: -> { auto_generate_citation == false }
 
   attribute :license, :string
-  validates :license, presence: true, on: :deposit
 
   attribute :work_type, :string
   validates :work_type, presence: true, on: :deposit
--- a/app/presenters/work_presenter.rb
+++ b/app/presenters/work_presenter.rb
@@ -41,7 +41,7 @@ class WorkPresenter < FormPresenter
   end
 
   def license_label
-    License.find_by(id: license)&.label
+    license_presenter.label
   end
 
   def publication_date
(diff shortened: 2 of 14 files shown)


justinlittman:
When required:
image

When depositor selects:
image

app/forms/work_form.rb
```diff
@@ -35,7 +35,6 @@ def persisted?
   validates :citation, presence: true, if: -> { auto_generate_citation == false }
 
   attribute :license, :string
-  validates :license, presence: true, on: :deposit
```
justinlittman, line 38: Works don't have to have a license.

app/presenters/work_presenter.rb
```diff
@@ -51,4 +51,8 @@ def publication_date
   private
 
   delegate :collection, :created_at, :user, to: :work
+
+  def license_presenter
+    @license_presenter ||= LicensePresenter.new(work_form: self, collection:)
```
mjgiarlo, line 56: Is `self` here the `WorkPresenter` instance? It's being passed to the license presenter as the work form? Hmmm.

mjgiarlo (approved):
Have a question about an arg. Approving anyway.

app/presenters/work_presenter.rb
justinlittman: The WorkPresenter is a SimpleDelegator that delegates to the work form, so this should be fine.

However, as you note, this is a bit off. I changed to get the work form from the work presenter.

mjgiarlo: @justinlittman Excellent.

PR #465, opened 2025-01-10
Align delete icon with input field in some forms

### LINKED ISSUE #337: Trash can alignment
Per @astridu ([comment](https://github.com/sul-dlss/hungry-hungry-hippo/pull/303#issuecomment-2537398671)), the trash can should be aligned with its text field. See contact email fields in collections and work forms.

lwrubel:
Resolves #337. The trash can should be aligned with the input field on repeatable forms with only one visible field (contact emails and keywords)

_Before_:

Screenshot 2025-01-10 at 8 54 10 AM

_After_: 
Screenshot 2025-01-10 at 11 02 50 AM

Screenshot 2025-01-10 at 11 02 57 AM

### DIFF
--- a/app/components/elements/forms/repeatable_nested_component.rb
+++ b/app/components/elements/forms/repeatable_nested_component.rb
@@ -6,7 +6,7 @@ module Elements
     # NOTE: Use the `NestedComponentPresenter` to invoke this component; do not directly instantiate it.
     class RepeatableNestedComponent < ApplicationComponent
       def initialize(form:, model_class:, field_name:, form_component:, hidden_label: false, bordered: true, # rubocop:disable Metrics/ParameterLists
-                     reorderable: false, label: nil)
+                     reorderable: false, single_field: false, label: nil)
         @form = form
         @model_class = model_class
         @field_name = field_name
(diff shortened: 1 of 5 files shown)


app/components/elements/forms/repeatable_nested_component.rb
```diff
@@ -6,14 +6,15 @@ module Forms
     # NOTE: Use the `NestedComponentPresenter` to invoke this component; do not directly instantiate it.
     class RepeatableNestedComponent < ApplicationComponent
       def initialize(form:, model_class:, field_name:, form_component:, hidden_label: false, bordered: true, # rubocop:disable Metrics/ParameterLists
-                     reorderable: false, label: nil)
+                     reorderable: false, single_field: false, label: nil)
```
lwrubel, line 9: I tried to avoid adding another param, but because we can't rely on the number of fields in the form--in the case of keywords there are additional hidden fields--ended having to take this route.

justinlittman (approved):

app/components/elements/forms/repeatable_nested_component.rb
mjgiarlo: @lwrubel the struggle is real! Thanks for the rationale. 😄

PR #466, opened 2025-01-10
Adds access to access settings.

### LINKED ISSUE #463: Add download access to work
Including adding to work form, work show, mapping to cocina, etc.

See https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=1116-5106&node-type=frame&t=VbNvdelVP8tjSSJT-0

justinlittman:
closes #463

image

### DIFF
--- a/app/components/works/edit/access_settings_component.rb
+++ b/app/components/works/edit/access_settings_component.rb
@@ -4,9 +4,26 @@ module Works
   module Edit
     # Component for rendering the access settings pane.
     class AccessSettingsComponent < ApplicationComponent
-      def initialize(form:)
+      def initialize(form:, collection:)
         @form = form
+        @collection = collection
         super()
+      end
+
+      attr_reader :form
+
+      delegate :depositor_selects_access?, to: :@collection
+
+      def access_options
+        [
+          [I18n.t('access.stanford'), 'stanford'],
+          [I18n.t('access.world'), 'world']
+        ]
+      end
+
+      def collection_selects_access_message
+        access_text = form.object.access == 'stanford' ? 'the Stanford Community' : 'anyone'
+        "The files in your deposit will be downloadable by #{access_text}."
       end
     end
   end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -18,14 +18,13 @@ class WorksController < ApplicationController
   end
 
   def new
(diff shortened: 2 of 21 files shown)


app/controllers/works_controller.rb
```diff
@@ -153,4 +152,13 @@ def editable?
   def set_presenter
     @work_presenter = WorkPresenter.new(work: @work, work_form: @work_form, version_status: @version_status)
   end
+
+  def new_work_form
+    WorkForm.new(
+      collection_druid: @collection.druid,
+      content_id: @content.id,
+      license: @collection.license,
+      access: @collection.stanford_access? ? 'stanford' : 'world'
```
lwrubel, line 161: Where is this defined?

app/components/works/edit/access_settings_component.rb
```diff
@@ -4,10 +4,27 @@ module Works
...
         @form = form
+        @collection = collection
         super()
       end
+
+      attr_reader :form
+
+      delegate :depositor_selects_access?, to: :@collection
+
+      def access_options
```
lwrubel, line 17: Is the default supposed to be Stanford Community? H2 says, "Depositor selects — default for all deposits will be that everyone can download the files, but Depositor can choose to restrict file download for their deposit to the Stanford community." 

The mockups have a blank default selection and other dropdowns have a "Select..." prompt.

justinlittman: The default is set to "world" here: https://github.com/sul-dlss/hungry-hungry-hippo/pull/466/files#diff-0b136dc68a89fcfa9a4565393b29722171443cc2cf4478e6c16b0e550eda9674

I got ready of Select... because there is a default.

app/controllers/works_controller.rb
justinlittman: It's a magic method that comes from the access enum.

app/components/works/edit/access_settings_component.rb
lwrubel: Oh, so that will override the order of the options in the dropdown. Got it.

app/controllers/works_controller.rb
lwrubel: Should have figured it was magic.

lwrubel (approved):

PR #491, opened 2025-01-16
Adds additional terms to work.

### LINKED ISSUE #482: Add additional terms of use to work form and show
See https://www.figma.com/proto/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1116-5067&t=tNlOqUva0dJvpIPs-1

justinlittman:
closes #482

image

### DIFF
--- a/app/services/cocina_support.rb
+++ b/app/services/cocina_support.rb
@@ -192,4 +192,8 @@ class CocinaSupport
   def self.release_date_for(cocina_object:)
     cocina_object.access&.embargo&.releaseDate
   end
+
+  def self.use_reproduction_statement_for(cocina_object:)
+    cocina_object.access.useAndReproductionStatement
+  end
 end
--- a/app/services/terms_of_use_support.rb
+++ b/app/services/terms_of_use_support.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Helper class for terms of use
+class TermsOfUseSupport
+  def self.full_statement(custom_rights_statement:)
+    [custom_rights_statement, I18n.t('license.terms_of_use')].compact_blank.join("\n\n")
+  end
+end
--- a/app/services/to_work_form/mapper.rb
+++ b/app/services/to_work_form/mapper.rb
@@ -36,7 +36,8 @@ module ToWorkForm
         access: CocinaSupport.access_for(cocina_object:),
         version: cocina_object.version,
         collection_druid: CocinaSupport.collection_druid_for(cocina_object:),
-        publication_date_attributes: CocinaSupport.event_date_for(cocina_object:, type: 'publication')
+        publication_date_attributes: CocinaSupport.event_date_for(cocina_object:, type: 'publication'),
+        custom_rights_statement:
       }.merge(work_type_params).merge(release_date_params)
     end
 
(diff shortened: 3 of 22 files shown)


app/services/to_work_form/mapper.rb
```diff
@@ -56,5 +57,18 @@ def release_date_params
...
       { release_date:, release_option: 'delay' }
     end
+
+    def custom_rights_statement
+      use_statement = CocinaSupport.use_reproduction_statement_for(cocina_object:)
+      return if use_statement.blank?
+      return if use_statement == default_terms_of_use
+
+      # Remove the default terms
+      use_statement&.delete_suffix(default_terms_of_use)&.strip
```
mjgiarlo, line 67: We're short-circuiting a few lines above if the use statement is blank, so I'm curious about the safe navigation here.

app/services/terms_of_use_support.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Helper class for terms of use
+class TermsOfUseSupport
+  def self.full_statement(custom_rights_statement:)
+    [custom_rights_statement, I18n.t('license.terms_of_use')].compact_blank.join("\n\n")
+  end
+end
```
mjgiarlo, line 8: This makes me wonder how we feel about helpers, generally. Helpers are rails-y, and we want to be "on the rails." But we've also intentionally avoided helpers in the past (maybe in the early H2 days?). But... this is an awfully tiny class, the documentation of which calls it out as a helper, and we do already use some helpers in H3. It may be worth chatting about when to use and when to avoid helpers in H3.

app/services/cocina_support.rb
```diff
@@ -192,4 +192,8 @@ def self.license_for(cocina_object:)
   def self.release_date_for(cocina_object:)
     cocina_object.access&.embargo&.releaseDate
   end
+
+  def self.use_reproduction_statement_for(cocina_object:)
+    cocina_object.access.useAndReproductionStatement
```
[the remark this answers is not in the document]
justinlittman: Not clear to me. If I'm wrong, we'll definitely find out ...

app/services/terms_of_use_support.rb
justinlittman: Everything I know about helpers, I learned from jcoyne: Helpers are for use in views only.

Since `full_statement` isn't used exclusively in a view I didn't make it a helper.

As far as I know, Rails doesn't have a convention for a non-view helper. I've adopted my own convention of calling them "support", e.g., "TermsOfUseSupport".

app/services/to_work_form/mapper.rb
justinlittman: Yeah, that got left behind when I refactored this method. Removing.

app/services/terms_of_use_support.rb
mjgiarlo: I bet we could come up with a way to write a module containing all the Rails helpers, which could then be included in various POROs (services, mappers, presenters), but I'm not sure that's better than what we have here. :shipit:

mjgiarlo (approved):

PR #495, opened 2025-01-17
Add collection owner as a manager on create.

### LINKED ISSUE #492: Make user a collection's manager by default
When a user is creating a collection, make the user a manager of it by default. Show this in the collection form as seen in the mockup:

https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=789-4316&t=WmkmZCNXExSCRDey-4

justinlittman:
closes #492

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Collection do
+  let(:user) { create(:user) }
+
+  describe 'add user to collection' do
+    it 'adds the collection owner to the collection' do
+      collection = described_class.create(title: collection_title_fixture, user:)
+      expect(collection.managers).to eq([user])
+    end
+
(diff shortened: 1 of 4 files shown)


spec/models/collection_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    context 'when collection owner is already a manager' do
+      it 'does not add the collection owner to the collection' do
+        collection = described_class.create(title: collection_title_fixture, user:, managers: [user])
+        expect(collection.managers).to eq([user])
+      end
+    end
+
+    context 'when no collection owner' do
```
lwrubel, line 21: What is the situation where there is a collection with no owner?

lwrubel (approved):

spec/models/collection_spec.rb
justinlittman: Tests?

As currently implemented, a collection model doesn't have to have a user so wanted to err on the side of making sure it was covered.

Issue #208, opened 2024-12-02
Implement email notifications


aaron-collier:
See: https://docs.google.com/spreadsheets/d/18zrWLqbU_j7PJtCOS5cqFVZnN_OnVninFsqByLVYM4M/edit?gid=1881110513#gid=1881110513 for necessary notification content and trigger logic

amyehodge:
@aaron-collier That link above is not right. That's to a file with some text changes that I'd like to see made to a few of the existing email notifications. There are many more than that. Let's discuss what you need here. I did pull out some info about the two that Donna indicated she used. I'll see if I can find that.

amyehodge:
Required email for capstones:

To depositor
- Invited to deposit: https://github.com/sul-dlss/happy-heron/blob/main/app/views/collections_mailer/invitation_to_deposit_email.html.erb
- Item submitted for review/approval: https://github.com/sul-dlss/happy-heron/blob/main/app/views/works_mailer/submitted_email.html.erb
-- this needs a text change that is noted in the google sheet linked to at the bottom of this comment
- Item returned for changes: https://github.com/sul-dlss/happy-heron/blob/main/app/views/works_mailer/reject_email.html.erb
- Item deposited: https://github.com/sul-dlss/happy-heron/blob/main/app/views/works_mailer/deposited_email.html.erb
-- this needs a text change that is noted in the google sheet linked to at the bottom of this comment
- Item accepted for deposit: https://github.com/sul-dlss/happy-heron/blob/main/app/views/works_mailer/approved_email.html.erb
-- this needs a text change that is noted in the google sheet linked to at the bottom of this comment
- New version of item deposited: https://github.com/sul-dlss/happy-heron/blob/main/app/views/works_mailer/new_version_deposited_email.html.erb
-- this needs a text change that is noted in the google sheet linked to at the bottom of this comment

To reviewer
- Invited to be a reviewer: https://github.com/sul-dlss/happy-heron/blob/main/app/views/collections_mailer/review_access_granted_email.html.erb

To manager
- Added as manager of collection: https://github.com/sul-dlss/happy-heron/blob/main/app/views/collections_mailer/manage_access_granted_email.html.erb

To reviewer or manager
- Item ready for review: https://github.com/sul-dlss/happy-heron/blob/main/app/views/reviewers_mailer/submitted_email.html.erb
-- this needs a text change that is noted in the google sheet linked to at the bottom of this comment


[...]

PR #493, opened 2025-01-17
Adds review workflow.

justinlittman:
closes #209

PR #501, opened 2025-01-21
Add emails.

justinlittman:
refs #208

This is for the first 2 emails only.

image


app/models/collection.rb
```diff
@@ -6,9 +6,12 @@ class Collection < ApplicationRecord
   belongs_to :user
   has_many :works, dependent: :destroy
 
-  has_and_belongs_to_many :depositors, class_name: 'User', join_table: 'depositors'
-  has_and_belongs_to_many :managers, class_name: 'User', join_table: 'managers'
-  has_and_belongs_to_many :reviewers, class_name: 'User', join_table: 'reviewers'
+  has_many :collection_depositor, dependent: :destroy
```
justinlittman, line 9: Switched to has_many through so that can take advantage of callbacks.

config/environments/test.rb
```diff
@@ -55,5 +55,5 @@
   # Raise error when a before_action's only/except options reference missing actions.
   config.action_controller.raise_on_missing_callback_actions = true
 
-  config.active_job.queue_adapter = :async
+  config.active_job.queue_adapter = :inline
```
justinlittman, line 58: Inline plays nicely with the tests.

jcoyne:
Should the rosette be removed?

justinlittman:
@astridu @amyehodge ?

app/views/collections_mailer/manage_access_granted_email.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render Emails::SalutationComponent.new(user: @user) %>
+
+<p>You have been invited to be a Manager of the <%= @collection.title %> collection in the
+   Stanford Digital Repository. Managers have the ability to manage global settings for the collection,
+   manage Depositors and other participants, and edit any deposits in the collection.</p>
+
+<p>To visit the collection where deposits are made, and for more information about SDR repository services,
+   please visit http://<%= Settings.host %>.</p>
```
lwrubel, line 8: Is there a way to avoid putting http:// in the URL and use https:// instead? I assume this is because it needs to work with localhost?

justinlittman: No, that is an oversight. Fixed.

app/models/collection.rb
```diff
@@ -6,9 +6,12 @@ class Collection < ApplicationRecord
...
+  has_many :collection_manager, dependent: :destroy
```
mjgiarlo, line 10: I was surprised to see two `has_many` relationships expressed as singular nouns instead of plural ones. I'm assuming there's a reason for them being singular despite the convention?

app/models/user.rb
```diff
@@ -5,9 +5,12 @@ class User < ApplicationRecord
   has_many :works, dependent: :destroy
   has_many :collections, dependent: :destroy
 
-  has_and_belongs_to_many :manages, class_name: 'Collection', join_table: 'managers'
-  has_and_belongs_to_many :depositor_for, class_name: 'Collection', join_table: 'depositors'
-  has_and_belongs_to_many :reviewer_for, class_name: 'Collection', join_table: 'reviewers'
+  has_many :collection_manager, dependent: :destroy
+  has_many :collection_depositor, dependent: :destroy
+  has_many :collection_reviewer, dependent: :destroy
```
mjgiarlo, line 10: Same question here about `has_many` expressed as singular nouns.

config/environments/test.rb
mjgiarlo: Do you know why/when `async` started not playing nicely? I can imagine it being a bit risky baking the assumption that jobs run synchronously into the tests, but I don't have a specific objection.

justinlittman: Now there are numerous actions that are triggered in tests that now result in mailer jobs being submitted. With async, those were being performed at unexpected times. Inline forces them to happen within the test that they were being submitted.

Async was probably a bad choice to begin with.

config/initializers/subscriptions.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Rails.application.config.after_initialize do
+  # Subscriptions for CollectionsMailer
+  Notifier.subscribe_mailer(event_name: Notifier::DEPOSITOR_ADDED, mailer_class: CollectionsMailer,
+                            mailer_method: :invitation_to_deposit_email)
+  Notifier.subscribe_mailer(event_name: Notifier::MANAGER_ADDED, mailer_class: CollectionsMailer,
+                            mailer_method: :manage_access_granted_email)
```
mjgiarlo, line 8: Can you say why the reviewer event doesn't need to be here?

spec/support/mail_matchers.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# require 'rspec/expectations'
```
mjgiarlo, line 3: Cruft?

app/models/collection.rb
justinlittman: Good catch. Fixed.

mjgiarlo:
> @astridu @amyehodge ?

The rest of the Infra portfolio already uses the new SUL logos, so I imagine we don't want to introduce the old logos in H3 now.

config/initializers/subscriptions.rb
justinlittman: Adding reviewer is pending https://github.com/sul-dlss/hungry-hungry-hippo/pull/493

spec/support/mail_matchers.rb
justinlittman: Yes. Removed.

mjgiarlo:
@justinlittman 💬 
> @astridu @amyehodge Please ticket separately if you want the design changed.

The rest of the Infra portfolio already uses the new SUL logos, FWIW. A follow-up PR to swap out the logos should be pretty easy. No need to hold this up over that, I don't think.

amyehodge:
I'm just seeing this about the logos. I haven't talked to @astridu but I can't imagine why we wouldn't want the new logos here.

amyehodge:
Since this has been merged I'll put in a new ticket for the new logo.

PR #502, opened 2025-01-21
Change work policy so that user has permissions for own works but not…

### LINKED ISSUE #498: Depositors should only be able to see their own items
If a user only has depositor permissions for a collection, they should only be able to see and edit their own items in that collection. 
- On the dashboard, they should only see their own items listed in the collection table.
- On the collection deposits page, they should only see their own items listed in the table there.

"their own items" above means any item the user created and that has not been reassigned to a different owner, and any item created by someone else that has been reassigned to the current user. Please let me know if that's not clear. This may not apply yet in H3 if the ability to change ownership has not yet been implemented.

Note that collection creators, managers, and reviewers should always see everything in the collection. I believe the only current exception to this in H2 is that only admins see decommissioned items.

justinlittman:
… others in collection.

closes #498

### DIFF
--- a/app/components/dashboard/show/works_list_component.rb
+++ b/app/components/dashboard/show/works_list_component.rb
@@ -12,8 +12,6 @@ module Dashboard
 
       attr_reader :collection
 
-      delegate :works, to: :collection
-
       def id
         @id ||= dom_id(collection, 'table')
       end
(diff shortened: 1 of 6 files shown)


aaron-collier (approved):

mjgiarlo (approved):
LGTM. I am wondering if/when we might want to consider shifting the responsibility for returning authorized resources to our policy objects (using AP scopes).

justinlittman:
@mjgiarlo https://github.com/sul-dlss/hungry-hungry-hippo/issues/510

PR #509, opened 2025-01-22
Adds additional emails

justinlittman:
refs #208

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -187,12 +187,9 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
     # Setting the deposit_job_started_at to the current time to indicate that the deposit job has started and user
     # should be "waiting".
     work.update!(deposit_job_started_at: Time.zone.now)
-    deposit = deposit?
-    if request_review?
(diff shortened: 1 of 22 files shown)


aaron-collier (approved):

app/controllers/works_controller.rb
```diff
@@ -187,12 +187,9 @@ def perform_deposit(work:)
...
     # should be "waiting".
     work.update!(deposit_job_started_at: Time.zone.now)
-    deposit = deposit?
-    if request_review?
-      work.request_review!
-      deposit = false # Will be saved, but not deposited until approved.
-    end
-    DepositWorkJob.perform_later(work:, work_form: @work_form, deposit:)
+    # If request review, deposit will be saved, but not deposited until approved.
+    deposit = request_review? ? false : deposit?
```
mjgiarlo, line 191: @justinlittman This seems like the intent is to short-circuit the deposit if review has been requested, on behalf of the `DepositWorkJob`. But the value of `request_review?` is being passed directly to `DepositWorkJob` and if that's a concern of the job, is there a reason it shouldn't be doing its own short-circuiting? I.e., why is this check here and not in the job?

PR #513, opened 2025-01-22
Adds state for deposit and RabbitMQ machinery.

justinlittman:
The RabbitMQ machinery is needed so that we know when accessioning is complete.

The deposit state is so that we can track when a work/collection is being persisted and accessioned.

### DIFF
--- a/app/models/concerns/deposit_state_machine.rb
+++ b/app/models/concerns/deposit_state_machine.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Defines the deposit states for an object.
+module DepositStateMachine
+  extend ActiveSupport::Concern
+
+  included do
+    # In general, the version status should be used for the status of an object.
+    # However, deposit_state is used for very specific purposes.
+    # In particular, persisting is used to indicate that the DepositWorkJob is queued / running.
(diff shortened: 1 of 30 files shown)


app/models/concerns/deposit_state_machine.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+
+      event :accession_complete do
+        transition accessioning: :deposit_none
+      end
+
+      before_transition accessioning: :deposit_none do |object|
+        Notifier.publish(Notifier::ACCESSIONING_COMPLETE, object:)
+      end
+    end
```
mjgiarlo, line 37: This feedback isn't as actionable as I'd like, but one source of confusion in H2 was the nomenclature we used for states and events, and I'm fuzzy on what I'm looking at above. I'd like to see us try to follow two guidelines when naming states and events:

1. Strive for consistency in using nouns / verbs / whatever in naming these (maybe opting to follow the advice here: https://stately.ai/blog/2024-01-23-state-machines-whats-in-a-name#what-or-how-should-i-name-this-thing)
1. Use terms that we developers already use (e.g., I'm not sure I've ever heard a developer talk about "deposits" being "persisted," even though I can make sense of that after a couple seconds of squinting)

mjgiarlo (approved):

PR #515, opened 2025-01-23
Add emails to be sent after accession.

justinlittman:
refs #208


app/jobs/deposit_complete_job.rb
```diff
@@ -13,19 +13,41 @@ class DepositCompleteJob
...
+
+  def message
+    @message ||= JSON.parse(@msg_str)
+  end
+
+  def druid
+    message.fetch('druid')
+  end
+
+  def sync(object:)
```
justinlittman, line 42: This will sync the version that we need to figure out which email to send.

app/mailers/application_mailer.rb
```diff
@@ -3,4 +3,22 @@
 class ApplicationMailer < ActionMailer::Base
   default from: 'no-reply@sdr.stanford.edu'
   layout 'mailer'
+
+  def set_work
```
justinlittman, line 7: Extracting this for re-use by multiple mailers.

app/services/model_sync/collection.rb
```diff
@@ -28,7 +28,8 @@ def call
 
     def update_params
       {
-        title:
+        title:,
+        version:
```
justinlittman, line 32: Now syncing the version too.

app/services/notifier.rb
```diff
@@ -22,13 +22,13 @@ def self.publish(event_name, **payload)
 
   def self.subscribe_mailer(event_name:, mailer_class:, mailer_method:)
     ActiveSupport::Notifications.subscribe(event_name) do |event|
-      mailer_class.with(**event.payload).send(mailer_method).deliver_later
+      mailer_class.with(**event.payload).send(mailer_method).deliver_later if Settings.notifications.enabled
```
justinlittman, line 25: This will allow us to turn off notifications in some tests where they prove pesky.

config/deploy.rb
```diff
@@ -73,7 +73,7 @@
 namespace :rabbitmq do
   desc 'Runs rake rabbitmq:setup'
   task setup: ['deploy:set_rails_env'] do
-    on roles(:worker) do
+    on roles(fetch(:sneakers_systemd_role)) do
```
justinlittman, line 76: QA doesn't have a worker, so this wasn't working correctly.

app/components/emails/works/searchworks_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Emails
+  module Works
+    # Component for displaying searchworks information for a work.
+    class SearchworksComponent < ApplicationComponent
+    end
+  end
+end
```
mjgiarlo, line 9: Can you share why we'd reach for a view component when there's no logic in it vs. making its HTML above a plain ol' Rails partial?

justinlittman: Consistency. That was as a developer, you know to look for things in components instead of components AND partials. (In the past, I've done exactly that and ended up confused.)

Also, easier to test. And if it does become more complicated, changes are easier.

That being said, not a very strong preference.

mjgiarlo: I buy that. And components are cheap. Much prefer their specs to view specs.

mjgiarlo (approved):
I have a non-blocking question about action mailer and spec peskiness, but otherwise this is shippable 👍🏻

app/services/notifier.rb
justinlittman: Some of the emails requiring hitting SDR to get the cocina. In test, SDR doesn't know anything about the test objects so an error is raised. This happens in the spec for DepositCompleteJob.

SDR could be mocked out in such tests, but since the email is tangential to the thing under test it seems cleaner to just turn off notifications.

mjgiarlo: @justinlittman Ugh. Yes, agree.

PR #517, opened 2025-01-23
Add deposits tab to collection show page

### LINKED ISSUE #276: Add Deposits tab to Collection show page
https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=1114-2956&node-type=frame&t=Ck5zdf8sRjc1YJsx-0

Matching the H2 collection show page, there are 3 tabs for navigation on the show page:
- Collection details
    - Collection information section
    - Details section
    - Links to related information section 
- Collection settings
    - Release and visibility section
    - Terms of use and licenses section
    - Collection participants section
    - Review workflow section
    - Histor section 
- Deposits
    - Works table

mjgiarlo:
Fixes #276

Screenshot from 2025-01-23 10-56-46

### DIFF
--- a/app/components/collections/show/deposits_component.rb
+++ b/app/components/collections/show/deposits_component.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module Collections
+  module Show
+    # Component for rendering a list of deposits for the collection show page.
+    class DepositsComponent < ApplicationComponent
+      def initialize(presenter:)
+        @presenter = presenter
+        super()
+      end
+      attr_reader :presenter
+
+      delegate :collection, :work_statuses, to: :presenter
+
+      def collection_deposits
+        @collection_deposits ||= collection.works.map do |work|
+          {
+            id: work.druid || work.id,
+            values: values_for(work)
+          }
+        end
+      end
+
+      def values_for(work)
+        [
+          helpers.link_to(work.title, link_for(work)),
+          work.user.name,
+          status_for(work),
+          work.object_updated_at ? helpers.l(work.object_updated_at, format: '%b %d, %Y') : nil,
+          persistent_link_for(work)
+        ]
+      end
+
+      def link_for(work)
+        return helpers.work_path(druid: work.druid) if work.druid
+
+        helpers.wait_works_path(work)
+      end
+
+      def status_for(work)
+        work_statuses.fetch(work.druid, VersionStatus::NilStatus.new).status_message
+      end
+
+      def persistent_link_for(work)
+        if work.druid.nil?
+          'N/A (still depositing)'
+        elsif work.doi_assigned?
+          helpers.link_to(nil, Doi.url(druid: work.druid))
+        else
+          helpers.link_to(nil, Sdr::Purl.from_druid(druid: work.druid))
+        end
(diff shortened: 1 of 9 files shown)


app/models/version_status.rb
```diff
@@ -5,13 +5,13 @@
...
   class NilStatus
-    def status_message
-      'Saving'
-    end
+    def status_message = 'Saving'
 
-    def draft?
-      false
-    end
+    def draft? = false
```
justinlittman, line 10: Wow, never seen this syntax before.

app/presenters/collection_presenter.rb
```diff
@@ -63,4 +64,51 @@ def license_label
   def review_workflow_status
     collection.review_enabled ? 'On' : 'Off'
   end
+
+  def works
```
justinlittman, line 68: This all feels like there is a missing view component (instead of being in the presenter).

app/presenters/collection_presenter.rb
```diff
@@ -63,4 +64,51 @@ def license_label
...
+    @works ||= collection.works.to_h do |work|
+      [
+        work.druid || work.id,
+        values_for(work)
+      ]
+    end
+  end
+
+  def work_statuses
+    @work_statuses ||= Sdr::Repository.statuses(
```
justinlittman, line 78: Querying SDR belongs in the controller, not presenter.

justinlittman (commented):
I didn't get very far -- it really feels like the presenter is bearing some of the responsibility of the controller and is getting some complexity that would be better handled by a view component for that specific table.

app/models/version_status.rb
mjgiarlo: @justinlittman Yeah, it's new as of Ruby 3.x.

app/presenters/collection_presenter.rb
mjgiarlo: I'll look into extracting a component.

spec/controllers/.keep
mjgiarlo, line 1: Cruft.

app/components/collections/show/deposits_component.rb
```diff
@@ -0,0 +1,67 @@
...
+          content.nil? ? 'N/A (still depositing)' : content.content_files.count,
+          content.nil? ? 'N/A (still depositing)' : helpers.number_to_human_size(content.content_files.sum(:size)),
+          persistent_link_for(work)
+        ]
+      end
+
+      def content_for(work)
+        return if work.druid.blank?
+
+        Contents::Builder.call(
```
justinlittman, line 40: 1. We shouldn't be doing `Sdr::Repository.find` in a component.
2. We can get the file count and sizes from the cocina without calling `Contents::Builder`.
3. I'd prefer not to do any of this, as it is going to be very expensive. Let me ask.

justinlittman (changes requested):

app/components/collections/show/deposits_component.rb
mjgiarlo: Sounds good, @justinlittman

justinlittman (approved):

PR #522, opened 2025-01-24
Collection show page

### LINKED ISSUE #474: Make collection details page match mockup
See Figma: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1114-2956&t=JkyqI1jxuts9EwH9-4

Some information is missing (being able to see the PURL would be nice) and not everything is displayed with the indicated labels or in the right order.

This includes adding the "Deposit to this collection" button.

lwrubel:
Resolves #474 to align collection show page with mockups. Moving the buttons upwards to make more responsive will be in a follow-up PR once the design is settled. 

Also adds a pencil icon for editing and a Deposit to this collection button. 

Screenshot 2025-01-24 at 10 23 39 AM


app/components/collections/show/header_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="d-flex justify-content-between">
```
lwrubel, line 1: Instead of sharing the header_component across works and collections, split it out to more easily accommodate collection-specific buttons and behavior. Also allows testing with collections and works separately.

app/models/version_status.rb
```diff
@@ -35,6 +35,10 @@ def draft?
     status.open?
   end
 
+  def first_draft?
+    status.open? && status.version == 1
+  end
```
lwrubel, line 40: Is this logic correct? I couldn't find any existing status that matched the situation where the initial draft is not yet deposited.

mjgiarlo: I'd have to defer to @justinlittman. I'm not sure when we'd lean on the status vs. the deposit state machine.

app/components/collections/show/header_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div>
+    <%= render Elements::HeadingComponent.new(level: :h1, text: title, classes: 'mb-4 d-inline-block') %>
+    <%= render Edit::EditIconButtonComponent.new(presenter:, classes: 'mx-2 py-0 mb-2', icon_classes: 'h4') %>
+
+    <span class="ms-4 fst-italic status"><%= status_message %></span>
+  </div>
+  <div class="d-flex align-items-start">
+    <%= render Edit::EditDraftButtonComponent.new(presenter:) %>
+    <%= render Edit::DiscardDraftButtonComponent.new(presenter:, classes: 'ms-2') %>
+    <% unless presenter.first_draft? %>
```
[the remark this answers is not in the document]
justinlittman: It would be nice to hide some of the code in the view component class instead in the erb. For example, `<% unless first_draft? %>`

app/components/collections/show/header_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+      <% if helpers.allowed_to?(:create_work?, presenter.collection) %>
+        <%= render Elements::ButtonLinkComponent.new(link: new_work_path(collection_druid: presenter.druid), label: 'Deposit to this collection', variant: :primary, classes: 'ms-2') %>
```
justinlittman, line 13: Could this just be `<%= render Elements::ButtonLinkComponent.new(...) if !first_draft? && allowed_to_create_work? %>`?

app/components/elements/tables/row_component.rb
```diff
@@ -5,10 +5,11 @@ module Tables
     # Component for rendering a table row.
     class RowComponent < ApplicationComponent
       renders_many :cells
+      renders_many :items
 
       def initialize(label: nil, values: [], id: nil)
         @label = label
-        # Provide either values or cells.
+        # Provide either values, cells, or items.
```
justinlittman, line 12: Can you expand this comment a bit to explain what each of them are?

app/models/version_status.rb
justinlittman: Yes, this logic is correct. Can you use this method in L23 to avoid dupe?

The deposit state machine has fairly specific uses. Whenever possible the version status should be used as it reflects the current status from SDR.

app/presenters/collection_presenter.rb
```diff
@@ -63,4 +63,19 @@ def license_label
   def review_workflow_status
     collection.review_enabled ? 'On' : 'Off'
   end
+
+  def purl_link
+    # No druid yet, so there's no PURL link yet either. Collection is likely still depositing.
+    return if druid.blank?
+
+    link_to(nil, Sdr::Purl.from_druid(druid:), target: '_blank', rel: 'noopener')
```
justinlittman, line 71: Extract an external link into a helper to avoid having to dupe the target and rel?

justinlittman (commented):
Minor comments / suggestions.

app/presenters/collection_presenter.rb
lwrubel: I'm not sure where the target and rel are being duped--could you explain where this might be extracted to?

app/components/elements/tables/row_component.rb
lwrubel: Added some more description.

justinlittman (approved):

app/presenters/collection_presenter.rb
justinlittman: https://github.com/sul-dlss/hungry-hungry-hippo/issues/529

PR #535, opened 2025-01-28
Refactor toggle component.

justinlittman:


app/components/contributors/edit_component.html.erb
```diff
@@ -1,20 +1,18 @@
 <%= tag.div data: contributors_data do %>
   <%= render Elements::Forms::ToggleComponent.new(form:, field_name: :role_type, label_classes: 'fw-bold',
                                                   label: t('works.edit.fields.contributors.role_type.label'), container_classes: 'mb-4') do |component| %>
-    <% component.with_toggle_option(form:,
+    <% component.with_left_toggle_option(form:,
```
justinlittman, line 4: There's always going to be 2 toggle options, so renaming to `left_toggle_option` and `right_toggle_option`.

app/components/elements/forms/toggle_component.rb
```diff
@@ -4,7 +4,22 @@ module Elements
   module Forms
     # Component for a toggle-like radio button group field
     class ToggleComponent < FieldComponent
-      renders_many :toggle_options, Elements::Forms::ToggleOptionComponent
+      renders_one :left_toggle_option, lambda { |form:, field_name:, label:, value:, label_data: {}, data: {}|
+        label_data[:action] = merge_actions(label_data[:action], 'click->toggle#leftSelected')
```
justinlittman, line 8: Inserting the data values necessary for the toggle controller.

app/forms/author_form.rb
```diff
@@ -19,7 +19,8 @@ class AuthorForm < ApplicationForm
   attribute :organization_role, :string
   validates :organization_role, presence: true, unless: :person?
 
-  attribute :role_type, :string
+  attribute :role_type, :string, default: 'person'
+  validates :role_type, inclusion: { in: %w[person organization] }
```
justinlittman, line 23: The validates isn't really necessary, but it documents the expected values.

app/javascript/controllers/contributors_controller.js
```diff
@@ -2,7 +2,7 @@ import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
   static targets = [
-    'contributorTypePerson', 'contributorTypeOrganization', 'contributorTypePersonLabel',
+    'contributorTypePerson', 'contributorTypeOrganization',
```
justinlittman, line 5: The toggle-y parts have been moved to the toggle controller.

app/javascript/controllers/toggle_controller.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  leftSelected () {
+    this.selectRadio(this.leftRadioTarget, this.leftLabelTarget)
+    this.unselectRadio(this.rightRadioTarget, this.rightLabelTarget)
+  }
+
+  rightSelected () {
+    console.log('rightCheckboxSelected')
```
lwrubel, line 22: cruft?

lwrubel:
Approved with a minor point.

lwrubel (approved):

app/components/application_component.rb
```diff
@@ -9,4 +9,12 @@ class ApplicationComponent < ViewComponent::Base
...
     ComponentSupport::CssClasses.merge(*)
   end
+
+  # Merge data-actions together.
+  #
+  # @param args [Array<String>, String] The actions to merge (array, classes, space separated classes).
+  # @return [String] The merged classes.
+  def merge_actions(*)
+    ComponentSupport::CssClasses.merge(*)
+  end
```
mjgiarlo, line 19: Identical implementation to `#merge_classes`?

app/javascript/controllers/contributors_controller.js
mjgiarlo: Nice!

app/components/contributors/edit_component.html.erb
jcoyne: What about RTL langs? 😆

mjgiarlo: @jcoyne Coyne.

justinlittman: They are in the current designs.

app/components/application_component.rb
justinlittman: Yes. Even though the implementation is currently the same, having separate method names is clearer.

mjgiarlo: @justinlittman No argument there. What I'm concluding: this change reveals that `Component::CssClasses.merge` has little (nothing?) to do with CSS classes. I suspect we named it `CssClasses` because of the output it was receiving, which now includes both CSS classes and actions (and may include more as we go along). The method is more about unique string joins, apparently. We could consider renaming the containing class?

PR #541, opened 2025-01-29
Extract link to new tab into helper.

### LINKED ISSUE #529: Create a link_to_new_tab helper
To avoid having `link_to(... target: '_blank', rel: 'noopener')` duplicated in the code.

See https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/presenters/collection_presenter.rb#L71

justinlittman:
closes #529


app/presenters/collection_presenter.rb
```diff
@@ -2,6 +2,8 @@
 
 # Presents a collection
 class CollectionPresenter < FormPresenter
+  include LinkHelper
```
mjgiarlo, line 5: Can we drop this `include` since it's also included in the parent class?

justinlittman: I forgot to hit save ...

app/helpers/link_helper.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Helper for creating links
+module LinkHelper
+  def link_to_new_tab(name, url, **)
+    link_to(name, url, target: '_blank', rel: 'noopener', **)
+  end
```
mjgiarlo, line 7: How would you feel about:

```suggestion
  def link_to_new_tab(url:, name: nil, **)
    link_to(name, url, target: '_blank', rel: 'noopener', **)
  end
```

That way consumers don't need to pass a `nil` name if all they have is a URL. Pass what you have/want and let the helper default the rest. So, in the authors show component above we'd have `helpers.link_to_new_tab(url: author.orcid)` for example.

mjgiarlo (commented):
Pondering a slightly diff API for our internal method

app/helpers/link_helper.rb
justinlittman: I think it would likely surprise a developer if it had a different API than `link_to` (especially since most of our links use plain-old `link_to`).

mjgiarlo: @aaron-collier @lwrubel have thoughts about ☝🏻❓ I could go either way

aaron-collier: @mjgiarlo / @justinlittman I don't have strong feelings about the order of params, but personally prefer kwargs with defaults for the reason Mike mentioned as well as the explicitness of "name is this" kind of approach. That said, I'm happy to go with either approach here and we can refactor/follow up if we find it difficult or confusing.

lwrubel: I'm not sure this even has to be componentized, since the existing `link_to` is familiar and there's minimal code that needs repetition. But maybe there are some use cases I'm not thinking of. 

If we are going to componentize, then kwargs in whatever order work for me.

justinlittman: I greatly prefer kwargs as well. However, I think it might be surprising for `link_to` and `link_to_new_tab` to take different arguments.

lwrubel: Oops, I called this a component, when it's a helper. I don't feel strongly about which way we go.

aaron-collier (approved):

PR #548, opened 2025-01-30
Rename author to contributors and normalize pluralization in text / labels.

justinlittman:


app/services/to_work_form/mapper.rb
```diff
@@ -94,7 +94,7 @@ def publication_date_params
       event_date = CocinaSupport.first_event_date_for(cocina_object:, type: 'publication')
       return {} if event_date.nil?
 
-      # Publication date is always a single date
+      # Publication date is always a single datecccccbg
```
mjgiarlo, line 97: Typo?

mjgiarlo (approved):
Approved pending typo removal

app/services/to_work_form/mapper.rb
justinlittman: covfefe

That's when I touch my yubi key.

PR #560, opened 2025-01-31
Refactors CocinaSupport.

### LINKED ISSUE #364: Spike on refactoring CocinaSupport and CocinaDescriptionSupport

justinlittman:
closes #364


app/services/to_collection_form/mapper.rb
```diff
@@ -26,8 +26,8 @@ def params
         lock: cocina_object.lock,
         title: CocinaSupport.title_for(cocina_object:),
         description: CocinaSupport.abstract_for(cocina_object:), # Cocina abstract maps to Collection description
-        contact_emails_attributes: CocinaSupport.contact_emails_for(cocina_object:),
-        related_links_attributes: CocinaSupport.related_links_for(cocina_object:),
+        contact_emails_attributes: ToForm::ContactEmailsMapper.call(cocina_object:),
+        related_links_attributes: ToForm::RelatedLinksMapper.call(cocina_object:),
```
mjgiarlo, line 30: (This is why I asked the question above. We see here a mix of using `CocinaSupport#method` and `ToForm::FooMapper.call`.)

app/services/to_form/base_mapper.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module ToForm
+  # Base class for mappers that convert Cocina::Models to form values
```
mjgiarlo, line 4: Form objects? Form instances? But yeah.

app/services/cocina_support.rb
```diff
@@ -1,88 +1,27 @@
 # frozen_string_literal: true
 
 # Helpers for working with Cocina objects
-class CocinaSupport # rubocop:disable Metrics/ClassLength
+class CocinaSupport
```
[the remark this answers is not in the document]
justinlittman: There are a handful of methods that are used outside of To*Form. I think they will remain in CocinaSupport, something like:
```
# Helpers for working with Cocina objects
class CocinaSupport
  # These are methods used to extract data from Cocina objects that are used outside To*Form classes.
  # Extraction methods that are only used in To*Form classes should NOT be here.
  def self.title_for(cocina_object:)
    ToForm::TitleMapper.call(cocina_object:)
  end

  def self.collection_druid_for(cocina_object:)
    ToWorkForm::CollectionDruidMapper.call(cocina_object:)
  end
```

app/services/to_collection_form/mapper.rb
justinlittman: Within To*Form classes, you will not see CocinaSupport. Outside To*Form classes, you may see a small amount of CocinaSupport.

We'll see how this works and can adjust if it ends up unclear.

justinlittman: The crux of the problem is that we're trying to extract metadata from cocina AND map to form data structures. Either we strictly separate these, which create a LOT of extra code or we accept a little inelegancy when only trying to extract metadata from cocina.

app/services/cocina_support.rb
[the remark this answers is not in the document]
justinlittman: I changed my mind. CocinaSupport is gone. Some simple, shared parsing methods go in `Cocina::Parsers`.

app/services/to_form/note_mapper.rb
```diff
@@ -0,0 +1,34 @@
...
+      new(...).call
+    end
+
+    def self.abstract(cocina_object:)
+      new(cocina_object:, type: ABSTRACT_TYPE).call
+    end
+
+    def self.citation(cocina_object:)
+      new(cocina_object:, type: CITATION_TYPE).call
+    end
```
mjgiarlo, line 19: I guess you deviated from the `.call` pattern here for DRY reasons? I expected `ToForm::AbstractMapper.call` for instance.

mjgiarlo (approved):
Approve pending question

app/services/to_form/note_mapper.rb
justinlittman: There is a `.call` method here. These are just for convenience (and an alternative to adding a class hierarchy).

PR #570, opened 2025-02-03
Adjust pane buttons.

justinlittman:
closes #467

PR #571, opened 2025-02-03
Discard collection drafts.

### LINKED ISSUE #524: Remove ability of users to save collections as draft
When modifying collection details, they will need to save or cancel. There will be no option to "save as draft." This means collections will only ever exist in a deposited state.

One implication of this is that the collection will always have just two buttons at the top of the view page: "Edit" and "+Deposit to this collection."

@astridu Please drop mockups in here when you have them. Thanks.

justinlittman:
closes #524

Note that part of this work was included in #570.


app/controllers/application_controller.rb
```diff
@@ -22,18 +22,7 @@ def deny_access
     redirect_to main_app.root_path
   end
 
-  def deposit?
```
justinlittman, line 25: Moving these to WorksController.

aaron-collier (approved):

PR #572, opened 2025-02-03
Support Stanford as a degree granting institution.

### LINKED ISSUE #117: Add Stanford? option for degree-granting institution selection for organization authors
- We are only capturing organizational affiliations for when a user selects the role as "degree granting institution."
- When this role is selected for an org, we display the question asking if Stanford is the degree-granting institution.
- If they answer no, then they are asked to type in the institution name. This happens rarely (it's almost always Stanford) and we are less concerned about collecting the persistent identifier for non-Stanford entries, so we decided this text field with no PID was just fine.
- If they answer yes, then we should be collecting "Stanford University" as the text entry for this organization, along with the ROR identifier, though we do not display the ROR (or other schema info we will save) to the user. We also give them the option to tell us what department they are getting the degree from. This is a free-text field.
- I think all the necessary mapping information is in this old ticket: https://github.com/sul-dlss/happy-heron/issues/3099
- Additionally, if we are generating a DOI, the affiliation information should be sent along with that. We should always send all affiliation info (especially the PID) to DataCite.
- I believe this design correctly reflects the bits the user sees: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=1116-3980&t=tYYowEXOoxH1DAWf-4

justinlittman:
closes #117

https://github.com/user-attachments/assets/6ef036fa-2137-4c5e-b08d-752529022489

### DIFF
--- a/app/components/elements/forms/field_component.rb
+++ b/app/components/elements/forms/field_component.rb
@@ -4,9 +4,9 @@ module Elements
   module Forms
     # Base component for all form fields.
     class FieldComponent < ApplicationComponent
-      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists
-                     disabled: false, hidden: false, data: {}, placeholder: nil, width: nil, label_classes: [],
-                     container_classes: [], input_classes: [])
+      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists, Metric/MethodLength
+                     disabled: false, hidden: false, data: {}, input_data: {}, placeholder: nil, width: nil,
+                     label_classes: [], container_classes: [], input_classes: [])
         @form = form
         @field_name = field_name
         @required = required
@@ -16,6 +16,7 @@ module Elements
         @hidden = hidden
         @disabled = disabled
         @data = data
+        @input_data = input_data
         @placeholder = placeholder
         @width = width
         @label_classes = label_classes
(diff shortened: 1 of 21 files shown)


app/components/elements/forms/field_component.rb
```diff
@@ -4,9 +4,9 @@ module Elements
   module Forms
     # Base component for all form fields.
     class FieldComponent < ApplicationComponent
-      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists
-                     disabled: false, hidden: false, data: {}, placeholder: nil, width: nil, label_classes: [],
-                     container_classes: [], input_classes: [])
+      def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists, Metric/MethodLength
+                     disabled: false, hidden: false, data: {}, input_data: {}, placeholder: nil, width: nil,
```
justinlittman, line 8: There was an inconsistency about whether `data` was for the container or the input. This makes it consistent.

app/javascript/controllers/contributors_controller.js
```diff
@@ -2,10 +2,13 @@ import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
   static targets = [
-    'contributorTypePerson', 'contributorTypeOrganization',
-    'contributorTypeOrganizationLabel', 'selectPersonRole', 'selectOrganizationRole', 'personName',
-    'organizationName', 'orcidField', 'useOrcidButton', 'resetOrcidButton', 'firstNameField',
-    'lastNameField'
+    'personOption', 'organizationOption',
```
justinlittman, line 5: I did a bunch of refactoring on this controller. The orcid code wasn't touched.

aaron-collier (approved):

PR #576, opened 2025-02-04
Increase the margin after the masthead

thatbudakguy:
Adds more space before the breadcrumbs show up, as per slack
discussion.


thatbudakguy:
For the darkening of the masthead border, see https://github.com/sul-dlss/component-library/pull/116.

justinlittman (approved):

PR #579, opened 2025-02-04
Fixes orcid input for work form.

### LINKED ISSUE #469: Update ORCID entry options
We will use Astrid's mockup with the toggle. See https://www.figma.com/proto/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1116-3660&t=hlKSwO2eX4CNOQTR-0&scaling=min-zoom&content-scaling=fixed&page-id=1092%3A2227&starting-point-node-id=1114%3A2853 

Defaults state for ORCID toggle is "Yes."
- _Shown in second contributor entry in above mockup_: Name fields should be disabled for editing until they have been populated.
-- Include note next to name fields that says "Name fields will be autopopulated based on the ORCID entry above."
- _Shown in first contributor entry in this mockup:_ Entry Option 1
-- When user clicks "Use my ORCID iD" the ORCID iD field and the name fields are populated. 
-- Both ORCID and names will be editable. 
-- A note is displayed below the ORCID field that says "Name associated with this ORCID iD is [name retrieved from ORCID]." This replaces the note about autopopulating that is present in the default state.
-- The "Use my ORCID iD" option should show up for every author entry.
- _Shown in first contributor entry in this mockup:_ Entry Option 2
-- User pastes an ORCID iD into the box, which populates the name fields. 
-- Both ORCID and names will be editable. 
-- A note is displayed below the ORCID field that says "Name associated with this ORCID iD is [name retrieved from ORCID]." (same as for entry option 1) This replaces the note about autopopulating that is present in the default state.

User can change toggle to "No."
- This state displays First and Last name entry boxes. See image below.

Image

justinlittman:
closes #469


app/components/contributors/edit_component.html.erb
justinlittman, line 1: Moved to `works/edit/contributor_component`

config/locales/en.yml
```diff
@@ -148,29 +148,6 @@ en:
             A unique, descriptive title may improve discovery of your work in web searches.
         license:
           help_text: Assigning a license may improve discovery of your work in web searches.
-        contributors:
```
justinlittman, line 151: Pulled these out of `en.yml` since they're only used in one place.

aaron-collier (approved):

PR #616, opened 2025-02-05
Add collection license tab

### LINKED ISSUE #272: Add License tab to collection form
https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=789-4185&node-type=frame&t=3O5r5DrUmDQn71IY-0

aaron-collier:
Fixes #272 

Note: I'm curious if saving the license to cocina for a collection makes sense. While I'm pretty sure it's not really used at all in H3, perhaps it is used elsewhere so we need to include it.

Screenshot 2025-02-05 at 2 00 52 PM

### DIFF
--- a/app/components/collections/edit/license_component.html.erb
+++ b/app/components/collections/edit/license_component.html.erb
@@ -1,0 +1,23 @@
+<fieldset>
+  <div class="form-check d-flex pt-2">
+    <div class="ps-2">
+      <%= form.radio_button :license_option, 'required', class: 'form-check-input' %>
+      <%= form.label :license_option_required, 'Require license for all deposits', class: 'form-check-label' %>
+    </div>
+    <div class="ps-2">
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
+    </div>
+    <div class="ps-2">
+      <a href="https://sdr.library.stanford.edu/documentation/license-options">Get help selecting a license</a>
+    </div>
+  </div>
+  <div class="form-check d-flex pt-2">
+    <div class="ps-2">
+      <%= form.radio_button :license_option, 'depositor_selects', class: 'form-check-input' %>
+      <%= form.label :license_option_depositor_selects, 'Depositor selects license', class: 'form-check-label' %>
+    </div>
+    <div class="ps-2">
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :suggested_license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
+    </div>
+  </div>
+</fieldset>
--- a/app/components/collections/edit/license_component.html.erb
+++ b/app/components/collections/edit/license_component.html.erb
@@ -17,7 +17,7 @@
       <%= form.label :license_option_depositor_selects, 'Depositor selects license', class: 'form-check-label' %>
     </div>
     <div class="ps-2">
-      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :suggested_license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :default_license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
     </div>
   </div>
 </fieldset>
--- a/app/components/collections/edit/license_component.html.erb
+++ b/app/components/collections/edit/license_component.html.erb
@@ -5,10 +5,10 @@
       <%= form.label :license_option_required, 'Require license for all deposits', class: 'form-check-label' %>
     </div>
     <div class="ps-2">
-      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, hidden_label: true, label:, width: 250, container_classes: 'mb-4') %>
     </div>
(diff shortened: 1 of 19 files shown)


app/components/collections/edit/license_component.html.erb
```diff
@@ -0,0 +1,23 @@
+<fieldset>
+  <div class="form-check d-flex pt-2">
+    <div class="ps-2">
+      <%= form.radio_button :license_option, 'required', class: 'form-check-input' %>
```
justinlittman, line 4: The required attribute is set with `required: true` for `form.radio_button`.

However, this will be client side validation which we don't want. Instead, can you set a default value in the form (which will set a default radio button)?

app/components/collections/edit/license_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <%= form.label :license_option_required, 'Require license for all deposits', class: 'form-check-label' %>
+    </div>
+    <div class="ps-2">
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
```
justinlittman, line 8: Can you not set required, as that will result in client side validation?

Why is the label_classes set to `h5`?

app/components/collections/edit/license_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    </div>
+    <div class="ps-2">
+      <a href="https://sdr.library.stanford.edu/documentation/license-options">Get help selecting a license</a>
```
justinlittman, line 11: Can you use `link_to_new_tab` helper?

app/components/collections/edit/license_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    </div>
+  </div>
+  <div class="form-check d-flex pt-2">
+    <div class="ps-2">
+      <%= form.radio_button :license_option, 'depositor_selects', class: 'form-check-input' %>
+      <%= form.label :license_option_depositor_selects, 'Depositor selects license', class: 'form-check-label' %>
+    </div>
+    <div class="ps-2">
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :default_license, options: license_options, required: true, hidden_label: true, label:, width: 250, container_classes: 'mb-4', label_classes: 'h5') %>
```
justinlittman, line 20: Same as above.

app/forms/collection_form.rb
```diff
@@ -29,6 +29,19 @@ def persisted?
   attribute :access, :string, default: 'world'
   validates :access, inclusion: { in: %w[world stanford depositor_selects] }
 
+  attribute :license_option, :string, default: 'required'
+  validates :license_option, inclusion: { in: %w[required depositor_selects] }
+
+  attribute :license, :string
+  with_options if: -> { license_option == 'required' } do
```
justinlittman, line 36: There is only one validation, so `with_options` not necessary.

app/forms/collection_form.rb
```diff
@@ -29,6 +29,19 @@ def persisted?
...
+    validate :license_present
+  end
+
+  attribute :default_license
+  with_options if: -> { license_option == 'depositor_selects' } do
```
justinlittman, line 41: Same.

app/forms/collection_form.rb
```diff
@@ -46,3 +59,15 @@ def duration_must_be_present
 
   errors.add(:release_duration, 'select a valid duration for release')
 end
+
+def license_present
```
justinlittman, line 63: Can't you just use a presence validator instead of custom?

app/controllers/collections_controller.rb
```diff
@@ -45,6 +45,8 @@ def create
                                       release_duration: @collection_form.release_duration,
                                       access: @collection_form.access,
                                       doi_option: @collection_form.doi_option,
+                                      license_option: @collection_form.license_option,
```
justinlittman, line 48: If I understand correctly, many of the values being set here are updated in the DepositCollectionJob. If that is the case, then only need to create the minimal Collection here.

app/services/to_cocina/collection/mapper.rb
```diff
@@ -46,6 +46,12 @@ def request_params
           params_hash[:administrative][:partOfProject] = Settings.project_tag
         end
       end
+
+      def access
+        return { view: 'world' } if collection_form.license.blank?
```
justinlittman, line 51: `{view: 'world', license: collection_form.license}.compact` would be more concise than L51-53

app/services/to_collection_form/mapper.rb
```diff
@@ -48,5 +49,11 @@ def participant_attributes(role)
         { sunetid: participant.sunetid }
       end
     end
+
+    def license
```
justinlittman, line 53: Can you call this `license_params`?

app/views/collections/form.html.erb
```diff
@@ -45,6 +46,13 @@
       <% end %>
     <% end %>
 
+    <% component.with_pane do %>
+      <%= render Collections::Edit::PaneComponent.new(tab_name: :license, label: I18n.t('collections.edit.panes.license.label')) do %>
+        <p><%= t('collections.edit.panes.license.help_text') %></p>
```
justinlittman, line 51: If this is only used here, put inline instead of in en.yml.

justinlittman (changes requested):

app/components/collections/edit/license_component.html.erb
aaron-collier: re: 'h5' - copy pasta, I missed that was getting set.

app/components/collections/edit/license_component.html.erb
aaron-collier: I may be misunderstanding, but in this case, the string `'required'` is the value and relates to the label `license_option_required`...

justinlittman: No, I misunderstood. This is correct.

app/views/collections/form.html.erb
justinlittman: Let's just leave as is. On the other hand, I don't want large blocks of text in form.html.erb either.

justinlittman (approved):

Issue #544, opened 2025-01-29
SUNet ID lookup for collection participant entry


amyehodge:
When collection manager enters a SUNet ID for a participant they want to add as a manager, depositor, or reviewer, look up SUNet ID to make sure it's valid like we do in H2. Screenshot of H2 below.

<img width="668" alt="Image" src="https://github.com/user-attachments/assets/10d29480-ec86-49bf-8737-914dbf6a63df" />

thatbudakguy:
todo:
- [ ] port sunet input stimulus controller from H2

aaron-collier:
Blocked by waiting on UIT for a cert (cc: @tallenaz)

Issue #578, opened 2025-02-04
Collection participant entry should match designs


amyehodge:
Current implementation does not match the designs.

Figma: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=789-4316&t=Y1CKoKbU4dxdK8Cy-4

PR #631, opened 2025-02-06
Set up sunetid lookups for collection contributors

thatbudakguy:
Ref #544 

https://github.com/user-attachments/assets/85771c7a-9a57-4c28-a952-3cfbc9105fef


app/components/collections/contributor_component.rb
justinlittman, line 1: Renamed to participant_component.

justinlittman:
This is on hold until UIT provides certs to Ops. @tallenaz is assisting with this.

app/services/account_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    new(...).call
+  end
+
+  def initialize(sunetid:)
+    @sunetid = sunetid
+  end
+
+  # @return [Account, nil] the account or nil if not found
+  def call
+    params = Rails.cache.fetch(sunetid, namespace: 'account', expires_in: 1.month) do
```
aaron-collier, line 17: A month seems like a long time to cache, but probably not an issue for sunet -> user...

aaron-collier (approved):
Approved pending minor comments. I don't think this quite brings things inline with the design expected in https://github.com/sul-dlss/hungry-hungry-hippo/issues/578 but that can be a follow up.

app/services/account_service.rb
justinlittman: That's what it was in H2.

app/controllers/accounts_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    authorize! with: AccountPolicy
+
+    @account = lookup
+    render layout: false
+  end
+
+  private
+
+  # Does a lookup from the account service in production mode, otherwise examines local database for users
+  def lookup
```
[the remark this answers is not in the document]
justinlittman: Makes sense. Fixing.

PR #633, opened 2025-02-06
Add workflow tab to collection form

### LINKED ISSUE #275: Add Review Workflow tab to the collection form
https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=789-4342&node-type=frame&t=3O5r5DrUmDQn71IY-0

aaron-collier:
Fixes #275 

https://github.com/user-attachments/assets/0283f9dd-b272-4d19-b22c-f8a738c622d0

Text added to form after screen record:

Screenshot 2025-02-06 at 12 31 26 PM

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -36,7 +36,7 @@ class CollectionsController < ApplicationController
     render :form
   end
 
-  def create
+  def create # rubocop:disable Metrics/AbcSize
     authorize! Collection
     @collection_form = CollectionForm.new(**collection_params)
     if @collection_form.valid?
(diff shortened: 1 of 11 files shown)


app/controllers/collections_controller.rb
```diff
@@ -46,7 +46,8 @@ def create
                                       access: @collection_form.access,
                                       doi_option: @collection_form.doi_option,
                                       user: current_user,
-                                      deposit_state_event: 'deposit_persist')
+                                      deposit_state_event: 'deposit_persist',
+                                      review_enabled: @collection_form.review_enabled)
```
justinlittman, line 50: Since this is also happening in the deposit job does it also need to be here?

justinlittman (approved):
Approved pending consideration of comment.

app/controllers/collections_controller.rb
aaron-collier: Nope - apparently doesn't need to be. Fixed.

PR #637, opened 2025-02-06
Allow collection depositor to view but not edit collecion settings

### LINKED ISSUE #625: Depositor should be able to view but not edit collection pages
Maricela is a deposit on the "Review Workflow Test Collection." From the dashboard she clicked on the title of the collection, which should take her to the collection details page, she saw the page below. She should also be able to view the settings and deposits tabs for the collection. 

Image

aaron-collier:
Fixes #625 

As the creator the edit button is available:

Screenshot 2025-02-06 at 2 15 03 PM

And now as a depositor you can see the page, but there is no edit button:

Screenshot 2025-02-06 at 2 14 20 PM

### DIFF
--- a/app/components/collections/show/header_component.html.erb
+++ b/app/components/collections/show/header_component.html.erb
@@ -1,12 +1,12 @@
 <div class="d-flex justify-content-between">
   <div>
     <%= render Elements::HeadingComponent.new(level: :h1, text: title, classes: 'mb-4 d-inline-block') %>
-    <%= render Edit::EditIconButtonComponent.new(presenter:, classes: 'mx-2 py-0 mb-2', icon_classes: 'h4') if helpers.allowed_to?(:edit?, presenter.collection) %>
+    <%= render Edit::EditIconButtonComponent.new(presenter:, classes: 'mx-2 py-0 mb-2', icon_classes: 'h4') if edit? %>
 
     <span class="ms-4 fst-italic status"><%= status_message %></span>
   </div>
   <div class="d-flex align-items-start">
-    <%= render Edit::EditDraftButtonComponent.new(presenter:, label: 'Edit') if helpers.allowed_to?(:edit?, presenter.collection) %>
+    <%= render Edit::EditDraftButtonComponent.new(presenter:, label: 'Edit') if edit? %>
     <% if !presenter.first_draft? && allowed_to_create_work? %>
       <%= render Elements::ButtonLinkComponent.new(link: new_work_path(collection_druid: presenter.druid),
                                                    label: 'Deposit to this collection',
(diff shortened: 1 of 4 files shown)


app/presenters/collection_presenter.rb
```diff
@@ -79,4 +79,10 @@ def created_by
   def created_datetime
     collection.created_at.localtime.strftime('%b %d, %Y, %l:%M%p %Z')
   end
+
+  def editable?
+    return false if collection.depositors.include? Current.user
```
justinlittman, line 84: This should be an ActionPolicy allowed_to? instead of duplicating auth logic here.

justinlittman (changes requested):

app/components/collections/show/header_component.html.erb
```diff
@@ -1,12 +1,12 @@
 <div class="d-flex justify-content-between">
   <div>
     <%= render Elements::HeadingComponent.new(level: :h1, text: title, classes: 'mb-4 d-inline-block') %>
-    <%= render Edit::EditIconButtonComponent.new(presenter:, classes: 'mx-2 py-0 mb-2', icon_classes: 'h4') %>
+    <%= render Edit::EditIconButtonComponent.new(presenter:, classes: 'mx-2 py-0 mb-2', icon_classes: 'h4') if helpers.allowed_to?(:edit?, presenter.collection) %>
```
aaron-collier, line 4: Putting these `if...allowed_to?` here made the most sense. I tried a couple of different patterns for using the component `render?` method but it felt pretty convoluted passing the results of `allowed_to?` through the controller/presenter. If we'd prefer a pattern like that I can do that as well, but this seems the most straightforward.

justinlittman: Minor, but I would put the `helpers.allowed_to?(:edit?, presenter.collection)` logic in the class part of the component in a method called `edit?`. It keeps logic out of the erb and makes it easier to read.

app/components/collections/show/header_component.html.erb
```diff
@@ -1,12 +1,12 @@
...
 
     <span class="ms-4 fst-italic status"><%= status_message %></span>
   </div>
   <div class="d-flex align-items-start">
-    <%= render Edit::EditDraftButtonComponent.new(presenter:, label: 'Edit') %>
+    <%= render Edit::EditDraftButtonComponent.new(presenter:, label: 'Edit') if helpers.allowed_to?(:edit?, presenter.collection) %>
```
justinlittman, line 9: Same here.

justinlittman (approved):
Pending consideration of comments.

app/components/collections/show/header_component.html.erb
aaron-collier: That's a good idea, I had gone a bit further than that and it felt convoluted, but this feels more like what I was hoping for. :+1:

PR #638, opened 2025-02-06
Render contact form in place of dashboard for unauthorized user

### LINKED ISSUE #446: Add welcome contact form for new user
Figma design: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1114-2611&t=PWgVYup7Cne4Tl7C-4

1. How does one get to that page? **One gets to this page if they do not have access to the dashboard, upon clicking "Enter here" on the home page.**
1. What are the menu options for the "How can we help you?" select box?
    * **'I want to become an SDR depositor'**
    * **'I want to report a problem'**
    * **'I want to ask a question'**
    * **'I want to provide feedback'**
    * **'Request access to another collection'**
1. Should we pre-populate any fields for the user, e.g., name, email address? **No**
1. When a user cancels the form, presumably we navigate back to the home page? **Yes**
1. When a user submits the form, what happens to the information? **It should be sent via e-mail to both sdr-support@lists.stanford.edu and sdr-contact@lists.stanford.edu mailing lists (which will create a JIRA ticket in the SDRSUPPORT project). When the form is successfully submitted, the user should see a page that says 'thank you for submitting the form and someone will get back to you soon' (look for text similar to this in H2)**

mjgiarlo:
Fixes #446 

Includes:
* **Make user roles overrideable in dev mode**
* **Render contact form in place of dashboard for unauthorized user**

Screenshot from 2025-02-06 14-40-43


Gemfile
```diff
@@ -79,6 +79,7 @@ end
 
 group :development do
   gem 'hotwire-spark'
+  gem 'letter_opener' # Preview ActionMailer messages
```
mjgiarlo, line 82: I can be convinced that popping up a window showing each formatted email in the browser as it would be sent is too heavyweight, in which case I'd opt for changing `delivery_method` in dev to `:file` and then documenting in the README where Rails stuffs those.

Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: ROLES=dlss:hydrus-app-administrators bin/rails server
+web: bin/rails server
```
mjgiarlo, line 1: Don't hardcode the dev role.

README.md
```diff
@@ -23,6 +23,33 @@ bin/dev
...
+
+```shell
+ROLES=dlss:hydrus-app-collection-creators bin/dev
+```
+
+To view the site as a first-time user:
+
+```shell
+ROLES=foo:bar:whatever bin/dev
+```
```
mjgiarlo, line 51: More or less lifted from the H2 README.

README.md
```diff
@@ -117,8 +144,8 @@ A "nested" field is one that is not "simple" (see prior section).
...
 1. Add a subscription to `config/initializers/subscriptions.rb`. For example:
-```
-Notifier.subscribe_mailer(event_name: Notifier::DEPOSITOR_ADDED, mailer_class: CollectionsMailer, mailer_method: :invitation_to_deposit_email)
-```
+    ```
+    Notifier.subscribe_mailer(event_name: Notifier::DEPOSITOR_ADDED, mailer_class: CollectionsMailer, mailer_method: :invitation_to_deposit_email)
+    ```
 1. Test the triggering of the notification.
-2. Test the email.
+1. Test the email.
```
mjgiarlo, line 151: Prefer automagic markdown list numbering.

app/components/structure/contact_form_component.html.erb
```diff
@@ -0,0 +1,30 @@
+<p>* Required fields</p>
+
+<%= form_with model: ContactForm.new do |form| %>
```
mjgiarlo, line 3: I didn't find the `@contact_form` instance variable in the controller to be necessary, so I dropped it.

app/components/structure/contact_form_component.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <%= form.hidden_field :bounce_location, value: helpers.root_path unless modal? %>
```
mjgiarlo, line 4: This component now allows the contact form to be rendered as a page or a modal.

app/components/structure/contact_form_success_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= tag.public_send(containing_tag_name, **containing_tag_args) do %>
+  <h2>Help request successfully sent</h2>
+  <p class="mt-3">Your message has been sent to the SDR team. We will respond to you soon.</p>
+  <%= render Elements::ButtonLinkComponent.new(**close_button_params) %>
+<% end %>
```
mjgiarlo, line 5: Componentized this so it can be used as a page or within a modal (turbo-frame).

app/policies/application_policy.rb
```diff
@@ -7,4 +7,10 @@ class ApplicationPolicy < ActionPolicy::Base
   def allow_admins
     allow! if Current.groups.include?(Settings.authorization_workgroup_names.administrators)
   end
+
+  private
+
+  def collection_creator?
+    Current.groups.include?(Settings.authorization_workgroup_names.collection_creators)
+  end
```
mjgiarlo, line 15: Moved from the collection policy to this one because it's used in multiple policies now.

app/policies/dashboard_policy.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class DashboardPolicy < ApplicationPolicy
+  def show?
+    collection_creator? || user.depositor_for.any? || user.reviewer_for.any? || user.manages.any?
+  end
+end
```
mjgiarlo, line 7: Don't need to account for administrators because that case is already handled by the application policy.

Gemfile
justinlittman: The less heavyweight option seems preferable.

app/controllers/contacts_controller.rb
```diff
@@ -4,16 +4,18 @@
...
-  end
+  def new; end
 
   def create
     contact_form = ContactForm.new(contact_form_params)
 
     ContactsMailer.with(contact_form.attributes.symbolize_keys).jira_email.deliver_later
 
-    redirect_to success_contacts_path
+    redirect_to success_contacts_path(bounce_location: params.dig('contact', 'bounce_location'))
```
justinlittman, line 14: Can you document what bounce location is doing?

app/controllers/contacts_controller.rb
```diff
@@ -4,16 +4,18 @@
...
+  end
+
+  def success
+    @bounce_location = params[:bounce_location]
```
justinlittman, line 18: What is this? I don't see where success is being called.

justinlittman (commented):
Here's my question: Contacts was already a page. Wouldn't it be less complex to include it in a turbo frame on new_user_contact_form.html?

app/controllers/contacts_controller.rb
mjgiarlo: It allows the `Close` link on the "success" view to point at the app root path instead of as a modal dismiss. (Not planning to document that yet as I consider whether to refactor as a turbo-frame.)

app/controllers/contacts_controller.rb
mjgiarlo: @justinlittman It's: https://github.com/sul-dlss/hungry-hungry-hippo/blob/25a7e5848d5e3509d87fe1783443980eab3ef057/app/controllers/contacts_controller.rb#L16 (route: https://github.com/sul-dlss/hungry-hungry-hippo/blob/25a7e5848d5e3509d87fe1783443980eab3ef057/config/routes.rb#L52)

justinlittman: Duh.

README.md
```diff
@@ -23,6 +23,33 @@ bin/dev
 
 Then browse to http://localhost:3000 to see the running application.
 
+#### Change default user role
+
+You can change your abilities within the app by setting the `ROLES` environment
+variable. By default you will be granted administrative abilities.
+
+Note that you may need to clear your browser session cookies to pick up the new
```
justinlittman, line 31: Can we say less here, like just the first sentence?

mjgiarlo: If you can believe it, this is already trimmed down from what's in H2, IIRC. I agree that it's still verbose. Will have another look.

Gemfile
mjgiarlo: I'll change it.

app/policies/application_policy.rb
```diff
@@ -7,4 +7,10 @@ class ApplicationPolicy < ActionPolicy::Base
   def allow_admins
     allow! if Current.groups.include?(Settings.authorization_workgroup_names.administrators)
   end
+
+  private
```
justinlittman, line 11: private or protected?

app/policies/dashboard_policy.rb
justinlittman: Would `.exist?` be faster than `.any?`?

app/controllers/contacts_controller.rb
```diff
@@ -5,20 +5,24 @@ class ContactsController < ApplicationController
...
 
   def create
     contact_form = ContactForm.new(contact_form_params)
 
     ContactsMailer.with(contact_form.attributes.symbolize_keys).jira_email.deliver_later
 
-    redirect_to success_contacts_path
+    @welcome = contact_form.welcome?
+
+    render :success, status: :created
```
justinlittman, line 20: Slight concern that this won't work with turbo. Did you double check that this existing contact modal still works?

justinlittman (approved):
A few minor questions.

app/policies/application_policy.rb
mjgiarlo: In my experience, `protected` is seldom (never?) a good idea in Ruby code.

app/controllers/contacts_controller.rb
mjgiarlo: Yep, already tested that both forms work as intended

app/policies/dashboard_policy.rb
mjgiarlo: @justinlittman `#exists?` and `#any?` both execute the same query, it turns out. I think `user.manages.any?` reads a bit more naturally than `user.manages.exists?`

PR #639, opened 2025-02-06
Set proper dashboard link in breadcrumbs

aaron-collier:
I noticed that the `Dashboard` link in the breadcrumbs goes to the home page which is a bit confusing and makes it really odd to have to go collection > home > enter just to get back to the dashboard.

This fixes that to `dashboard_path`


justinlittman (changes requested):
It is also incorrect in collection and work's `form.html.erb`. Can you get those as well?

aaron-collier:
@justinlittman good catch, I forgot the dashboard was on the forms. Done.

PR #647, opened 2025-02-07
Renamed autocomplete controller and keywords component.

justinlittman:


app/components/keywords/edit_component.html.erb
justinlittman, line 1: This is only used in works, so moved to `works/edit/`

app/javascript/controllers/autocomplete_keywords_controller.js
justinlittman, line 1: The naming didn't make clear what this controller was related to.

mjgiarlo (approved):

PR #679, opened 2025-02-10
Changes validation logic and label for contributor names.

### LINKED ISSUE #662: Update error message for author/contributor name entry
When a name entry field is left blank, what's being displayed is shown below.

For the first name field this should say, "You must provide a first name." 
For the last name field this should say, "You must provide a last name."

Image

Also, weirdly, when only one of these fields is blank, the error message is displaying twice under the one that is blank:

Image

Image

justinlittman:
closes #662

This PR:
* Changes the label for a missing first / last name.
* Simplifies the logic by getting rid of the special handling for ORCID names. (Now when there is an ORCID it is handled like other names.)
* Avoid duplicating validations.

### DIFF
--- a/spec/forms/contributor_form_spec.rb
+++ b/spec/forms/contributor_form_spec.rb
@@ -93,12 +93,26 @@ RSpec.describe ContributorForm do
       end
     end
 
+    context 'when depositing and missing part of a name' do
+      let(:first_name) { 'Jane' }
+      let(:last_name) { '' }
+
+      it 'is not valid' do
+        expect(form.valid?(:deposit)).to be false
+        expect(form.errors.size).to eq(1)
+        expect(form.errors[:last_name]).to eq(['must provide a last name'])
+      end
+    end
+
     context 'when depositing and missing name' do
       let(:first_name) { '' }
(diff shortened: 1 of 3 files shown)


spec/forms/contributor_form_spec.rb
```diff
@@ -94,11 +94,25 @@
...
+      let(:last_name) { '' }
+
+      it 'is not valid' do
+        expect(form.valid?(:deposit)).to be false
+        expect(form.errors.size).to eq(1)
+        expect(form.errors[:last_name]).to eq(['must provide a last name'])
+      end
+    end
+
+    context 'when depositing and missing part of a name' do
```
mjgiarlo, line 107: The context here looks like it's missing the entire name. Maybe the context strings on lines 107 and 96 should be swapped? Not blocking, just wondering.

mjgiarlo (approved):
Approve pending feedback on RSpec context strings

spec/forms/contributor_form_spec.rb
justinlittman: Good catch. Fixing.

PR #699, opened 2025-02-11
Changes error message for email address.

### LINKED ISSUE #660: Make error message for email address in work form clearer
Instead of "is invalid" this should say "You must provide a valid email address."

Image

justinlittman:
closes #660

### DIFF
--- a/app/forms/contact_email_form.rb
+++ b/app/forms/contact_email_form.rb
@@ -6,6 +6,7 @@ class ContactEmailForm < ApplicationForm
   validates :email, presence: true, on: :deposit
   validates :email, format: {
     with: URI::MailTo::EMAIL_REGEXP,
-    allow_blank: true
+    allow_blank: true,
+    message: 'must provide a valid email address' # rubocop:disable Rails/I18nLocaleTexts
   }
 end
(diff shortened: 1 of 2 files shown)


aaron-collier (approved):

app/forms/contact_email_form.rb
```diff
@@ -6,6 +6,7 @@ class ContactEmailForm < ApplicationForm
   validates :email, presence: true, on: :deposit
   validates :email, format: {
     with: URI::MailTo::EMAIL_REGEXP,
-    allow_blank: true
+    allow_blank: true,
+    message: 'must provide a valid email address' # rubocop:disable Rails/I18nLocaleTexts
```
aaron-collier, line 10: Does something like: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/config/locales/en.yml#L42 not work here? It seems if we're disabling a rubocop rule and have a possible approach, I would lean towards agreeing with rubocop. But that's a minor, non-blocking issue.

mjgiarlo: Concur with @aaron-collier

justinlittman: I'm going to adjust this and other validation messages in a follow-up PR.

PR #715, opened 2025-02-11
Make custom validation messages explicit instead of magic.

justinlittman:


config/locales/en.yml
```diff
@@ -29,21 +29,6 @@
 
 en:
   app_name: 'Stanford Digital Repository'
-  activemodel:
```
justinlittman, line 32: Prefer messages over explicit over Rails magic. The current state of these messages are evidence why Rails magic leads to cruft.

config/locales/en.yml
```diff
@@ -29,21 +29,6 @@
...
-    errors:
-      models:
-        related_work:
-          attributes:
-            relationship:
-              inclusion: '%{value} is not a valid relationship for related works'
```
justinlittman, line 38: No need for a custom message. The validation is only for documentation and could never be false.

config/locales/en.yml
```diff
@@ -80,9 +67,6 @@ en:
   depositors:
     edit:
       legend: 'Depositors'
-  publication_date:
```
justinlittman, line 83: Alphabetized.

config/locales/en.yml
```diff
@@ -29,21 +29,6 @@
...
-  activemodel:
-    errors:
-      models:
-        related_work:
-          attributes:
-            relationship:
-              inclusion: '%{value} is not a valid relationship for related works'
-        contact_email:
-          attributes:
-            email:
```
justinlittman, line 41: Not actually being used.

aaron-collier (approved):

PR #719, opened 2025-02-12
Adjusts work subtype validation.

### LINKED ISSUE #657: Validation error causes confusion when trying to correct.
I think the fact that the error message continues to display causes confusion on the Type of work page. 

Steps to reproduce:
1. Create item and choose "Music" as the work type but do not select a subtype.
2. Attempt to deposit to trigger validation error.
3. Return to "Type of work" page
4. Choose any of the other work types except "other." 
5. Error message indicates that you have to choose a subtype, however, no subtypes are required for the following:

- text
- data
- software/code
- image
- sound
- video

Image

justinlittman:
closes #657


app/components/works/edit/work_subtypes_component.html.erb
```diff
@@ -3,6 +3,8 @@
     <%= content %>
   <% else %>
     <%= render Elements::Forms::FieldsetComponent.new(label: t('works.edit.fields.work_subtypes.label')) do %>
+      <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name: error_field_name) if error_field_name %>
```
justinlittman, line 6: This allows providing an error that is specific to a work subtype instead of using the same error message for all work subtypes.

app/components/works/edit/work_type_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <div data-controller="work-type">
-  <div class="pane-section pb-0">
+  <div class="pane-section">
```
justinlittman, line 2: Formatting snafu.

app/components/works/edit/work_subtypes_component.html.erb
mjgiarlo: Would it have unintended consequences if we removed the `if error_field_name` here and changed [`InvalidFeedbackComponent#render?`](https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/components/elements/forms/invalid_feedback_component.rb#L23) to:

```ruby
      def render?
        errors.present? && field_name.present?
      end
```

(I assume it *would* have bad side effects, but it's worth asking if we can use VC's built-in conditional rendering semantics here.)

mjgiarlo (approved):
Approve pending consideration of VC rendering question

app/components/works/edit/work_subtypes_component.html.erb
justinlittman: That'll work. Fixing.

PR #720, opened 2025-02-12
Allow user to resend contact.

### LINKED ISSUE #636: Allow user to submit the contact form multiple times without reloading the page.
I'm having trouble being able to get the Contact form for the new home page on qa (https://sul-h3-qa.stanford.edu/) to appear again after I have submitted it.

Steps to replicate:
1. Go to above link
2. click on Contact us in the header.
3. fill in form and submit
4. once submitted I see this.

Image

5. I then clicked the x at the top right.
6. If I then click on the Contact us link, I see the same thing as above, with no way to submit another form.

I would expect that once I've closed this message, if I clicked on the Contact us link again I would again see the form.

justinlittman:
closes #636

image

### DIFF
--- a/app/views/contacts/success.html.erb
+++ b/app/views/contacts/success.html.erb
@@ -1,9 +1,11 @@
 <turbo-frame id="contact-form">
-  <h2>Help request successfully sent</h2>
-  <p class="mt-3">Your message has been sent to the SDR team. We will respond to you soon.</p>
+  <%= render Elements::AlertComponent.new(variant: :success, dismissible: !@welcome) do %>
+    <div class="h2">Help request successfully sent</div>
+    <p class="mt-3">Your message has been sent to the SDR team. We will respond to you soon.</p>
+  <% end %>
   <% if @welcome %>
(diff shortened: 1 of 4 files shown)


app/views/contacts/success.html.erb
```diff
@@ -1,9 +1,11 @@
 <turbo-frame id="contact-form">
-  <h2>Help request successfully sent</h2>
-  <p class="mt-3">Your message has been sent to the SDR team. We will respond to you soon.</p>
+  <%= render Elements::AlertComponent.new(variant: :success, dismissible: !@welcome) do %>
+    <div class="h2">Help request successfully sent</div>
```
mjgiarlo, line 3: Can you say more about changing from an `h2` to a `div`?

mjgiarlo (approved):
It looks like this removes the `Cancel` link from the welcome form which was part of the design. Do we need that element on the page still or can we ditch it? Question for Astrid and Amy, I assume.

Approve pending consideration of ☝🏻 and the HTML question

app/views/contacts/success.html.erb
justinlittman: This is an H2 display, not semantically a heading.

justinlittman:
There's a Cancel link at the bottom of the form, so I think this is OK.

PR #721, opened 2025-02-12
Changes roundtripping error message.

### LINKED ISSUE #619: Display yellow alert if user is not able to edit deposit/collection
For those users who are not able to edit their deposit/collection because we cannot roundtrip their data, display yellow alert (in component library) that user can dismiss. The alert will appear when user clicks 'edit' and then disappear as soon as the session is over, or if user closes the toast. 

Text: "If you need to make a change to this work, contact sdr-contact@lists.stanford.edu"

Image

justinlittman:
closes #619

image

### DIFF
--- a/app/components/structure/contact_form_modal_component.html.erb
+++ b/app/components/structure/contact_form_modal_component.html.erb
@@ -3,7 +3,7 @@
     <h1><%= contact_sdr %></h1>
   <% end %>
   <% component.with_body do %>
-    <%= tag.turbo_frame id: 'contact-form', src: new_contacts_path do %>
+    <%= tag.turbo_frame id: 'contact-form', src: new_contacts_path, loading: 'lazy' do %>
       <%= render Elements::SpinnerComponent.new %>
     <% end %>
   <% end %>
--- a/app/components/structure/terms_of_deposit_modal_component.html.erb
+++ b/app/components/structure/terms_of_deposit_modal_component.html.erb
@@ -12,7 +12,7 @@
     </div>
   <% end %>
   <% component.with_body do %>
-    <%= tag.turbo_frame id: 'terms-of-deposit', src: terms_path do %>
+    <%= tag.turbo_frame id: 'terms-of-deposit', src: terms_path, loading: 'lazy' do %>
       <%= render Elements::SpinnerComponent.new %>
     <% end %>
   <% end %>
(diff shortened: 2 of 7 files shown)


app/components/structure/contact_form_modal_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <h1><%= contact_sdr %></h1>
   <% end %>
   <% component.with_body do %>
-    <%= tag.turbo_frame id: 'contact-form', src: new_contacts_path do %>
+    <%= tag.turbo_frame id: 'contact-form', src: new_contacts_path, loading: 'lazy' do %>
```
justinlittman, line 6: This should be lazy so only loaded when needed. This change was made to avoid flash getting rendered here.

app/components/structure/terms_of_deposit_modal_component.html.erb
```diff
@@ -12,7 +12,7 @@
     </div>
   <% end %>
   <% component.with_body do %>
-    <%= tag.turbo_frame id: 'terms-of-deposit', src: terms_path do %>
+    <%= tag.turbo_frame id: 'terms-of-deposit', src: terms_path, loading: 'lazy' do %>
```
justinlittman, line 15: Same

app/controllers/works_controller.rb
```diff
@@ -35,8 +35,8 @@ def edit
     authorize! @work
 
     unless editable?
-      flash[:danger] = helpers.t('works.edit.messages.cannot_be_edited')
-      return redirect_to work_path(druid: params[:druid])
+      flash[:warning] = helpers.t('works.edit.messages.cannot_be_edited_html', support_email: Settings.support_email)
+      return redirect_to work_path(druid: params[:druid]), status: :see_other
```
justinlittman, line 39: Turbo requires :see_other

spec/system/show_terms_of_deposit_spec.rb
```diff
@@ -12,6 +12,9 @@
 
     click_link_or_button 'Terms of Deposit'
 
+    # The lazy loading of the modal fails sporadically without this sleep.
+    sleep 0.25
```
mjgiarlo, line 16: Could we replace this with a `find` call that, e.g., looks for unique text in the modal? Then we can rely on Capybara's built-in waiting ability instead of a hard-coded sleep value.

mjgiarlo (approved):
Approve pending test feedback

spec/system/show_terms_of_deposit_spec.rb
justinlittman: You'd think it would. I tried multiple approaches (looking for text, looking for the completed attribute on the turbo frame) but none would consistently work.

PR #725, opened 2025-02-13
Release duration select prompt should always be available in collection access form

lwrubel:
Previously, if a saved collection with a release duration was then edited, the prompt "Select an option" would not be present in the dropdown if the user then decided to change the selection. This is because the values loaded into the dropdown in were only the valid options. This tweak adds it back in that situation and also handles the invalid state. 

Screenshot 2025-02-13 at 9 34 33 AM


justinlittman (approved):
Thanks for the fix @lwrubel !

PR #737, opened 2025-02-13
Adds home and dashboard banner.

### LINKED ISSUE #646: Allow admin to edit banner text
Admins should be able to edit the banner text that displays on the home page and at the top of the dashboard. We need to be able to include links in this text.

The text should be in a yml file that is easily accessible by the admins for making changes.

justinlittman:
closes #646

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,6 +28,12 @@
 #       enabled: "ON"
 
 en:
+  banner:
+    # Home banner must always be present.
+    # To hide dashboard banner, set to ''.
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -28,6 +28,12 @@
 #       enabled: "ON"
 
 en:
+  banner:
```
justinlittman, line 31: Amy will modify these values. There are separate entries for home and dashboard because home banner must always be present, but dashboard banner doesn't.

aaron-collier (approved):

Issue #670, opened 2025-02-07
Add collections list to contact form when user chooses "Request access to another collection"


amyehodge:

amyehodge:
Info for the tool tips for these are below. Notice these have been updated from the content used in H2.

        label: Faculty, Student, and Staff Publications
        description: >
          An open collection for assembling publications by Stanford faculty, graduate students,
          post-docs, and staff from the full range of disciplines.


        label: Stanford Research Data
        description: >
          This collection includes research data, software, supporting code, protocols and other materials related to research from Stanford-associated researchers and scientists on the wide variety of topics and fields under investigation at Stanford University, including biology, chemistry, engineering, environmental sciences, geosciences, history, languages, mathematics, medicine, music, physics, public policy, social sciences, and statistics. Data sets in this collection are in many different formats, and each is presented with descriptive information that identifies the people and organizations responsible for its creation, related publications, and other relevant information. The collection was created to support the preservation of the research output of the University and to enable data sharing and re-use.


        label: Stanford Theses and Dissertations
        description: >
          This collection includes theses and dissertations authored by Stanford graduate and undergraduate
          students from a variety of schools and representing a broad range of scholarly disciplines. Established
          in 2014, the collection supports the preservation, sharing, and re-use of the research output of the
          University, particularly for those programs for which there is no formal or mandatory requirement for
          thesis deposit.


        label: Stanford University Libraries staff publications and research
        description: >
          Presentations, publications and research produced and contributed by the staff of Stanford Libraries on a broad range of topics relevant to academic and research libraries.

    
        label: Stanford University Open Access Articles
        description: >
          A collection of open access articles published by faculty, students, and other researchers at Stanford
          University.


  
```

PR #738, opened 2025-02-13
Implement tooltips

### LINKED ISSUE #450: Add tooltips
Consider providing these in yml file for easier editing by PO.

List of text for all the tooltips is at https://docs.google.com/document/d/12RkTahpFaZ5VGJY-idrhY8184Hw0TI4QmQ5-AzjZk0U/edit?tab=t.0. This includes new ones for the "What's changing" box on both forms and the citation. A few of the ones we had in H2 are no longer needed.

mjgiarlo:
Fixes #450

Screenshot from 2025-02-13 13-27-54

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -264,7 +264,7 @@ body {
 
 .opacity-85 {
   opacity: 0.85;
-}  
+}
 
 i.icon-lg {
   font-size: 2rem;
@@ -272,6 +272,14 @@ i.icon-lg {
 
 i.icon-xl {
   font-size: 3rem;
+}
+
+.tooltip-info {
+  color: var(--stanford-digital-blue);
+}
+
+.tooltip-header {
+  display: inline-block;
 }
 
 .secondary-card {
--- a/app/components/elements/forms/label_component.html.erb
+++ b/app/components/elements/forms/label_component.html.erb
@@ -1,1 +1,2 @@
-  <%= form.label field_name, label_text, class: classes %>
+<%= form.label field_name, label_text, class: classes %>
+<%= render Elements::TooltipComponent.new(tooltip:) %>
--- a/app/components/elements/forms/label_component.rb
+++ b/app/components/elements/forms/label_component.rb
@@ -5,17 +5,18 @@ module Elements
     # Component for rendering a form label.
     class LabelComponent < ApplicationComponent
       def initialize(form:, field_name:, label_text: nil, default_label_class: 'form-label', hidden_label: false, # rubocop:disable Metrics/ParameterLists
-                     classes: [])
+                     classes: [], tooltip: nil)
         @form = form
         @label_text = label_text
         @field_name = field_name
         @hidden_label = hidden_label
         @default_label_class = default_label_class
         @classes = classes
+        @tooltip = tooltip
         super()
       end
 
-      attr_reader :field_name, :form
+      attr_reader :field_name, :form, :tooltip
 
       def label_text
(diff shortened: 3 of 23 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -274,6 +274,14 @@ i.icon-xl {
   font-size: 3rem;
 }
 
+.tooltip-info {
+  color: var(--stanford-digital-blue);
+}
+
+.tooltip-shift-right {
+  left: 15px !important;
```
mjgiarlo, line 282: I tweaked a great many things to get the Bootstrap tooltip to render directly atop the icon element, but it kept wanting to be 15px to the left. This shifts it back where we want it, I think. I'm open to alternatives, of course.

app/components/elements/forms/label_component.html.erb
```diff
@@ -1 +1,4 @@
-  <%= form.label field_name, label_text, class: classes %>
+<%= form.label field_name, label_text, class: classes %>
+<% if render_tooltip? %>
+  <%= helpers.info_icon(fill: true, classes: 'px-3 tooltip-info', data: { bs_custom_class: 'tooltip-shift-right', bs_toggle: 'tooltip', bs_title: tooltip, bs_trigger: 'click focus', tooltips_target: 'icon' }) %>
```
justinlittman, line 3: Componentize so that can be used elsewhere?

app/components/elements/forms/label_component.rb
```diff
@@ -5,18 +5,28 @@ module Forms
...
         @hidden_label = hidden_label
         @default_label_class = default_label_class
         @classes = classes
+        @tooltip = tooltip
         super()
       end
 
       attr_reader :field_name, :form
 
+      def render_tooltip?
```
justinlittman, line 21: Opinion: I'd go with `tooltip?` instead of `render_tooltip?`.

app/javascript/controllers/tooltips_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from '@hotwired/stimulus'
+import * as bootstrap from 'bootstrap' // eslint-disable-line no-unused-vars
```
justinlittman, line 2: Are you sure this is necessary? (It might be, but might already be imported elsewhere.)

app/components/elements/forms/label_component.rb
mjgiarlo: @justinlittman Sounds good to me. 👍🏻

app/javascript/controllers/tooltips_controller.js
mjgiarlo: Sadly, it is necessary here.

app/components/elements/forms/label_component.html.erb
justinlittman: I was under the assumption that we had tooltips all over the place, not just in form labels. If that is mistaken, then agree that no need to componentize.

amyehodge: Other than the collection and work forms, there are also tool tips in the help form. See #670

app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -274,6 +274,19 @@ i.icon-xl {
   font-size: 3rem;
 }
 
+.tooltip-info {
+  color: var(--stanford-digital-blue);
+}
+
+.tooltip-shift-right {
```
justinlittman, line 281: I'd just do this as an inline style.

justinlittman (approved):

PR #764, opened 2025-02-19
Re-implement move work to new collection.

### LINKED ISSUE #458: Admin feature: ability to move item to another collection
This feature displays for admins on item pages.

justinlittman:
closes #458

https://github.com/user-attachments/assets/055240ac-dfb0-4294-afad-97d2ea0fcf14


app/controllers/admin/move_controller.rb
```diff
@@ -3,44 +3,57 @@
...
+        flash[:success] = I18n.t('messages.work_moved')
+        # This breaks out of the turbo frame.
+        respond_to do |format|
+          format.turbo_stream do
+            render turbo_stream: turbo_stream.action(:full_page_redirect, wait_works_path(work.id))
+          end
+        end
+      else
+        render :form, status: :unprocessable_entity
+      end
```
mjgiarlo, line 31: Minor tweak: how would you feel about shifting the `else` logic into a guard clause? Why? It removes one level of indentation from the `if` logic. I generally find code that is indented multiple levels to be harder to grok, because you have to hold in your head what each level of indentation means. Shooting for more "vertical" code than "horizontal."

```suggestion
      return render :form, status: :unprocessable_entity unless @move_form.valid?

      work_form.content_id = @move_form.content_id
      Admin::Move.call(work_form:, work:, collection: @move_form.collection, version_status:)

      flash[:success] = I18n.t('messages.work_moved')
      # This breaks out of the turbo frame.
      respond_to do |format|
        format.turbo_stream do
          render turbo_stream: turbo_stream.action(:full_page_redirect, wait_works_path(work.id))
        end
      end
```

app/forms/admin/move_form.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      self.collection_druid = collection_druid.prepend('druid:')
+    end
+
+    def validate_collection_is_targetable
+      return if validate_collection_blank
+
+      return if validate_collection_found
+
+      return if validate_already_part_of_collection
```
mjgiarlo, line 28: It looks like these methods conditionally add an error but all return a boolean. I wondered about renaming these to be question-mark methods since we use them to determine validation flow. Would it be surprising for a question-mark method to also add an error, though? (Maybe not since we're in a form validation context?)

mjgiarlo, line 28: It looks like these methods conditionally add an error but all return a boolean. I wondered about renaming these to be question-mark methods since we use them to determine validation flow. Would it be surprising for a question-mark method to also add an error, though? (Maybe not since we're in a form validation context?)

E.g.:

```suggestion
      return if collection_druid_missing? || collection_not_found? already_part_of_collection?
```

app/javascript/application.js
```diff
@@ -1,3 +1,8 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import '@hotwired/turbo-rails'
 import 'controllers'
+
+// This is to help with breaking out of a turbo frame.
+Turbo.StreamActions.full_page_redirect = function () { // eslint-disable-line no-undef
```
mjgiarlo, line 6: How do you feel about newer JS idioms like the fat arrow function syntax?

```suggestion
Turbo.StreamActions.full_page_redirect = () => { // eslint-disable-line no-undef
```

app/controllers/admin/move_controller.rb
```diff
@@ -3,44 +3,57 @@
...
+        work_form.content_id = @move_form.content_id
+        Admin::Move.call(work_form:, work:, collection: @move_form.collection, version_status:)
+
+        flash[:success] = I18n.t('messages.work_moved')
+        # This breaks out of the turbo frame.
+        respond_to do |format|
+          format.turbo_stream do
+            render turbo_stream: turbo_stream.action(:full_page_redirect, wait_works_path(work.id))
+          end
+        end
```
mjgiarlo, line 28: Would `redirect_to wait_works_path(work)` not break out of the frame here?

app/controllers/admin/move_controller.rb
```diff
@@ -3,44 +3,57 @@
...
+
+      if @move_form.valid?
+        work_form.content_id = @move_form.content_id
+        Admin::Move.call(work_form:, work:, collection: @move_form.collection, version_status:)
+
+        flash[:success] = I18n.t('messages.work_moved')
+        # This breaks out of the turbo frame.
+        respond_to do |format|
+          format.turbo_stream do
+            render turbo_stream: turbo_stream.action(:full_page_redirect, wait_works_path(work.id))
```
mjgiarlo, line 26: Do we need `work.id` or would the `work` suffice?

app/controllers/admin/move_controller.rb
justinlittman: No, it won't. This is definitely one of the turbo rough edges.

app/forms/admin/move_form.rb
mjgiarlo: @justinlittman I buy that. I might still like the `validate` method names to be tweaked. Right now, it's not clear whether `validate_collection_blank` and `validate_collection_found` are consider "blank" and "found" to be the valid conditions or the invalid ones. Especially because the former marks "blank" invalid and "found" valid. The question mark method names above make the intent clearer and read more like English to me, at least. We could rename the methods to those names but remove the question marks?

app/javascript/application.js
justinlittman: Javascript is a mystery ...
image

app/controllers/admin/move_controller.rb
mjgiarlo: All the other `if ...valid?` blocks in the controllers are like 2-3 lines. This `if` block is a bit longer and gets up to four levels of indentation relative to the method. 

Patternwise, I'm not too worried about us not understanding what `unless ...valid?` means. My sense is that we use one or the other depending on context. In general, I'd say we prefer guard clauses to indentation where we can justify it. I think we can make a case that it makes sense here.

app/javascript/application.js
mjgiarlo: Sigh. Yeah, that makes no sense to me either.

app/controllers/admin/move_controller.rb
```diff
@@ -3,44 +3,61 @@
...
-      Admin::Move.call(work_form: @work_form, work: @work, collection:, version_status: @version_status)
+      @move_form = Admin::MoveForm.new(work_form:, **move_params)
 
-      flash[:success] = I18n.t('messages.work_moved')
-      redirect_to wait_works_path(@work.id)
+      if @move_form.valid?
+        work_form.content_id = @move_form.content_id
+        Admin::Move.call(work_form:, work:, collection: @move_form.collection, version_status:)
+
+        render_move_success
```
mjgiarlo, line 22: Rubocop-related extraction?

app/forms/admin/move_form.rb
mjgiarlo: @justinlittman I like this even better, thanks!

app/controllers/admin/move_controller.rb
justinlittman: It allowed preserving the `if ... then` in the controller method by reducing the cognitive load.

PR #776, opened 2025-02-20
Handle hidden files when mapping object type.

### LINKED ISSUE #580: Enable support to display files for works that only have pdfs in the document viewer
If a work has one or more pdf files attached to it and no other types of files, display these files automatically in the document viewer.

It would also be great if we could exclude hidden files. So, for example, if someone deposits a pdf that is not hidden and a csv file that is hidden, we would still display this in the document viewer.

If an item is later edited to include not hidden files that are not pdfs, or if hidden files that are not pdfs are unhidden, then we would want to revert back to the file viewer.

justinlittman:
closes #580

### DIFF
--- a/app/services/to_cocina/work/mapper.rb
+++ b/app/services/to_cocina/work/mapper.rb
@@ -57,9 +57,13 @@ module ToCocina
       def document?
         return false if content.content_files.empty?
 
-        content.content_files.all? do |content_file|
-          !content_file.hidden? && content_file.mime_type == 'application/pdf'
+        # All files are PDF or hidden
+        return false unless content.content_files.all? do |content_file|
+          content_file.hidden? || content_file.pdf?
         end
+
+        # There is at least one non-hidden PDF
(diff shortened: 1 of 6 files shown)


aaron-collier (approved):

app/services/to_cocina/work/mapper.rb
```diff
@@ -57,9 +57,13 @@ def object_type
...
 
-        content.content_files.all? do |content_file|
-          !content_file.hidden? && content_file.mime_type == 'application/pdf'
+        # All files are PDF or hidden
+        return false unless content.content_files.all? do |content_file|
+          content_file.hidden? || content_file.pdf?
         end
+
+        # There is at least one non-hidden PDF
+        content.content_files.any? { |content_file| !content_file.hidden? && content_file.pdf? }
```
mjgiarlo, line 66: @justinlittman Any reason to be wary of this simplification? https://github.com/sul-dlss/hungry-hungry-hippo/pull/783/files

PR #783, opened 2025-02-20
Simplify logic determining whether to use document viewer in Cocina mapper

mjgiarlo:


app/services/to_cocina/work/mapper.rb
```diff
@@ -55,13 +55,6 @@ def object_type
       end
 
       def document?
-        return false if content.content_files.empty?
-
-        # All files are PDF or hidden
-        return false unless content.content_files.all? do |content_file|
-          content_file.hidden? || content_file.pdf?
-        end
```
mjgiarlo, line 63: @justinlittman Was there a reason we needed these checks too?

justinlittman: It shouldn't be the document type if there is file that is not hidden and not a PDF.

mjgiarlo: @justinlittman Doesn't line 59 cover that? It returns `true` if there are any content files that are both not hidden and PDF. A not hidden not PDF would cause it to return `false`.

justinlittman: Suppose 2 files:
1. Non-hidden PDF
2. Non-hidden JPEG

This should NOT be a document because of 2. However, L59 would be true because of 1.

justinlittman: This seems to indicate that there is a missing test: https://github.com/sul-dlss/hungry-hungry-hippo/pull/789

mjgiarlo: Good call, thanks for adding a test

mjgiarlo: @justinlittman Take another look?

app/services/to_cocina/work/mapper.rb
```diff
@@ -55,15 +55,12 @@ def object_type
...
+        # All shown (not hidden) files must be PDFs to be considered a document type
+        shown_files.any? && shown_files.all?(&:pdf?)
+      end
 
-        # There is at least one non-hidden PDF
-        content.content_files.any? { |content_file| !content_file.hidden? && content_file.pdf? }
+      def shown_files
+        @shown_files ||= content.content_files.reject(&:hidden?)
```
justinlittman, line 63: Scope on content?

justinlittman (approved):

app/services/to_cocina/work/mapper.rb
mjgiarlo: I like that idea. I changed this to a scope on ContentFile and a helper method on Content. Have a look-see.

PR #786, opened 2025-02-21
[HOLD] Reset work and collection forms.

### LINKED ISSUE #775: What's changing text should be cleared after deposit
I attempted to edit this item https://sul-h3-stage.stanford.edu/works/druid:ff595qh2652/edit and when I get to the deposit tab, the what's changing box already had text in it. Presumably this is what was entered the last time, but this is not relevant for this deposit. Can we clear this out when the user starts a new draft?

Image

justinlittman:
closes #775

### DIFF
--- a/app/components/edit/tab_form/tab_list_component.html.erb
+++ b/app/components/edit/tab_form/tab_list_component.html.erb
@@ -12,7 +12,7 @@
         <%= pane %>
       <% end %>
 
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):

justinlittman:
Turned out this wasn't the solution to the reported problem.

PR #787, opened 2025-02-21
[SPIKE] Convert FilenameSupport service to a FilePath value object

mjgiarlo:
This spike PR reconceives one of our "support" classes as a value object. "Support" classes can be a bit of a smell, as can a service class that is merely a bucket of methods. This seemed like an opportunity to consider a different design.

Depending on how we feel about this, we have a small number of remaining "support" classes we could consider as additional value objects, in addition to a handful of "models" that might also be more value-y than model-y. (While acknowledging that the boundary there is not so well defined as we might like.)


app/services/contents/builder.rb
```diff
@@ -30,13 +30,13 @@ def call
     def content_file_params_for(content:) # rubocop:disable Metrics/AbcSize
       cocina_object.structural.contains.flat_map do |file_set|
         file_set.structural.contains.map do |file|
-          filepath = file.filename
+          file_path = FilePath.new(filepath: file.filename)
```
justinlittman, line 33: I find having `file_path` and `filepath` confusing. Is there some way to avoid this?

mjgiarlo: I took another hack at it. Better? Worse? Different?

mjgiarlo:
@justinlittman 💬 
> I still find this confusing. `file_path` and `filepath` aren't distinct enough. What about calling the value object `FilepathValue`?

The `filepath` that's confusing, do you mean the DB field defined on the content_files table?

justinlittman:
When I look at the code, trying to understand when filepath is a string value and when it is a value object.

mjgiarlo:
@justinlittman 💬 
> When I look at the code, trying to understand when filepath is a string value and when it is a value object.

If it's worth putting another 30-40 minutes into, I would rename the `filepath` column in the `content_files` table to something like `original_path` or `original_filename` and then reflect that change throughout the code to isolate that field such that any usage of `file...path` in the code is for `FilePath` value objects. If that doesn't sound fruitful, I'd opt to set this spike aside for now.

justinlittman:
I try to be consistent with:
* filepath = path + filename
* path = directory only
* filename = filename only

Using "filepath" or "file_path" to mean something different in just this codebase is likely to be confusing. (The reason Pathname barely works in the Ruby library is because it is a marginally unique name that isn't often used for anything else.)

I'd be perfectly happy with calling the value object something that is none of the above but don't like overloading one of these existing terms.

PR #788, opened 2025-02-21
Display filepath on paginated display.

### LINKED ISSUE #780: Paginated file display is not displaying file path
I uploaded over 2000 files to one item in multiple folders. The file list on the view page is showing the full paths, but in the form itself, these are missing.

Image

justinlittman:
closes #780


app/views/contents/_show.html.erb
```diff
@@ -8,7 +8,7 @@
       <% content_files.each do |content_file| %>
         <% component.with_row do |row_component| %>
           <% row_component.with_cell do %>
-            <%= content_file.filename %>
+            <%= content_file.filepath %>
```
justinlittman, line 11: This is the change.

aaron-collier (approved):

PR #824, opened 2025-02-21
Update year validation message and style invalid input text to not wrap

### LINKED ISSUE #661: Error validation icon obscures date entry
On the dates page when a year does not pass validation, an exclamation icon shows up in the box. This obscures part of the data entry field and does not go away when the user deletes the previous entry and starts typing again, making it hard to see if you are typing the entry correctly. Also, the error message for both date entries on this page should say "Must be in the past."

@astridu what should this look like?

Image

aaron-collier:
Fixes #661 

Message for dates that are too early:
Screenshot 2025-02-21 at 1 44 39 PM

Message for dates that are in the future:
Screenshot 2025-02-21 at 1 44 52 PM

### DIFF
--- a/app/components/elements/forms/invalid_feedback_component.rb
+++ b/app/components/elements/forms/invalid_feedback_component.rb
@@ -31,7 +31,7 @@ module Elements
 
       def classes
         # Adding is-invalid to trigger the tab error.
-        merge_classes(%w[invalid-feedback is-invalid d-block text-nowrap], @classes)
+        merge_classes(%w[invalid-feedback is-invalid d-block], @classes)
       end
     end
   end
--- a/app/components/elements/forms/invalid_feedback_component.rb
+++ b/app/components/elements/forms/invalid_feedback_component.rb
@@ -31,7 +31,7 @@ module Elements
 
       def classes
         # Adding is-invalid to trigger the tab error.
-        merge_classes(%w[invalid-feedback is-invalid d-block], @classes)
+        merge_classes(%w[invalid-feedback is-invalid d-block text-nowrap], @classes)
       end
     end
   end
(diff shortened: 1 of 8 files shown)


app/forms/date_form.rb
```diff
@@ -3,8 +3,10 @@
 # Form object for handling date input
 class DateForm < ApplicationForm
   attribute :year, :integer
-  validates :year,
-            numericality: { greater_than_or_equal_to: 1000, less_than_or_equal_to: Time.zone.now.year, allow_nil: true }
+  validates :year, numericality: { greater_than_or_equal_to: 1000, allow_nil: true }
+  validates :year, numericality: { less_than_or_equal_to: Time.zone.now.year,
+                                   allow_nil: true,
+                                   message: 'Must be in the past' } # rubocop:disable Rails/I18nLocaleTexts
```
justinlittman, line 9: For consistency, "must" should be lowercase. Also for consistency, it should be in en.yml.

app/components/elements/forms/invalid_feedback_component.rb
```diff
@@ -31,7 +31,7 @@ def errors
 
       def classes
         # Adding is-invalid to trigger the tab error.
-        merge_classes(%w[invalid-feedback is-invalid d-block], @classes)
+        merge_classes(%w[invalid-feedback is-invalid d-block text-nowrap], @classes)
```
justinlittman, line 34: Should this be added to every invalid feedback, it just the one in question?

app/components/elements/forms/text_field_component.rb
```diff
@@ -6,8 +6,11 @@ module Forms
     class TextFieldComponent < FieldComponent
       def initialize(**args)
         args[:container_classes] = merge_classes('field-container', args[:container_classes])
+        @error_classes = args.delete(:error_classes) || []
```
aaron-collier, line 9: This avoided having to add another argument to `FieldComponent`.

justinlittman: I'd suggest that there is no reason to avoid this.

justinlittman (approved):

PR #825, opened 2025-02-21
Contact email cannot be blank

### LINKED ISSUE #799: Collection form should require contact email to save
Validation in the collection form is not picking up when the contact email section is blank. 

It looks like it's getting these things correct:

It won't allow multiple addresses in the same box:

Image

And it's finding when it's just plain text: 

Image

aaron-collier:
Fixes #799 

Screenshot 2025-02-21 at 2 01 10 PM

### DIFF
--- a/app/forms/contact_email_form.rb
+++ b/app/forms/contact_email_form.rb
@@ -6,7 +6,7 @@ class ContactEmailForm < ApplicationForm
   validates :email, presence: true, on: :deposit
   validates :email, format: {
     with: URI::MailTo::EMAIL_REGEXP,
-    allow_blank: false,
+    allow_blank: true,
     message: I18n.t('contact_email.validation.email.invalid')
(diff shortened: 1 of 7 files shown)


app/forms/contact_email_form.rb
```diff
@@ -6,7 +6,7 @@ class ContactEmailForm < ApplicationForm
   validates :email, presence: true, on: :deposit
   validates :email, format: {
     with: URI::MailTo::EMAIL_REGEXP,
-    allow_blank: true,
+    allow_blank: false,
```
justinlittman, line 9: This is going to require an email for a draft. Email should be optional for a draft.

justinlittman (changes requested):

justinlittman (approved):
Oh sweet, this ended up being an easy change.

PR #840, opened 2025-02-24
Remove whitespace from email before validation to avoid confusion

### LINKED ISSUE #833: On collection form, trailing space on email is triggering confusing validation
In this case I entered a valid email address with a space after it.

If we don't want the trailing space, can we just strip it? 

It's confusing to have a single valid email address and get a warning that you haven't provided a valid email address.

Image

aaron-collier:
Fixes #833

### DIFF
--- a/app/forms/contact_email_form.rb
+++ b/app/forms/contact_email_form.rb
@@ -9,4 +9,12 @@ class ContactEmailForm < ApplicationForm
     allow_blank: true,
     message: I18n.t('contact_email.validation.email.invalid')
   }
+
+  before_validation :strip_whitespace
+
+  private
+
+  def strip_whitespace
+    email&.strip!
+  end
 end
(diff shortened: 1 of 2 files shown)


app/forms/contact_email_form.rb
```diff
@@ -9,4 +9,12 @@ class ContactEmailForm < ApplicationForm
     allow_blank: true,
     message: I18n.t('contact_email.validation.email.invalid')
   }
+
+  before_validation :strip_whitespace
+
+  private
+
+  def strip_whitespace
+    self.email.try(&:strip!)
```
justinlittman, line 18: `self.email&.strip!` is probably more idiomatic.

justinlittman (approved):
Approve pending consideration of suggestion.

app/forms/contact_email_form.rb
aaron-collier: Good call, fixed.

PR #842, opened 2025-02-25
Change mechanism for refreshing files for file upload.

### LINKED ISSUE #796: Folders should display after uploading
When one or more folders are uploaded at the same time along with a single file at the top level of the hierarchy, the file displays when it is uploaded but the folder and its contents do not. If I delete the single file, the rest of the contents then displays.

Everything should be displaying when it is uploaded.

This is reproducible in multiple collections. I saw this behavior when I was uploading one file with one, two, or four folders. Also, if I have two files at the top level and one or two folders, everything works as it should.

https://github.com/user-attachments/assets/9026204b-bbee-43b9-b0da-89616001e44c

To reproduce:
1. Click "Deposit to this collection" for any collection.
2. On the file upload page, drag over one file and one or more folders at the same time.
3. When upload is finished, only the file will be displayed.
4. Click the trash can to remove the file. You'll then see the rest of the contents.

justinlittman:
closes #796

### DIFF
--- a/app/controllers/contents_controller.rb
+++ b/app/controllers/contents_controller.rb
@@ -20,12 +20,6 @@ class ContentsController < ApplicationController
 
     update_files
 
-    # Broadcast the update to the content if completed
-    if params[:content][:completed] == 'true'
-      Turbo::StreamsChannel.broadcast_update_later_to @content, target: 'content-files', partial: 'contents/show',
-                                                                locals: { content: @content }
-    end
-
     head :ok
   end
 
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -7,8 +7,9 @@ export default class extends Controller {
     existingFiles: { type: Number, default: 0 },
     maxFiles: Number,
     maxFilesize: Number
-
   }
+
+  static outlets = ['dropzone-files']
 
   // Expected error handling behavior:
   // - A blank directory is ignored.
@@ -31,6 +32,7 @@ export default class extends Controller {
     this.shouldClearErrors = false
     this.basePath = null
     this.dropzone.on('processingmultiple', () => {
+      this.element.dataset.dropzoneReady = false
       // Using processingmultiple instead of addedfiles for showing the progress bar
       // since addedfiles is triggered by dropping an empty directory.
       // This is not.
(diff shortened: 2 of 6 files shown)


app/controllers/contents_controller.rb
```diff
@@ -20,12 +20,6 @@ def update
 
     update_files
 
-    # Broadcast the update to the content if completed
```
justinlittman, line 23: No longer broadcasting the change.

app/javascript/controllers/dropzone_controller.js
```diff
@@ -61,7 +64,6 @@ export default class extends Controller {
         if (this.basePath) path = `${this.basePath}/${path}`
         data.append(`content[paths][${i}]`, path)
       }
-      data.append('content[completed]', this.dropzone.getActiveFiles().length === files.length)
```
justinlittman, line 64: This mechanism was unreliable because the upload of files was asynchronous, which meant that files might still be active even though they passed through this callback.

app/javascript/controllers/dropzone_controller.js
```diff
@@ -51,6 +53,10 @@ export default class extends Controller {
       this.dropzone.element.classList.remove('dz-started')
       this.progressTarget.classList.add('d-none')
       this.updateProgress(0, true)
+      // Reload the files section to show the newly uploaded files.
+      this.dropzoneFilesOutlets.forEach(dropzoneFiles => dropzoneFiles.reload())
+      this.element.dataset.dropzoneReady = true
+      console.log(this.element.dataset)
```
mjgiarlo, line 59: Cruft?

spec/system/manage_files_spec.rb
```diff
@@ -21,6 +21,9 @@
       # Add one file
       # Can't test folder upload, so no hierarchy.
       find('.dropzone').drop('spec/fixtures/files/hippo.png')
+      # expect(page).to have_css('#file-dropzone[data-dropzone-ready="true"]')
+      # expect(page).to have_css('.dropzone-files[complete]')
```
mjgiarlo, line 25: Cruft?

mjgiarlo (approved):

mjgiarlo (approved):
Approve pending potential cruft removal

justinlittman:
@mjgiarlo Sorry, I thought I de-crufted.

PR #846, opened 2025-02-25
Fix default license entry in collection form.

### LINKED ISSUE #844: "No license" option in collection form is sticky
In the collection form, if you choose the "depositor selects" option for the license and the select anything other than "No license" for the default and then save, when you go back in to edit the collection this option will display "No license" again.

Steps to reproduce:
1. Create or edit a collection.
2. If creating a collection, fill in title and description (and contact, but that fix may not be pushed out yet).
3. On license tab, select the second option and choose any license other that "No license."
4. Save collection.
5. Edit collection.
6. Return to license tab and see that the second option is still selected but the license displayed is back to "No license."

justinlittman:
closes #844

### DIFF
--- a/app/components/collections/edit/license_component.html.erb
+++ b/app/components/collections/edit/license_component.html.erb
@@ -5,7 +5,7 @@
       <%= form.label :license_option_required, 'Require license for all deposits', class: 'form-check-label' %>
     </div>
     <div class="ps-2">
-      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, hidden_label: true, label:, width: 250, container_classes: 'mb-4') %>
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, hidden_label: true, label: 'Required license', width: 250, container_classes: 'mb-4') %>
     </div>
     <div class="ps-2">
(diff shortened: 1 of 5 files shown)


app/components/collections/edit/license_component.html.erb
```diff
@@ -5,7 +5,7 @@
       <%= form.label :license_option_required, 'Require license for all deposits', class: 'form-check-label' %>
     </div>
     <div class="ps-2">
-      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, hidden_label: true, label:, width: 250, container_classes: 'mb-4') %>
+      <%= render Elements::Forms::SelectFieldComponent.new(form:, field_name: :license, options: license_options, hidden_label: true, label: 'Required license', width: 250, container_classes: 'mb-4') %>
```
justinlittman, line 8: Gave each of these selects unique and descriptive labels.

app/forms/collection_form.rb
```diff
@@ -33,7 +33,7 @@ def persisted?
   validates :license_option, inclusion: { in: %w[required depositor_selects] }
 
   attribute :license, :string
-  attribute :default_license
+  attribute :default_license, :string
```
justinlittman, line 36: This was the fix.

mjgiarlo (approved):

PR #854, opened 2025-02-25
Adjust type of deposit layout.

### LINKED ISSUE #694: Subtypes - formatting issues
On the Type of deposit page, can you please adjust the following:
1. subtypes should be alphabetized down columns, not across rows
Current (H3 stage):
Image

Proposed (this is how it looks like in H2 now):
Image

2. The radio buttons and checkboxes should align left with the text headers. They currently start slightly to the left of the section headers.
Current (H3 stage):
Image

Proposed:
Image

3. In Chrome, if you click the 'See more options', the checkboxes in the left most column only render the right half of the checkbox, and then there is a slight delay before the other half of the checkboxes are rendered. This is a nitpicky item, but wondering why this happens.

justinlittman:
closes #694

image

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -307,3 +307,17 @@ fieldset.form-fieldset legend {
 .admin-card {
   --bs-card-bg: var(--stanford-10-black);
 }
+
+// Allow for types to list alphabetically in the vertical direction while
+// still maintaining responsive columns.
+.types-container {
+  columns: 3;
+
+  @include media-breakpoint-down(lg) {
+    columns: 2;
+  }
+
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -307,3 +307,17 @@ fieldset.form-fieldset legend {
 .admin-card {
   --bs-card-bg: var(--stanford-10-black);
 }
+
+// Allow for types to list alphabetically in the vertical direction while
+// still maintaining responsive columns.
+.types-container {
```
justinlittman, line 313: Borrowed from H2.

mjgiarlo (approved):

PR #856, opened 2025-02-25
[SPIKE] Replace `EdtfSupport` service class with a value object

mjgiarlo:
I'm continually side-eyeing the giant junk drawer `app/services/` is becoming and the sus `...Support` classes that tend to be little more than method buckets, so I'm looking for opportunities to extract objects from `app/services/` that feel more object-y in terms of their interfaces and responsibilities. Here's the latest attempt.


app/forms/date_form.rb
```diff
@@ -16,10 +16,10 @@ class DateForm < ApplicationForm
...
   def to_s
-    EdtfSupport.to_edtf_s(year:, month:, day:, approximate:)
+    to_date&.to_fs(:edtf)
   end
 
-  def to_edtf
-    EdtfSupport.to_edtf(year:, month:, day:, approximate:)
+  def to_date
+    EdtfDate.for(year:, month:, day:, approximate:)
   end
```
mjgiarlo, line 24: This method rename makes more sense to me for two reasons:

1. The returned object is a `Date` instance, not an EDTF instance. All `Date` instances can be EDTF-ified because the gem decorates all `Date` instances with EDTF goodness.
2. The `edtf` gem patches all `Date` instances so that they respond to `to_edtf`. Changing our method name here to something different avoids that confusion.

app/services/to_cocina/work/events_mapper.rb
```diff
@@ -34,16 +34,14 @@ def creation_date_event_params
...
+        CocinaGenerators::Description.event(type: 'creation', date:)
       end
 
       def publication_date_event_params
         date = publication_date.to_s
         return if date.nil?
 
-        CocinaGenerators::Description.event(type: 'publication', primary: true,
-                                            date:)
+        CocinaGenerators::Description.event(type: 'publication', primary: true, date:)
```
mjgiarlo, line 44: Arbitrary linebreaks?

app/values/edtf_date.rb
```diff
@@ -0,0 +1,30 @@
...
+    date = Date.new(*[year, month, day].compact)
+    if month.nil?
+      date.year_precision!
+    elsif day.nil?
+      date.month_precision!
+    end
+    date.approximate! if approximate
+    date
+  end
+end
```
mjgiarlo, line 30: Notice the value object that replaces the `EdtfSupport` "service" class does not itself define any methods to coerce to string. Why? Because Rails already has a convention for formatting `Date` objects as strings: https://api.rubyonrails.org/classes/Date.html#method-i-to_fs-label-Adding+your+own+date+formats+to+to_fs Let's build on that for our EDTF dates?

justinlittman (commented):
I'm "meh" on this. I doesn't make anything clearer for me, and having something an object called "EdtfDate" is likely to confuse me between that and the Edtf gem. But I don't feel very strongly either way.

thatbudakguy (approved):
I like this. The `EdtfSupport` class encapsulates no state and has only class methods, so I don't see any value in it over an object that actually holds date data.

I don't know about broader Rails conventions, but I usually expect a service to encapsulate something that callers don't need to know about, like an HTTP client instance and an API key.

On method naming, I think I'd expect something called `to_date` to return something structured, and `to_edtf` to return a string, which is what this PR does.

mjgiarlo:
@justinlittman OK if we ship this given the other reviews?

PR #857, opened 2025-02-26
[HOLD] Redirect authorized users to the dashboard from the homepage

### LINKED ISSUE #830: Return users who already have access to a collection should go straight to dashboard when going to sdr.stanford.edu
For users who have already received access to a collection, they should be taken to the dashboard when going to sdr.stanford.edu instead of having to go through the homepage.

aaron-collier:
Fixes #830

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -5,6 +5,8 @@ class HomeController < ApplicationController
   skip_verify_authorized only: :show
 
   def show
+    redirect_to dashboard_path if allowed_to?(:show?, :dashboard)
+
     @hero_image = "home#{rand(1..18)}.webp"
   end
 end
--- a/spec/system/show_home_spec.rb
+++ b/spec/system/show_home_spec.rb
@@ -4,18 +4,33 @@ require 'rails_helper'
 
 RSpec.describe 'Show home', :rack_test do
   include ActionView::Helpers::SanitizeHelper
-  before do
-    sign_in(create(:user))
+
+  context 'when the user is not authorized to see the dashboard' do
+    before do
+      sign_in(create(:user))
+    end
+
+    it 'shows the home page' do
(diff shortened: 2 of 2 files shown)


app/controllers/home_controller.rb
```diff
@@ -5,6 +5,8 @@ class HomeController < ApplicationController
   skip_verify_authorized only: :show
 
   def show
+    redirect_to dashboard_path if allowed_to?(:show?, :dashboard)
```
justinlittman, line 8: Here's a problem: This means the user can't get to the homepage, e.g., by clicking on "Stanford Digital Repository" in the header.

@amyehodge Is it acceptable that the user can't get to the home page?

mjgiarlo: This makes me wonder whether https://github.com/sul-dlss/hungry-hungry-hippo/issues/830 came out of user testing or if it's something we thought useful. Maybe it's more confusing than useful?

justinlittman: It came from me being annoyed that I had to click "Enter here" every time I went to the site.

amyehodge: @mjgiarlo @justinlittman I hadn't realized this would prevent them from ever getting there via clicking on the link in the header. I hadn't thought that through. I do think that's not desirable. Maybe we should discuss.

aaron-collier: @justinlittman thank you for adding hold, I meant to for this reason, thinking it needed discussion.

mjgiarlo: Maybe we should set this aside until one of our production or testing users asks for it.

aaron-collier: @mjgiarlo I'm totally OK with that. I'm going to draft this just so there is no expectation of it being ready to merge.

Issue #274, opened 2024-12-06
Add Terms of use tab to the collection form


aaron-collier:
https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H2-2024?node-id=789-4282&node-type=frame&t=3O5r5DrUmDQn71IY-0

- [x] create a `Collections::Edit::TermsOfUseComponent` based on the Figma design
- [x] render the component with a matching pane on the collection edit form
- [x] make the component set the values for `Collection.custom_rights_statement_option`
- [x] migration and changes to support ensuring `custom_rights_statement_instructions` is always set
- [ ] write a test?

PR #864, opened 2025-02-26
Ensure custom terms of use entry instructions are populated

thatbudakguy:
This implements a change discussed at standup that sets the
instructions for entering custom terms of use at import time,
so that they are never empty after migrating data from H2.

Previously, we were substituting empty instructions with a
default value at render time in the Work form. It was possible
to have selected 'depositor enters terms of use' but not
provide any instructions for doing that.

This also allows for renaming the field to something slightly
shorter and easier to remember.

Prep for #274, where we want to make this field in the
Collection form required and also pre-populate it with our
default.


justinlittman (approved):
Approve pending consideration of whether test is needed to cover untested line.

thatbudakguy:
It's a pre-existing line and not new, so not sure what codecov is on about...

PR #865, opened 2025-02-26
Restrict length of file description and whats changing fields

### LINKED ISSUE #683: Restrict length of file description and "What's changing" fields.
File description field should be a maximum of 200 characters.
What's changing field for both the work and collection forms should be a maximum of 100 characters.

mjgiarlo:
Fixes #683

### DIFF
--- a/app/components/elements/forms/text_field_component.html.erb
+++ b/app/components/elements/forms/text_field_component.html.erb
@@ -1,6 +1,6 @@
 <%= tag.div class: container_classes, hidden:, data: do %>
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, tooltip:, label_text: label, hidden_label:, classes: label_classes) %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.text_field field_name, class: 'form-control', required:, aria: field_aria, disabled:, data: input_data, placeholder:, style: styles, form: form_id %>
+  <%= form.text_field field_name, class: 'form-control', required:, aria: field_aria, disabled:, data: input_data, placeholder:, style: styles, form: form_id, maxlength:, size: %>
   <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name:, classes: error_classes) %>
 <% end %>
--- a/app/components/elements/forms/text_field_component.rb
+++ b/app/components/elements/forms/text_field_component.rb
@@ -4,9 +4,26 @@ module Elements
   module Forms
     # Component for form text fields
     class TextFieldComponent < FieldComponent
-      def initialize(**args)
+      # This is the default `size` attribute of the `<input>` element. See https://www.w3schools.com/TAGs/att_input_size.asp
+      DEFAULT_SIZE = 20
+
+      def initialize(maxlength: nil, size: nil, **args)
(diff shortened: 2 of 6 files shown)


app/components/elements/forms/text_field_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= tag.div class: container_classes, hidden:, data: do %>
   <%= render Elements::Forms::LabelComponent.new(form:, field_name:, tooltip:, label_text: label, hidden_label:, classes: label_classes) %>
   <%= render Elements::Forms::HelpTextComponent.new(id: help_text_id, help_text:) %>
-  <%= form.text_field field_name, class: 'form-control', required:, aria: field_aria, disabled:, data: input_data, placeholder:, style: styles, form: form_id %>
+  <%= form.text_field field_name, class: 'form-control', required:, aria: field_aria, disabled:, data: input_data, placeholder:, style: styles, form: form_id, maxlength:, size: %>
```
justinlittman, line 4: Why is size being set?

mjgiarlo: @justinlittman Good question. Because without a size set, the size of the box will default to the value of `maxlength` (giant).

justinlittman: I took out `size` locally and couldn't notice any difference in the UI. What am I missing?

mjgiarlo: @justinlittman Here's what I see in both Firefox and Brave (Chromium):

Screenshot from 2025-02-26 08-48-56

mjgiarlo: Maybe I see it because I use hot reloading and you don't? 🤷🏻‍♂️

justinlittman: No, that is the one part of the UI I didn't look. The reason it is a problem here is because it is in a `<td>` with no inherent width. I think you probably want to just add a width here (`app/views/content_files/edit.html`) like:

mjgiarlo: @justinlittman That produces a similar look & feel, but it still results in a bogus `size` on the `<input>` (the value set for `maxlength`). It also pushes the rightward fields over. In this case, my sense is that setting an explicit `size` on the input is more "correct" and more visually consistent. We only need to set the width **because** we've now set `maxlength` without setting `size`. See:

# Set the TD width

text_field_width_tweak

# Set the input size

text_field_size_tweak

mjgiarlo: If we want to do further design on the display of the description input, I'd prefer to ship what's here now and get a design for the box when editing (which I don't see in figma now) and a new ticket.

justinlittman: What I'm confused about is why the default size is "20"? Why "20"? Could it be "30"? When would I want to override the default?

mjgiarlo: @justinlittman Ah, another good question. The answer is that `20` is the default size in HTML. How about I toss in a new code comment to help us remember that `size` and `maxlength` are related and that there is default behavior?

justinlittman (approved):

PR #866, opened 2025-02-26
Ignore certain files.

### LINKED ISSUE #814: Exclude hidden files
We should exclude hidden files, like this ~$FebGoogleDMPToolStats.xlsx one shown below. I can't see it in the folder on my computer that I uploaded.

Whatever we've been excluding for weird files like this in H2 we probably need to continue doing in H3.

Image

justinlittman:
closes #814


app/services/ignore_file_service.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Service for determining if a file provided by a user should be ignored.
+class IgnoreFileService
+  # @return [Boolean] true if the file should be ignored, false otherwise
+  def self.call(filepath:)
```
justinlittman, line 6: Borrowed from H2.

lwrubel: Does this address the example in Amy's ticket of a temporary file with a name starting with `~$`?

justinlittman: It does now. It turns out these are MS Office temporary files and were not ignored in H2.

lwrubel (approved):

app/services/ignore_file_service.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  START_WITH_PATTERNS = ['__MACOSX', '._', '~$'].freeze
+  END_WITH_PATTERNS = ['.DS_Store', "Icon\r"].freeze
+
+  # @return [Boolean] true if the file should be ignored, false otherwise
+  def call
+    filepath.start_with?(*START_WITH_PATTERNS) \
+    || File.basename(filepath).start_with?(*START_WITH_PATTERNS) \
+    || filepath.end_with?(*END_WITH_PATTERNS)
```
mjgiarlo, line 20:
```suggestion
    filepath.start_with?(*START_WITH_PATTERNS) || 
      File.basename(filepath).start_with?(*START_WITH_PATTERNS) ||
      filepath.end_with?(*END_WITH_PATTERNS)
```

PR #868, opened 2025-02-26
Remove collection edit pencil.

### LINKED ISSUE #860: Remove pencil icon on Collections view page
Remove the pencil icon since we already have the Edit button.

Image

justinlittman:
closes #860


lwrubel:
`Edit::EditIconButtonComponent` is not used anywhere else at the moment. It was added for the pencil icon in the Collection show page. If we're not going to need it for some other future pencil you could delete it, I think.

PR #872, opened 2025-02-26
Refactor of multiline conjunction

mjgiarlo:
Per #871 

This is the last of this pattern in H3. @justinlittman do you know if your editor does this automagically for some reason? Interesting choice, if so!


justinlittman (approved):

justinlittman:
No, I think this is a bad pattern of my own design.

justinlittman (approved):

PR #874, opened 2025-02-26
Go directly to edit tabs from show page.

### LINKED ISSUE #862: Go directly to the tab in the deposit form
I know we had discussed this previously, and it was challenging, but it would be very helpful to our users if they can click on a specific section like "Access Settings" and go directly to that page in the deposit form.

justinlittman:
closes #862

https://github.com/user-attachments/assets/c94bae7b-fb6f-4082-9db1-f6565476192a

### DIFF
--- a/app/components/edit/edit_icon_button_component.rb
+++ b/app/components/edit/edit_icon_button_component.rb
@@ -3,20 +3,21 @@
 module Edit
   # Component for a pencil icon button to edit a draft
   class EditIconButtonComponent < Elements::IconButtonLinkComponent
-    def initialize(presenter:, icon_classes:, **args)
+    def initialize(presenter:, tab: nil, **args)
       @presenter = presenter
-      @icon_classes = icon_classes
+      @tab = tab
       # polymorphic_path cannot be used in initializer, so overrriding link below
       args[:link] = nil
-      super(icon: :edit, label: 'Edit', icon_classes:, **args)
+      args[:data] = { turbo_frame: '_top', turbo_preserve_scroll: false }
+      super(icon: :edit, label: 'Edit', **args)
     end
 
     def link
-      edit_polymorphic_path(@presenter)
+      edit_polymorphic_path(@presenter, tab: @tab)
     end
 
     def render?
-      @presenter.editable?
+      @presenter&.editable?
     end
   end
 end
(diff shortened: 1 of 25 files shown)


app/components/edit/edit_icon_button_component.rb
```diff
@@ -3,16 +3,18 @@
 module Edit
   # Component for a pencil icon button to edit a draft
   class EditIconButtonComponent < Elements::IconButtonLinkComponent
-    def initialize(presenter:, icon_classes:, **args)
+    def initialize(presenter:, icon_classes: [], tab: nil, **args)
```
mjgiarlo, line 6: Just curious: I only see one use of `EditIconButtonComponent` in H3, and it does not inject `icon_classes`. Why we do we need that arg here?  Maybe I've missed some metaprogramming?

app/components/show/table_heading_component.rb
```diff
@@ -2,11 +2,21 @@
 
 module Show
   # Component for rendering a table heading.
-  class TableHeadingComponent < Elements::HeadingComponent
-    def initialize(**args)
-      args[:level] = :h2
```
mjgiarlo, line 7: What happened to the `h2` concern here? No longer relevant due to other changes?

mjgiarlo (approved):
Approve pending consideration of non-blocking questions

app/components/edit/edit_icon_button_component.rb
justinlittman: It's cruft. Removing.

app/components/show/table_heading_component.rb
justinlittman: Yes. Before TableHeadingComponent was a subclass of Elements::HeadingComponent. It now renders Elements::HeadingComponent and provides the level as an argument.

app/components/show/table_heading_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render Elements::HeadingComponent.new(level: :h2, classes:) do %>
+  <%= text %><%= render Edit::EditIconButtonComponent.new(presenter:, tab:) if presenter.present? %>
```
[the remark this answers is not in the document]
justinlittman: Seems cleaner to me. Fixing.

PR #875, opened 2025-02-26
Adjust layout and provide prompts for collection license tab.

### LINKED ISSUE #863: Collection form - License page
Could we make a few adjustments to the License page in the Collection form?

Here is what it looks like on stage:

Image

Here is what we'd like it to look like:
1. The 2 drop downs should not have a default value.
2. First drop down default value should be "Select an option..."
3. Second drop down default value should be "Select license to display by default..."
4. Bonus points if the 2 dropdowns are left aligned! (this is how it currently is in H2).

Image

justinlittman:
closes #863

image

### DIFF
--- a/config/licenses.yml
+++ b/config/licenses.yml
@@ -82,7 +82,7 @@
   deprecated: false
   label: "PDDL-1.0 Public Domain Dedication and License"
   group: "Open Data Commons (ODC) licenses"
-- id: ""
+- id: "no-license"
   deprecated: false
   label: "No License"
   group: "Other"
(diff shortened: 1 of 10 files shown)


config/licenses.yml
```diff
@@ -82,7 +82,7 @@
   deprecated: false
   label: "PDDL-1.0 Public Domain Dedication and License"
   group: "Open Data Commons (ODC) licenses"
-- id: ""
+- id: "no-license"
```
justinlittman, line 85: Using "" to indicate was no license was problematic because it conflicted with use of "" for a prompt. Opted to change this to stay on the rails.

mjgiarlo (approved):

PR #877, opened 2025-02-27
Improve test coverage

mjgiarlo:
My strategy for this spike was to improve test coverage by asking, for each test coverage gap:

1. Is this dead code? Remove it. 
2. Is this useful to test? Test it. 
3. Is it more annoying to test than useful? Skip it.


.rubocop.yml
```diff
@@ -65,6 +65,10 @@ RSpec/InstanceVariable:
   Exclude:
     - spec/system/**/* # We use instance variables in system specs to stub Cocina operations
 
+RSpec/LetSetup:
+  Exclude:
+    - spec/requests/**/*
```
mjgiarlo, line 70: It's useful to be able to create objects when setting up request specs.

app/controllers/works_controller.rb
```diff
@@ -143,8 +143,8 @@ def check_deposit_registering_or_updating
   def set_work_form_from_cocina
     @cocina_object = Sdr::Repository.find(druid: params[:druid])
     version_description = @version_status.open? ? @version_status.version_description : nil
-    @work_form = ToWorkForm::Mapper.call(cocina_object: @cocina_object,
-                                         doi_assigned: doi_assigned?, agree_to_terms: current_user.agree_to_terms?,
+    @work_form = ToWorkForm::Mapper.call(cocina_object: @cocina_object, doi_assigned: doi_assigned?,
+                                         agree_to_terms: current_user.agree_to_terms?,
```
mjgiarlo, line 147: Whitespace formatting change.

app/forms/collection_form.rb
```diff
@@ -49,10 +49,18 @@ def persisted?
...
   attribute :email_depositors_status_changed, :boolean, default: true
-end
 
-def duration_must_be_present
-  return if release_duration.present?
+  def duration_must_be_present
+    return if release_duration.present?
+
+    errors.add(:release_duration, 'select a valid duration for release')
+  end
```
mjgiarlo, line 57: I was surprised to find this method floating outside the form class, so I shifted it up into the form.

app/forms/collection_form.rb
```diff
@@ -49,10 +49,18 @@ def persisted?
...
 
-  errors.add(:release_duration, 'select a valid duration for release')
+  def selected_license
+    if license_option == 'required' && license.present?
+      license
+    elsif license_option == 'depositor_selects' && default_license.present?
+      default_license
+    end
+  end
```
mjgiarlo, line 65: Extracted from `ToCocina::Collection::Mapper`

app/helpers/icon_helper.rb
```diff
@@ -9,10 +9,6 @@ def icon(icon_classes:, classes: nil, **)
     content_tag(:i, nil, class: all_classes, **)
   end
 
-  def right_arrow_icon(**)
-    icon(icon_classes: 'bi bi-arrow-right', **)
-  end
```
mjgiarlo, line 14: Dead code.

app/mailers/application_mailer.rb
```diff
@@ -18,8 +18,4 @@ def set_work_presenter
                                         version_description: version_status.version_description)
     @work_presenter = WorkPresenter.new(work: @work, work_form:, version_status:)
   end
-
-  def doi_assigned?
-    DoiAssignedService.call(cocina_object: @cocina_object, work: @work)
-  end
```
mjgiarlo, line 24: Dead code.

app/policies/work_policy.rb
```diff
@@ -28,9 +28,8 @@ def collection_manager?
   end
 
   def collection_depositor?
-    return record.depositors.include?(user) if record.is_a? Collection
-
-    record.collection.depositors.include?(user)
```
mjgiarlo, line 33: This was dead code. `WorksController#create` *only* authorizes collections, not works, because there is no work to authorize yet.

app/services/cocina/prettier.rb
```diff
@@ -3,18 +3,6 @@
...
-      new(...).clean
-    end
-
-    def self.pretty(...)
-      new(...).pretty
-    end
-
-    def self.json(...)
-      new(...).json
-    end
```
mjgiarlo, line 16: Dead code.

app/services/contents/analyzer.rb
```diff
@@ -37,39 +37,14 @@ def updates_for(content_file)
...
-      [{ md5_digest: md5.hexdigest, sha1_digest: sha1.hexdigest }, first_buffer]
-    end
-
-    def base64_to_hex(base64_string)
-      binary_data = Base64.decode64(base64_string)
-      binary_data.unpack1('H*')
-    end
 
-    def zip_filepath
-      ActiveStorageSupport.filepath_for_blob(content.zip_file.blob)
```
mjgiarlo, line 72: Dead code.

app/services/contents/stager.rb
```diff
@@ -45,9 +45,5 @@ def staging_content_path
     def create_directory(filepath)
       FileUtils.mkdir_p File.dirname(filepath)
     end
-
-    def zip_filepath
-      ActiveStorageSupport.filepath_for_blob(content.zip_file.blob)
-    end
```
mjgiarlo, line 51: Dead code

app/services/importers/collection.rb
```diff
@@ -7,8 +7,8 @@ def self.call(...)
       new(...).call
     end
 
-    def initialize(collection_json:)
-      @collection_json = collection_json
+    def initialize(collection_hash:)
+      @collection_hash = collection_hash
```
mjgiarlo, line 11: Just a rename to be clearer that this thing is a Ruby hash, not a JSON string.

app/services/to_collection_form/roundtrip_validator.rb
```diff
@@ -4,7 +4,7 @@ module ToCollectionForm
   # Validates that a cocina object can be converted to a collection form and then back without loss.
   class RoundtripValidator
     def self.roundtrippable?(...)
-      new(...).call
+      new(...).roundtrippable?
```
mjgiarlo, line 7: We tend to name the class method after the instance method.

app/services/to_form/note_mapper.rb
```diff
@@ -6,10 +6,6 @@ class NoteMapper
     ABSTRACT_TYPE = 'abstract'
     CITATION_TYPE = 'preferred citation'
 
-    def self.call(...)
-      new(...).call
-    end
-
```
mjgiarlo, line 12: Dead code.

app/services/to_work_form/related_works_mapper.rb
```diff
@@ -30,15 +30,10 @@ def related_work_values_from(related_resource) # rubocop:disable Metrics/AbcSize
...
+      elsif related_resource.identifier&.first&.[](:uri).present?
+        [nil, related_resource.identifier&.first&.[](:uri)]
       end
-
-      if related_resource.identifier&.first&.[](:uri).present?
-        return [nil,
-                related_resource.identifier&.first&.[](:uri)]
-      end
-
-      []
```
mjgiarlo, line 41: Unreachable branch.

app/services/to_work_form/roundtrip_validator.rb
```diff
@@ -4,7 +4,7 @@ module ToWorkForm
   # Validates that a cocina object can be converted to a work form and then back without loss.
   class RoundtripValidator
     def self.roundtrippable?(...)
-      new(...).call
+      new(...).roundtrippable?
```
mjgiarlo, line 7: See above.

lib/tasks/import.rake
```diff
@@ -11,12 +11,12 @@ namespace :import do
   task :collections, [:filename] => :environment do |_t, args|
     Rails.application.config.action_mailer.perform_deliveries = false
 
-    collections_json = JSON.parse(File.read(args[:filename] || 'collections.json'))
-    collections_json.each do |collection_json|
-      next unless collection_json['druid']
+    collections_hash = JSON.parse(File.read(args[:filename] || 'collections.json'))
+    collections_hash.each do |collection_hash|
+      next unless collection_hash['druid']
```
mjgiarlo, line 16: See above.

spec/models/collection_spec.rb
```diff
@@ -4,96 +4,104 @@
 
 RSpec.describe Collection do
   let(:user) { create(:user) }
-  let(:collection) { described_class.create(title: collection_title_fixture, user:) }
+  let(:collection) { create(:collection, title: collection_title_fixture, user:) }
```
mjgiarlo, line 7: Prefer factories.

spec/models/collection_spec.rb
```diff
@@ -4,96 +4,104 @@
...
 
   before do
     allow(Notifier).to receive(:publish)
   end
 
-  describe 'Add participants to a collection' do
```
mjgiarlo, line 13: I judged this an unnecessary `describe` block. Just makes the spec output noisier.

spec/models/collection_spec.rb
```diff
@@ -4,96 +4,104 @@
...
-    context 'when adding a manager to collection' do
-      it 'only sends a MANAGER_ADDED event' do
-        collection.managers << user
-        expect(Notifier).to have_received(:publish).with(Notifier::MANAGER_ADDED, collection:,
-                                                                                  user:)
-        expect(Notifier).not_to have_received(:publish).with(Notifier::MANAGER_REMOVED)
```
mjgiarlo, line 19: Instead of checking if this is not received, change the expectation below that `:published` is ONLY received once.

spec/models/collection_spec.rb
```diff
@@ -4,96 +4,104 @@
...
+  context 'when adding a reviewer' do
+    it 'sends only a REVIEWER_ADDED event' do
+      collection.reviewers << user
+      expect(Notifier).to have_received(:publish).once.with(Notifier::REVIEWER_ADDED, collection:, user:)
     end
   end
 
-  describe 'Remove participants from a collection' do
-    context 'when removing a manager to collection' do
-      let(:manager) { create(:user) }
```
mjgiarlo, line 44: We already have a user above, no need to create more.

spec/services/sdr/repository_spec.rb
```diff
@@ -255,12 +255,10 @@
     let(:version_client) do
       instance_double(Dor::Services::Client::ObjectVersion, status: version_status, discard: true)
     end
-
```
mjgiarlo, line 258: Arbitrary whitespace. Remove it.

app/helpers/icon_helper.rb
justinlittman: Not anymore: https://github.com/sul-dlss/hungry-hungry-hippo/pull/878/files#diff-3e3bacebbd8b846ea714d0285efdd51b7a909b8690c3e3265f3714a1f2e1281bR2

app/services/cocina/prettier.rb
justinlittman: Dead code, but by convention we have static methods for most of our services. If I were to go to use this service in the future, I'd be surprised that it didn't have static methods.

(I don't feel too strongly about this.)

app/services/to_collection_form/roundtrip_validator.rb
justinlittman: Then I would change the static method to `call`. When our services have a single method, that method is usually (always?) `call`.

app/services/to_form/note_mapper.rb
justinlittman: I'd leave it in for the reasons previously noted.

app/services/to_work_form/roundtrip_validator.rb
justinlittman: Same as above.

spec/models/collection_spec.rb
justinlittman: Yes, but this is a test of the Collection class so I'd suggest it should be tested directly instead of via a factory.

spec/services/to_collection_form/roundtrip_validator_spec.rb
```diff
@@ -3,31 +3,43 @@
...
+  subject(:validator) do
+    described_class.new(collection_form: collection_form_fixture, cocina_object:)
   end
 
   context 'when roundtrippable' do
     let(:cocina_object) { collection_with_metadata_fixture }
 
     it 'returns true' do
-      expect(roundtrippable?).to be true
+      expect(validator).to be_roundtrippable
```
aaron-collier, line 14: Nice.

aaron-collier (approved):

spec/services/contents/analyzer_spec.rb
```diff
@@ -3,8 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe Contents::Analyzer do
-  let(:content) { create(:content) }
+  include WorkMappingFixtures
```
justinlittman, line 6: Why?

spec/services/to_form/base_mapper_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ToForm::BaseMapper do
+  describe '#call' do
+    let(:cocina_object) { build(:dro, title: title_fixture) }
+
+    it 'raises' do
+      expect { described_class.call(cocina_object:) }.to raise_error(NotImplementedError)
```
justinlittman, line 10: This seems unnecessary to test.

justinlittman (commented):
A couple quibbles, nothing major.

app/helpers/icon_helper.rb
mjgiarlo: @justinlittman Thanks!

app/services/cocina/prettier.rb
mjgiarlo: I don't feel *very* strongly, but maybe a bit more strongly than you in this case 😆 Why? My sense is that we generally prefer a just-in-time approach to a just-in-case one, lest we accumulate dead, untested, YAGNI code. I'm in favor of dropping these until we need them.

app/services/to_collection_form/roundtrip_validator.rb
mjgiarlo: I suspect you're right about that. I guess what feels different about this class is it doesn't feel like a service to me, but rather a validator, so I may try to recast this as a validator, which more naturally would have a single question-mark method?

mjgiarlo: It doesn't really work as a validator. I'm inclined to agree with you here, so I'm standardizing on `.call` for these two classes. 🍻

app/services/to_form/note_mapper.rb
mjgiarlo: I'm still inclined to kill it until we need it. The `NoteMapper` wants to be opinionated about whether it's being invoked to map an abstract or a citation, so I'm not sure I see a near-term future where it'll need `.call`.

app/services/to_work_form/roundtrip_validator.rb
mjgiarlo: Reverted!

spec/models/collection_spec.rb
mjgiarlo: @justinlittman `FactoryBot.create(:collection)` does create an instance of the `Collection` class and it doesn't seem to have drawbacks. And it gets us FactoryBot goodness like automagic clean-up.

spec/services/contents/analyzer_spec.rb
mjgiarlo: Good catch. This itself is dead code because I wound up removing the test coverage I added for `Contents::Analyzer` which was flappy. Will remove this.

spec/services/to_form/base_mapper_spec.rb
mjgiarlo: I was on the fence about this one. Good call.

PR #878, opened 2025-02-27
Rejigger work form footer buttons.

### LINKED ISSUE #774: Add "cancel" button back to deposit form
After testing, we have determined that we still need a cancel button on the work deposit form.

The current "Discard draft" button completely discards any current work as well as any previously saved draft.
We need the "Cancel" button to allow users to back out of the currently open form without losing any draft currently in progress. 

This button needs to be present all the time. 

@astridu Can you let us know where to put this back in the UI? Thanks.

justinlittman:
closes #774

image


app/javascript/controllers/tab_nav_controller.js
```diff
@@ -0,0 +1,24 @@
...
+import * as bootstrap from 'bootstrap'
+
+export default class extends Controller {
+  static values = { selector: String }
+
+  showNext (event) {
+    event.preventDefault()
+
+    const nextTab = this.currentTab.nextElementSibling
+    bootstrap.Tab.getOrCreateInstance(nextTab).show() // eslint-disable-line no-undef
```
mjgiarlo, line 11: Did you find you needed the eslint disables after having the `import` on line 2?

app/javascript/controllers/tab_nav_controller.js
```diff
@@ -0,0 +1,24 @@
...
+
+  showPrevious (event) {
+    event.preventDefault()
+
+    const previousTab = this.currentTab.previousElementSibling
+    bootstrap.Tab.getOrCreateInstance(previousTab).show() // eslint-disable-line no-undef
+  }
+
+  get currentTab () {
+    return document.querySelector(`#${this.selectorValue}`)
```
mjgiarlo, line 22: Trivial, but: do you think adding the `#` here makes more sense than when setting the selector value in rubyland? This handles the concern in one place (vs. three), but it seems a little unexpected to set a selector value in rubyland that is inherently missing key information. It isn't really a fully useful selector when set in Rubyland.

Either way's fine, just curious.

mjgiarlo (approved):
Approve pending consideration of feedback around the Stimulus controller

app/javascript/controllers/tab_nav_controller.js
justinlittman: No, not needed. Removing.

PR #881, opened 2025-02-27
Add search by druid admin function.

### LINKED ISSUE #537: Move search for druid to Admin dashboard
This functionality is currently available in the drop-down menu, but I'd like it put on the dashboard directly since we use this a lot.

It will now appear with the drop-down list of other options and the new search for collection (#536).

justinlittman:
closes #537

https://github.com/user-attachments/assets/14ea5f47-aee1-42ca-916c-e13d55be8e3b

### DIFF
--- a/app/controllers/admin/druid_search_controller.rb
+++ b/app/controllers/admin/druid_search_controller.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller for searching for work or collection by druid
+  class DruidSearchController < Admin::ApplicationController
+    def new
+      authorize!
+
+      @druid_search_form = DruidSearchForm.new
+    end
+
+    def search
+      authorize!
+
+      @druid_search_form = DruidSearchForm.new(druid_search_form_params)
+      if @druid_search_form.valid?
+        redirect_to_work_or_collection
+      else
+        render :new, status: :unprocessable_entity
+      end
+    end
+
+    private
+
+    def redirect_to_work_or_collection
+      path = if @druid_search_form.collection.present?
+               collection_path(druid: @druid_search_form.druid)
+             else
+               work_path(druid: @druid_search_form.druid)
+             end
+      render turbo_stream: turbo_stream.action(:full_page_redirect, path)
+    end
+
+    def druid_search_form_params
+      params.expect(admin_druid_search: DruidSearchForm.user_editable_attributes)
+    end
+  end
+end
--- a/spec/requests/admin/druid_search_spec.rb
+++ b/spec/requests/admin/druid_search_spec.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Search for druid' do
+  describe 'GET /admin/druid_search/new' do
+    context 'when the user is not authorized' do
+      before do
+        sign_in(create(:user))
+      end
+
+      it 'redirects to root' do
+        get '/admin/druid_search/new'
+
+        expect(response).to redirect_to(root_path)
+      end
+    end
+
+    context 'when the user is an admin' do
+      before do
+        sign_in(create(:user), groups: ['dlss:hydrus-app-administrators'])
+      end
+
+      it 'shows search form' do
+        get '/admin/druid_search/new'
+
+        expect(response).to have_http_status(:ok)
+      end
+    end
+  end
+
+  describe 'GET /admin/druid_search/search' do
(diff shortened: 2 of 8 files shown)


app/controllers/admin/druid_search_controller.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller for searching for work or collection by druid
+  class DruidSearchController < Admin::ApplicationController
+    def new
+      authorize!
```
mjgiarlo, line 7: I assume this controller is automagically using this policy: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/policies/admin_policy.rb Confirm/deny?

justinlittman: It's not magic, it's code: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/controllers/admin/application_controller.rb

app/controllers/admin/druid_search_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    private
+
+    def redirect_to_work_or_collection
+      path = if @druid_search_form.collection.present?
+               collection_path(druid: @druid_search_form.druid)
+             else
+               work_path(druid: @druid_search_form.druid)
+             end
```
mjgiarlo, line 30: I'm guessing this lives in the controller instead of the `DruidSearchForm` because the latter does not have path helpers?

spec/requests/admin/druid_search_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(response).to redirect_to(root_path)
+      end
+    end
+
+    context 'when the user is an admin' do
+      before do
+        sign_in(create(:user), groups: ['dlss:hydrus-app-administrators'])
+      end
+
+      it 'show search form' do
```
mjgiarlo, line 24:
```suggestion
      it 'shows search form' do
```

app/controllers/admin/druid_search_controller.rb
justinlittman: Yeah, that feels like overloading the form. I don't think any of our other forms support similar.

spec/requests/admin/druid_search_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      context 'when no results' do
+        it 'show search results' do
+          get '/admin/druid_search/search?admin_druid_search[druid]=foo&commit=Submit'
+
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(response.body).to include('not found')
+        end
+      end
+    end
```
mjgiarlo, line 58: Since there's no setup for this `context` block, and there's no test in the surrounding block here for when there *are* results, I'd be tempted to dump the inner `context` block and amend the surrounding `context` with more text. That sounded confusing when I typed it out, so _in lieu_ of all that text:

```suggestion
    context 'when the user is an admin and no results' do
      before do
        sign_in(create(:user), groups: ['dlss:hydrus-app-administrators'])
      end

      it 'show search results' do
        get '/admin/druid_search/search?admin_druid_search[druid]=foo&commit=Submit'

        expect(response).to have_http_status(:unprocessable_entity)
        expect(response.body).to include('not found')
      end
    end
```

spec/system/admin/druid_search_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    before do
+      create(:work, druid:, title: title_fixture)
+    end
+
+    it 'redirects to the work page' do
+      visit admin_dashboard_path
+
+      expect(page).to have_css('h1', text: 'Admin dashboard')
+
+      fill_in 'Search for DRUID', with: druid.delete_prefix('druid:')
```
mjgiarlo, line 52: Worth calling out that this test tests with a bare druid?

app/controllers/admin/druid_search_controller.rb
mjgiarlo: Sweet! TIL, thanks

app/controllers/admin/druid_search_controller.rb
mjgiarlo: @justinlittman Agree, this makes more sense

mjgiarlo (approved):
Approve pending consideration of test feedback

spec/system/admin/druid_search_spec.rb
justinlittman: Yes. Added comment.

Issue #351, opened 2024-12-13
Menu handling for mobile


astridu:
- For the mobile view, can we convert the top level nav to a hamburger menu that appears when user clicks the hamburger, 
- and the left nav converts to a dropdown menu. There should be some spacing inbetween each menu item so user can comfortably select the items in the dropdown menu. The page the user is on should be bold in the menu. 

![image](https://github.com/user-attachments/assets/d8e64737-ef2c-46b5-ad8c-38a88ba229c1)

justinlittman:
@mjgiarlo This is done, right?

mjgiarlo:
@justinlittman Yep, closing.

PR #885, opened 2025-02-28
Display form tab list as columns at all widths

mjgiarlo:
Connects to #351

Peek 2025-02-27 16-13

### DIFF
--- a/app/components/edit/tab_form/tab_component.rb
+++ b/app/components/edit/tab_form/tab_component.rb
@@ -36,7 +36,7 @@ module Edit
       end
 
       def classes
-        merge_classes('nav-link', selected? ? 'active' : nil)
+        merge_classes('nav-link w-100', selected? ? 'active' : nil)
       end
 
       def id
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@amyehodge @astridu I already see a hamburger menu for main nav at smaller widths. For the form nav, Bootstrap nav pills are not a dropdown menu so they don't render as a dropdown menu. What I did do was ensure the pills always display as a vertical list instead of collapsing to rows, so in that way it stays consistent with the full size width and is spiritually similar to a dropdown. Is this sufficient for now?

app/components/edit/tab_form/tab_list_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= tag.div class: classes, 'data-controller': 'tab-error' do %>
   <div class="col-md-3">
-    <%= tag.div class: 'nav nav-pills d-flex flex-md-column rounded-top rounded-bottom', role: 'tablist', 'aria-orientation': 'vertical' do |component| %>
+    <%= tag.div class: 'nav nav-pills d-flex flex-column rounded-top rounded-bottom', role: 'tablist', 'aria-orientation': 'vertical' do |component| %>
```
mjgiarlo, line 3: Always display vertically. This is consistent with the `aria-orientation`.

app/components/edit/tab_form/tab_component.rb
```diff
@@ -36,7 +36,7 @@ def selected?
       end
 
       def classes
-        merge_classes('nav-link', selected? ? 'active' : nil)
+        merge_classes('nav-link w-100', selected? ? 'active' : nil)
```
mjgiarlo, line 39: Never place form tabs beside one another. Keep them as a stack of vertical pills.

amyehodge:
This is certainly a huge improvement. I can live with this.

justinlittman (approved):

PR #887, opened 2025-02-28
Implement deleting an item as an admin

### LINKED ISSUE #728: Ability for admin to delete an item
Add ability for admin to delete an item from H3. This is a replacement for our current Decommission function that was more complex.

aaron-collier:
Fixes #728 

It looks like the only uncovered line here is the `unprocessible entity` state, which we don't test in the move_controller from what I can tell, so not sure why that one line would fail this... 

https://github.com/user-attachments/assets/b5e9544d-c579-4e19-b1c5-d176d7106a3c

Note: This redirects back to the dashboard on delete - if we'd like to redirect to the collection page that's an easy change, however if we want to redirect to either the dashboard or the collection depending on where the user has come from there is an odd situation on the works page (a refresh I think) that makes it challenging to keep the referrer available, happy to continue working on that as a follow up if we wish.


amyehodge:
@aaron-collier I think redirecting to the dashboard is fine.

app/controllers/admin/delete_work_controller.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller for deleting a work as an administrator
+  class DeleteWorkController < Admin::ApplicationController
+    def new
+      authorize!
+
+      @delete_form = Admin::DeleteForm.new
```
justinlittman, line 9: Using a form isn't needed for this function.

app/controllers/admin/delete_work_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      render :form
+    end
+
+    def create
+      authorize!
+
+      @delete_form = Admin::DeleteForm.new(work_form:)
+      if @delete_form.valid?
```
justinlittman, line 17: There's nothing to be validated.

app/services/admin/delete_work.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Admin
```
justinlittman, line 3: A service isn't necessary because all you're doing is calling work.destroy.

app/controllers/admin/delete_work_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  # Controller for deleting a work as an administrator
+  class DeleteWorkController < Admin::ApplicationController
+    def new
+      authorize!
+
+      @delete_form = Admin::DeleteForm.new
+      render :form
+    end
+
+    def create
```
justinlittman, line 13: This should probably be a destroy method instead of create.

app/controllers/admin/delete_work_controller.rb
aaron-collier: I'll work through this, but what I found was that not having a form screws up how we render form fields and I didn't want to have a form that goes away from our establish patterns.

aaron-collier: That was the special sauce I was missing!

PR #889, opened 2025-02-28
Sort collections on dashboard.

### LINKED ISSUE #800: Collections on the dashboard should appear in alphabetical order
Right now they appear to move around. This is confusing. If we make this alphabetical they will always appear in the same order, unless the title changes.

justinlittman:
closes #800

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -7,20 +7,27 @@ class DashboardController < ApplicationController
   def show
     return render :welcome_form unless allowed_to?(:show?, :dashboard)
 
+    set_draft_works_and_status_map
+    @your_collections = current_user.your_collections.to_a.sort_by!(&:title)
+  end
+
+  private
+
+  def druid_to_status_map
+    @druid_to_status_map ||= begin
+      druids = current_user.your_works.where.not(druid: nil).pluck(:druid)
+      Sdr::Repository.statuses(druids:)
+    end
+  end
+
+  def set_draft_works_and_status_map
     # This is going to be slow because it is going to block rendering since it is synchronous.
     # Also, it processes all of a user's works, but not all may be rendered on the dashboard.
-    druid_to_status_map = Sdr::Repository.statuses(druids: druids_for_works_for_current_user)
     @draft_works = []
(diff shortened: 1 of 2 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -16,6 +16,7 @@ def show
       @draft_works << work if version_status.draft? && work.user == current_user
       [work.id, version_status]
     end
+    @your_collections = current_user.your_collections.to_a.sort_by!(&:title)
```
justinlittman, line 19: Unfortunately the underlying query is too complicated to just slap an `order` on it.

app/controllers/dashboard_controller.rb
```diff
@@ -7,20 +7,27 @@ class DashboardController < ApplicationController
   def show
     return render :welcome_form unless allowed_to?(:show?, :dashboard)
 
+    set_draft_works_and_status_map
+    @your_collections = current_user.your_collections.to_a.sort_by!(&:title)
```
justinlittman, line 11: This is the real change.

Unfortunately the underlying query is too complicated to just slap an order on it.

mjgiarlo: @justinlittman Naïve question incoming: why is it complicated? Is it because the joins on `User#your_collections` mean calling `.order(:title)` would not work as expected?

mjgiarlo (approved):

app/controllers/dashboard_controller.rb
justinlittman: Yes, PG is not happy with the SQL when just add `.order(:title)`. I tried a number of variations, but couldn't get any to work. (I even asked Copilot who confidently told me to add `.order(:title)`.

mjgiarlo: @justinlittman FWIW, when I hop in the console on stage, I seem to be able to get results ordered by title via:

```ruby
user = User.first # or any of them really
user.your_collections.order(:title)
```

Odd.

justinlittman: I can verify that works, but don't understand why.

https://github.com/sul-dlss/hungry-hungry-hippo/pull/896

PR #890, opened 2025-02-28
Include contributor citation status on work show page.

### LINKED ISSUE #823: Indicate on view page who is included in the citation
@astridu I don't think we have this in the mockups anywhere because we changed how we were doing authors after the mockup was created.

Let's try to come up with something here.

justinlittman:
closes #823

image


mjgiarlo (approved):
Approve pending consideration of comment

PR #892, opened 2025-02-28
Adds depositing email for collection managers.

### LINKED ISSUE #870: Add email notification to managers when an item is deposited
Send an email to all collection managers when an item has been deposited. This is for when the review workflow is not on and items go straight to depositing.

If a collection manager is the one performing the task, they should only get the depositor notification about the deposit and not this one.

This text will be slightly different than what is in H2, because that's proven a bit confusing. 

This is the text:

> Dear [user name],
> 
> The item [item.title] has been deposited in the [collection name] collection by [owner name], a collection manager, or an SDR administrator.
> 
> If you have any questions, contact the SDR team at: <%= link_to contact_form_url, contact_form_url %>.
> 
> The Stanford Digital Repository Team

justinlittman:
closes #870

### DIFF
--- a/spec/jobs/deposit_collection_job_spec.rb
+++ b/spec/jobs/deposit_collection_job_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe DepositCollectionJob do
   before do
     allow(ToCocina::Collection::Mapper).to receive(:call).and_call_original
     allow(Sdr::Repository).to receive(:accession)
+    allow(Settings.notifications).to receive(:enabled).and_return(false)
   end
 
   context 'when a new collection' do
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -5,7 +5,7 @@ class ApplicationMailer < ActionMailer::Base
   layout 'mailer'
 
   def set_work
-    @work = params[:work]
+    @work = params[:work] || params[:object]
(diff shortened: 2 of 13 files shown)


spec/jobs/deposit_collection_job_spec.rb
```diff
@@ -11,6 +11,7 @@
   before do
     allow(ToCocina::Collection::Mapper).to receive(:call).and_call_original
     allow(Sdr::Repository).to receive(:accession)
+    allow(Settings.notifications).to receive(:enabled).and_return(false)
```
mjgiarlo, line 14: Do we usually want `Settings.notifications` to be `true` when running tests, but `false` in job specs?

mjgiarlo (commented):
Question about notifications and testing

spec/jobs/deposit_collection_job_spec.rb
justinlittman: Hmmm ... since only triggers emails, probably mostly want it to be false. Maybe worth a follow-up refactor to default to false for tests and only enable for tests when necessary?

mjgiarlo: @justinlittman Sounds good. 🙂

mjgiarlo (approved):

PR #893, opened 2025-02-28
Display relationship type of work show page.

### LINKED ISSUE #665: Display relation type for related published works separately on item show page
Currently this info is being appended in parentheses to the end of the text entered for the work, making it look like the user entered it there. I think that will be confusing

@astridu I think we should pull this out into its own line in the table. Can you mock that up and include what it looks like when there are multiple works, each with their own relationship? 

And I think the text the user sees in the UI would be better. What's displayed in the example below is the DataCite mapping, I think (or the MODS mapping). For this one it would have been "It has specific versions or instances, one of which is this related work" instead of "has version."

Image

justinlittman:
closes #665

image

### DIFF
--- a/app/views/works/show.html.erb
+++ b/app/views/works/show.html.erb
@@ -60,11 +60,7 @@
   <% component.with_row(label: 'Deposit subtypes', values: [@work_presenter.all_work_subtypes]) %>
 <% end %>
 
-<%= render Elements::Tables::TableComponent.new(id: 'related-content-table', classes: 'mb-5') do |component| %>
-  <% component.with_caption do %><%= render Show::TableHeadingComponent.new(text: 'Related content', presenter: @work_presenter, tab: :related_content) %><% end %>
-  <% component.with_row(label: 'Related published work', values: @work_presenter.related_works) %>
-  <% component.with_row(label: 'Related links', values: @work_presenter.related_links) %>
-<% end %>
+<%= render Works::Show::RelatedContentShowComponent.new(work_presenter: @work_presenter) %>
 
 <%= render Elements::Tables::TableComponent.new(id: 'access-table', classes: 'mb-5') do |component| %>
   <% component.with_caption do %><%= render Show::TableHeadingComponent.new(text: 'Access settings', presenter: @work_presenter, tab: :access) %><% end %>
--- a/app/components/works/show/related_content_show_component.html.erb
+++ b/app/components/works/show/related_content_show_component.html.erb
@@ -1,0 +1,18 @@
+<%= render Elements::Tables::TableComponent.new(id: 'related-content-table', classes: 'mb-5') do |component| %>
+  <% component.with_caption do %><%= render Show::TableHeadingComponent.new(text: 'Related content', presenter: work_presenter, tab: :related_content) %><% end %>
+  <% related_works.each do |related_work| %>
+    <% component.with_row(label: 'Related published work') do |row_component| %>
+      <% row_component.with_cell do %>
+        <p class="my-0">
+          <%= related_work %>
+        </p>
+        <% if related_work.relationship %>
+          <p class="my-0">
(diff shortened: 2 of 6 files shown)


app/components/works/show/related_contents_show_component.rb
```diff
@@ -0,0 +1,25 @@
...
+        @work_presenter = work_presenter
+        super()
+      end
+
+      attr_reader :work_presenter
+
+      delegate :related_links, to: :work_presenter
+
+      def related_works
+        work_presenter.related_works_attributes.select { |related_work| related_work.to_s.present? }
```
mjgiarlo, line 17: If you use `work_presenter.related_works` here, I think you can drop the explicit `to_s` in the block? Worth a quick try.

mjgiarlo, line 17: If you use `work_presenter.related_works` here, I think you might be able to drop the explicit `to_s` in the block. Like:

```suggestion
        work_presenter.related_works.select(&:present?)
```

Worth a quick try?

app/components/works/show/related_contents_show_component.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def relationship_label_for(related_work)
+        t("related_works.edit.fields.relationship.options.#{related_work.relationship}")
```
mjgiarlo, line 21: Will all values of `related_work.relationship` result in a translation being found in our locales? If not, the return value will be a missing translation string. If that's an issue, we can consider:

```suggestion
        t("related_works.edit.fields.relationship.options.#{related_work.relationship}", default: nil)
```

Which will get us a `nil` relationship label.

app/views/works/show.html.erb
```diff
@@ -60,11 +60,7 @@
   <% component.with_row(label: 'Deposit subtypes', values: [@work_presenter.all_work_subtypes]) %>
 <% end %>
 
-<%= render Elements::Tables::TableComponent.new(id: 'related-content-table', classes: 'mb-5') do |component| %>
-  <% component.with_caption do %><%= render Show::TableHeadingComponent.new(text: 'Related content', presenter: @work_presenter, tab: :related_content) %><% end %>
-  <% component.with_row(label: 'Related published work', values: @work_presenter.related_works) %>
-  <% component.with_row(label: 'Related links', values: @work_presenter.related_links) %>
-<% end %>
+<%= render Works::Show::RelatedContentsShowComponent.new(work_presenter: @work_presenter) %>
```
mjgiarlo, line 63: Trivial but:

```suggestion
<%= render Works::Show::RelatedContentShowComponent.new(work_presenter: @work_presenter) %>
```

(I think we tend to refer to this as related content not related contents. 🤷🏻‍♂️)

justinlittman: Makes sense. Fixing.

app/components/works/show/related_contents_show_component.rb
justinlittman: I messed around with this a bit when I was writing the code, but couldn't get it to work. Just verified with your suggestion:
image

justinlittman: Nevermind, I didn't read your suggestion closely. It works. Fixing.

app/components/works/show/related_contents_show_component.rb
justinlittman: They all have translations.

app/components/works/show/related_contents_show_component.rb
justinlittman: No, I was wrong, it doesn't. `.related_works` returns an array of strings.

mjgiarlo: And what you want here is the instance, got it.

mjgiarlo (approved):

PR #912, opened 2025-03-03
Add ability to search for depositor SUNet IDs as admin function

### LINKED ISSUE #750: Admin function: retrieve SUNet IDs for a list of druids
Admin should be able to input a list of druids and have returned two columns: the druids entered with the SUNet ID of the depositor for each of those works.

This can be either a table displayed onscreen or a csv that can be downloaded, similar to the item and collection reports.

mjgiarlo:
Fixes #750

Peek 2025-03-04 13-49

### DIFF
--- a/app/controllers/admin/depositors_search_controller.rb
+++ b/app/controllers/admin/depositors_search_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller to allow searching for depositor SUNet IDs given druids
+  class DepositorsSearchController < Admin::ApplicationController
+    def new
+      authorize!
+
+      @depositors_search_form = Admin::DepositorsSearchForm.new
+
+      render :search
+    end
+
+    def search
+      authorize!
+
+      @depositors_search_form = Admin::DepositorsSearchForm.new(**depositors_search_params)
+
+      if @depositors_search_form.valid?
+        @druid_to_sunetid_map = Admin::DepositorsSearch.call(form: @depositors_search_form)
+      else
+        render :search, status: :unprocessable_entity
+      end
+    end
+
+    private
+
+    def depositors_search_params
+      params.expect(admin_depositors_search: Admin::DepositorsSearchForm.user_editable_attributes)
+    end
+  end
+end
--- a/app/forms/admin/depositors_search_form.rb
+++ b/app/forms/admin/depositors_search_form.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Admin
+  # Admin form object for searching for depositor information
+  class DepositorsSearchForm < ApplicationForm
+    before_validation :normalize_druids
+    attribute :druids, :string
+    validate :works_present
+
+    def druid_list
+      druids.split(/\s+/)
+    end
+
+    def normalize_druids
+      self.druids = druid_list.uniq.filter_map do |druid|
+        next if druid.blank?
+
+        druid.starts_with?('druid:') ? druid : druid.prepend('druid:')
+      end.join(' ')
+    end
+
+    def works_present
+      druid_list.each do |druid|
+        next if Work.exists?(druid:)
+
+        errors.add(:druids, "work not found: #{druid}")
+      end
+    end
+  end
+end
--- a/app/services/admin/depositors_search.rb
+++ b/app/services/admin/depositors_search.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Admin
+  # Return a list of depositors based on druids
+  class DepositorsSearch
+    def self.call(...)
+      new(...).call
+    end
+
+    # @param [Admin::DepositorsSearchForm] form an instance of the depositors search form
+    def initialize(form:)
+      @druid_list = form.druid_list
+    end
+
+    # @return [Array<Hash{Symbol => String, Array<String>}>] an array of key-value
+    #   pairs corresponding to work DRUIDs and work depositor SUNet IDs, a data
+    #   structure used by, e.g., the `with_rows` slot of the `Elements::Tables::TableComponent`
+    def call
+      selected_works.map do |work|
+        {
+          id: work.druid,
+          values: [work.druid, work.email_address.delete_suffix(User::EMAIL_SUFFIX)]
+        }
+      end
+    end
+
+    private
+
+    attr_reader :druid_list
+
+    def selected_works
+      Work
+        .eager_load(:user)
+        .where(druid: druid_list)
+        .select('works.druid, works.user_id, users.email_address')
+    end
+  end
+end
--- a/app/views/admin/depositors_search/search.html.erb
+++ b/app/views/admin/depositors_search/search.html.erb
@@ -1,0 +1,26 @@
+<% content_for :breadcrumbs do %>
+  <%= render Elements::BreadcrumbNavComponent.new do |component| %>
+    <% component.with_breadcrumb(text: 'Dashboard', link: dashboard_path) %>
+    <% component.with_breadcrumb(text: 'Admin dashboard', link: admin_dashboard_path) %>
+    <% component.with_breadcrumb(text: 'Get SUNet IDs of depositors') %>
+  <% end %>
+<% end %>
+
+<div class="d-flex justify-content-end">
+  <%= render Admin::FunctionDropdownComponent.new %>
+</div>
+
+<%= render Elements::HeadingComponent.new(level: :h1, text: 'Get SUNet IDs of depositors', classes: 'h2 mb-4') %>
+
+<%= form_with model: @depositors_search_form, url: search_admin_depositors_search_path, method: :get, class: 'mb-5' do |form| %>
+  <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :druids, label: 'Enter work druids separated by spaces', hidden_label: true) %>
+  <%= render Elements::Forms::SubmitComponent.new(label: 'Submit', classes: 'mt-4') %>
+<% end %>
+
+<% if @druid_to_sunetid_map.present? %>
+  <%= render Elements::HorizontalRuleComponent.new %>
+  <%= render Elements::Tables::TableComponent.new(id: 'depositors-table', classes: 'mb-5', label: 'Depositor SUNet IDs', show_label: false) do |component| %>
+    <% component.with_header(headers: ['Work DRUID', 'Depositor SUNet ID']) %>
+    <% component.with_rows(@druid_to_sunetid_map) %>
+  <% end %>
+<% end %>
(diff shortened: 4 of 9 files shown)


app/controllers/admin/depositors_search_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller to allow searching for depositor SUNet IDs given druids
+  class DepositorsSearchController < Admin::ApplicationController
```
mjgiarlo, line 5: My first challenge when thinking about adding an admin function is how best to set it up, in terms of routes and controller actions, in line with the sort of RESTful resource-oriented architecture that Rails makes so easy. I'm not sure I like what I decided upon here. Do our current admin functions follow any set of practices in this area? If so, I'm all ears. (I could just pick a random one and cargo-cult my way to implementation, but I think it's more important that we developers **grok** how we're mapping diff admin functions in H3.)

app/controllers/admin/depositors_search_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      authorize!
+
+      @depositors_search_form = Admin::DepositorsSearchForm.new(**depositors_search_params)
+
+      if @depositors_search_form.valid?
+        @druid_to_depositor_sunetid_map = Admin::DepositorsSearch.call(form: @depositors_search_form)
+        render :index, status: :created
+      else
+        render :new, status: :unprocessable_entity
+      end
```
mjgiarlo, line 16: Is it weird for `render :index` to need `status: :created` (instead of defaulting to `status: :ok`)? Because I found that this was needed, probably to escape some underlying behavior. Maybe turbo-related? I'm not sure at the moment.

app/controllers/admin/depositors_search_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    def index
+      authorize!
+
+      @depositors_search_form = Admin::DepositorsSearchForm.new(**depositors_search_params)
+
+      if @depositors_search_form.valid?
+        @druid_to_depositor_sunetid_map = Admin::DepositorsSearch.call(form: @depositors_search_form)
+        render :index, status: :created
+      else
+        render :new, status: :unprocessable_entity
```
mjgiarlo, line 15: Is it weird for `#index` to render the `:new` view?

app/views/admin/depositors_search/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<div class="d-flex justify-content-end">
+  <%= render Admin::FunctionDropdownComponent.new %>
+</div>
+
+<%= render Elements::HeadingComponent.new(level: :h1, text: 'Get SUNet IDs of depositors', classes: 'h2 mb-4') %>
+
+<%= form_with model: @depositors_search_form, url:  admin_depositors_search_index_path, class: 'mb-5' do |form| %>
+  <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :druids, label: 'Enter work druids separated by spaces', hidden_label: true) %>
+  <%= render Elements::Forms::SubmitComponent.new(label: 'Submit', classes: 'mt-4') %>
+<% end %>
```
mjgiarlo, line 18: This is exactly the same as the new template, which is gross. Good evidence I haven't arrived at the best technical design for this feature.

config/routes.rb
```diff
@@ -75,6 +75,10 @@
   namespace :admin do
     get 'dashboard', to: 'dashboard#show'
 
+    resources :depositors_search, only: %i[new] do
+      collection { post :index }
+    end
```
mjgiarlo, line 80: I'm all ears. Should it maybe have a single new endpoint that can take GETs (to render the initial search form) and POSTs (to render the filled-out form + any results)? I'm open to ideas.

app/forms/admin/depositors_search_form.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Admin
+  # Admin form object for searching for depositor information
+  class DepositorsSearchForm < ApplicationForm
+    before_validation :normalize_druids
+    attribute :druids, :string
+    validate :works_present
```
aaron-collier, line 8: It feels strange to me to invalidate the form if a druid isn't found, wouldn't that invalidate the whole process? Perhaps a way to list druids that aren't found so that individual druids in a long list do not invalidate the entire form. I could be misunderstanding what's happening here though.

config/routes.rb
aaron-collier: Is there a reason to post to `:index` and not `:create`?

mjgiarlo: My logic, which I am struggling with, was that the POST doesn't create anything so `create` feels odd. It provides a filtered list of works and their depositors' SUNet IDs.

app/forms/admin/depositors_search_form.rb
mjgiarlo: It's perhaps (❓) surprising how this set of deltas in this branch leads to the following behavior which I *think* may be what we want:

Peek 2025-03-03 16-45

app/controllers/admin/depositors_search_controller.rb
justinlittman: Yeah, this is tricky. My general philosophy has been to follow Rails conventions as long as it seemed like a natural sense and would make sense to other Rails devs.

Here's an example of another search controller: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/controllers/admin/druid_search_controller.rb

My thought is that `new` is appropriate for rendering a search form but none of the other Rails routes are a good fit for performing the search.

app/forms/admin/depositors_search_form.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Admin
+  # Admin form object for searching for depositor information
+  class DepositorsSearchForm < ApplicationForm
+    before_validation :normalize_druids
+    attribute :druids, :string
```
justinlittman, line 7: Hmm ... I'm not sure about this being passed as a string but then transformed to an array. Perhaps they should be separate attributes (where the array attribute has `array: true`).

app/forms/admin/depositors_search_form.rb
```diff
@@ -0,0 +1,30 @@
...
+      self.druids = druid_list.filter_map do |druid|
+        next if druid.blank?
+
+        druid.starts_with?('druid:') ? druid : druid.prepend('druid:')
+      end.join(' ')
+    end
+
+    def works_present
+      druid_list.each do |druid|
+        next if Work.find_by(druid:)
```
justinlittman, line 24: `exists?` is faster since don't load record.

app/services/admin/depositors_search.rb
```diff
@@ -0,0 +1,29 @@
...
+      new(...).call
+    end
+
+    def initialize(form:)
+      @form = form
+    end
+
+    def call
+      form.druid_list.map do |druid|
+        work = Work.find_by!(druid:)
```
justinlittman, line 16: There's probably a faster query that only retrieves the email address.

Also, might be able to do this as a single query instead of a query for each work.

app/views/admin/depositors_search/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<% content_for :breadcrumbs do %>
```
justinlittman, line 1: Can new and index be combined into a single form?

config/routes.rb
justinlittman: Take a look at routes for collection_search and druid_search?

app/forms/admin/depositors_search_form.rb
mjgiarlo: @justinlittman I'd started with this as an array, but I've forgotten why I switched. Maybe because the supporting field in the HTML is a `textarea`? I'll take another hack at it. I prefer thinking of `druids` as an array, too.

app/views/admin/depositors_search/new.html.erb
mjgiarlo: That's what I'll try to do.

app/services/admin/depositors_search.rb
mjgiarlo: I changed this so that it issues two queries total. Maybe worth another look once I push up changes.

app/forms/admin/depositors_search_form.rb
mjgiarlo: I'm going to skip this data type change for now, since the textarea component wants its field to be string-y.

app/controllers/admin/depositors_search_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def search
+      authorize!
+
+      @depositors_search_form = Admin::DepositorsSearchForm.new(**depositors_search_params)
+
+      if @depositors_search_form.valid?
+        render turbo_stream: [
+          turbo_stream.update('depositors-search-form',
```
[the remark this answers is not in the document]
mjgiarlo: Necessary? No. I'll yank them out.

app/services/admin/depositors_search.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Admin
+  # Return a list of depositors based on druids
+  class DepositorsSearch
+    def self.call(...)
+      new(...).call
+    end
+
+    def initialize(form:)
```
justinlittman, line 10: Documenting the parameter would be helpful.

app/views/admin/depositors_search/search.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% end %>
+
+<div class="d-flex justify-content-end">
+  <%= render Admin::FunctionDropdownComponent.new %>
+</div>
+
+<%= render Elements::HeadingComponent.new(level: :h1, text: 'Get SUNet IDs of depositors', classes: 'h2 mb-4') %>
+
+<%= turbo_frame_tag 'depositors-search-form' do %>
+  <%= form_with model: @depositors_search_form, url: search_admin_depositors_search_path, method: :get, class: 'mb-5', data: { turbo_frame: 'depositors-search' } do |form| %>
```
justinlittman, line 16: Is this turbo frame necessary?

mjgiarlo: I'd removed it locally and failed to push. So no!

justinlittman (approved):
It would great to also have a system test that proves that everything works together.

mjgiarlo:
@justinlittman 💬 
> It would great to also have a system test that proves that everything works together.

At the moment, there are few moving parts in this PR and no turbo or JS, so I'm inclined to move on without system tests for the time being. I'm open to adding one or more later if some of the interactions here are falling in test gaps.

PR #918, opened 2025-03-04
Implement admin delete collection function

### LINKED ISSUE #729: Ability for admin to delete a collection
Add the ability for an admin to delete a collection from H3. This is a replacement for our current collection decommission function that was more complex.

We should likely verify that there are no items in the collection in order to do this.

aaron-collier:
Fixes #729 

Delete with an empty collection:
https://github.com/user-attachments/assets/1b2c1897-18ab-4e42-ad2d-b33a4aeca48a

Delete with a collection with works showing failure:
https://github.com/user-attachments/assets/6dba427d-db62-47a2-bf03-15d937549376

### DIFF
--- a/app/controllers/admin/delete_collection_controller.rb
+++ b/app/controllers/admin/delete_collection_controller.rb
@@ -12,8 +12,12 @@ module Admin
     def destroy
       authorize!
 
-      collection.destroy!
-      render_delete_success
+      if collection.works.empty?
+        collection.destroy!
+        render_delete_success
+      else
+        render_delete_failure
+      end
     end
 
     attr_reader :collection_form
@@ -30,6 +34,16 @@ module Admin
       respond_to do |format|
         format.turbo_stream do
           render turbo_stream: turbo_stream.action(:full_page_redirect, dashboard_path)
+        end
+      end
+    end
+
+    def render_delete_failure
+      flash[:danger] = I18n.t('messages.collection_delete_failed')
+      # This breaks out of the turbo frame.
+      respond_to do |format|
+        format.turbo_stream do
+          render turbo_stream: turbo_stream.action(:full_page_redirect, collection_path(collection.druid))
         end
       end
     end
--- a/app/controllers/admin/delete_collection_controller.rb
+++ b/app/controllers/admin/delete_collection_controller.rb
@@ -3,6 +3,8 @@
 module Admin
   # Controller for deleting a Collection as an administrator
   class DeleteCollectionController < Admin::ApplicationController
+    before_action :set_collection
+
     def new
       authorize!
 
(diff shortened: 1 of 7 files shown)


app/controllers/admin/delete_collection_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      flash[:success] = I18n.t('messages.collection_deleted')
+      # This breaks out of the turbo frame.
+      respond_to do |format|
+        format.turbo_stream do
+          render turbo_stream: turbo_stream.action(:full_page_redirect, dashboard_path)
+        end
+      end
+    end
+
+    def render_delete_failure
```
justinlittman, line 41: Seems like this should report a validation error instead of breaking out of the frame and notifying with a flash.

justinlittman (changes requested):

app/controllers/admin/delete_collection_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      collection.destroy!
+      render_delete_success
+    end
+
+    attr_reader :collection_form, :collection
+
+    private
+
+    def druid
+      params[:druid] || params[:collection_druid]
```
justinlittman, line 26: Are you sure you need both of these?

app/views/admin/delete_collection/form.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <%= render Elements::CardComponent.new(classes: 'admin-card') do |component| %>
+      <% component.with_body do %>
+        <%= form_with url: collection_admin_delete_path(druid: params[:collection_druid]), method: :delete do |form| %>
+          <%= render Elements::Forms::CheckboxComponent.new(form:, field_name: :confirm_decommissioned, label: 'I confirm this collection has been decommissioned in Argo and understand that this action cannot be undone.', required: true) %>
+          <%= render Elements::Forms::SubmitComponent.new(label: 'Delete', classes: 'mt-3') %>
+        <% end %>
+      <% end %>
+    <% end %>
+  <% else %>
+    <%= render Elements::AlertComponent.new(variant: :danger, value: t('messages.collection_not_empty')) %>
```
justinlittman, line 12: Doesn't feel like this should be a danger.

justinlittman (approved):

app/controllers/admin/delete_collection_controller.rb
aaron-collier: Nah - druid was coming from the route, updated that to collection_druid so only need the one now. :+1+

app/views/admin/delete_collection/form.html.erb
aaron-collier: Cool, changed to :warning

PR #920, opened 2025-03-04
Do not link shareable links when they are not ready.

### LINKED ISSUE #781: Do not make purl link clickable in H3 until item is deposited
Currently this is created and displayed as a live link, even when saving the first draft. This is confusing as the link doesn't work yet. I think we should display the link but have it be inactive.

Ditto for display on the deposits page for the collection.

Dashboard:

Image

Deposits page:

Image

PURL at this stage:

Image

justinlittman:
closes #781

### DIFF
--- a/app/components/collections/show/deposits_component.html.erb
+++ b/app/components/collections/show/deposits_component.html.erb
@@ -1,4 +1,4 @@
 <%= render Elements::Tables::TableComponent.new(id: 'deposits-table', classes: 'mb-5', label: 'Deposits for this collection', show_label: false) do |component| %>
-  <% component.with_header(headers: ['Deposit', 'Owner', 'Status', 'Modified', 'Link for sharing']) %>
+  <% component.with_header(headers: ['Deposit', 'Owner', 'Status', 'Modified', t('sharing_link.label')], each_tooltips: [nil, nil, nil, nil, t('sharing_link.tooltip_html')]) %>
   <% component.with_rows(collection_deposits) %>
 <% end %>
--- a/app/components/collections/show/deposits_component.rb
+++ b/app/components/collections/show/deposits_component.rb
@@ -22,31 +22,22 @@ module Collections
       end
 
       def values_for(work)
+        presenter = work_presenter_for(work)
         [
           helpers.link_to(work.title, work_or_wait_path(work)),
           work.user.name,
-          status_for(work),
+          presenter.status_message,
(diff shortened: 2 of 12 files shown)


app/components/collections/show/deposits_component.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= render Elements::Tables::TableComponent.new(id: 'deposits-table', classes: 'mb-5', label: 'Deposits for this collection', show_label: false) do |component| %>
-  <% component.with_header(headers: ['Deposit', 'Owner', 'Status', 'Modified', 'Link for sharing']) %>
+  <% component.with_header(headers: ['Deposit', 'Owner', 'Status', 'Modified', t('sharing_link.label')], each_tooltips: [nil, nil, nil, nil, t('sharing_link.tooltip_html')]) %>
```
mjgiarlo, line 2: Beginning to wonder if we should consider table headers as their own objects (components, structs, hashes) that must have a label and may have a tooltip and one or more classes. But that's more food for thought for follow-up tickets than feedback on this PR.

app/components/collections/show/deposits_component.rb
```diff
@@ -22,31 +22,22 @@ def collection_deposits
...
       def values_for(work)
+        presenter = work_presenter_for(work)
         [
           helpers.link_to(work.title, work_or_wait_path(work)),
           work.user.name,
-          status_for(work),
+          presenter.status_message,
           work.object_updated_at ? helpers.l(work.object_updated_at, format: '%b %d, %Y') : nil,
-          persistent_link_for(work)
+          presenter.sharing_link
```
mjgiarlo, line 31: I appreciate moving more of this logic into the presenter objects once we discover it's not component-specific.

app/presenters/work_presenter.rb
```diff
@@ -32,10 +48,6 @@ def keywords
     keywords_attributes.map(&:text).join(', ')
   end
 
-  def related_works
-    related_works_attributes.map(&:to_s)
-  end
-
```
mjgiarlo, line 38: I was a bit surprised to see this since it's not obviously link-related. Am I wrong about that? Or is this handled elsewhere now?

mjgiarlo (approved):
Approve pending consideration of comments

app/presenters/work_presenter.rb
justinlittman: I'm surprised too. However, it isn't used anywhere, so maybe I was just removing cruft?

PR #927, opened 2025-03-04
Avoid and handle blank work subtypes.

### LINKED ISSUE #821: Remove trailing comma from list of work subtypes
Can we get rid of the comma at the end of this list on the item view page? See https://sul-h3-stage.stanford.edu/works/druid:rg185fj5376

Image

justinlittman:
closes #821

### DIFF
--- a/app/services/to_cocina/work/types_mapper.rb
+++ b/app/services/to_cocina/work/types_mapper.rb
@@ -28,7 +28,7 @@ module ToCocina
       delegate :work_type, to: :work_form
 
       def work_subtypes
-        @work_subtypes ||= work_form.work_subtypes + [work_form.other_work_subtype].compact
+        @work_subtypes ||= work_form.work_subtypes + [work_form.other_work_subtype.presence].compact
       end
 
       def self_deposit_resource_types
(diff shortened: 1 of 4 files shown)


app/services/to_cocina/work/types_mapper.rb
```diff
@@ -28,7 +28,7 @@ def call
       delegate :work_type, to: :work_form
 
       def work_subtypes
-        @work_subtypes ||= work_form.work_subtypes + [work_form.other_work_subtype].compact
+        @work_subtypes ||= work_form.work_subtypes + [work_form.other_work_subtype.presence].compact
```
mjgiarlo, line 31: Not blocking, just a question: do you think this logic could live in the form itself or does it make more sense to stay here in the mapper?

mjgiarlo (approved):
Approve pending comment consideration

app/services/to_cocina/work/types_mapper.rb
justinlittman: I could go either way, but slightly feel that blank is the correct value for the form (instead of nil).

PR #942, opened 2025-03-04
Do not use invalid license for collection in cocina

### LINKED ISSUE #939: Fix no-license bug on collection deposit
https://app.honeybadger.io/projects/128495/faults/117973801

aaron-collier:
Fixes #939

### DIFF
--- a/app/services/to_cocina/collection/mapper.rb
+++ b/app/services/to_cocina/collection/mapper.rb
@@ -50,6 +50,7 @@ module ToCocina
       def access
         { view: 'world' }.tap do |params|
           next unless (collection_license = collection_form.selected_license)
+          next if collection_license == License::NO_LICENSE_ID
 
           params[:license] = collection_license
         end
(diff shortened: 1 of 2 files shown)


justinlittman (changes requested):
That value for none licenses is deprecated. See https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L1319

aaron-collier:
@justinlittman ah, well that's annoying. Can't say I'm a big fan of that... seems to require scattered logic that leads to bugs like this... but oh well. I'll update.

justinlittman:
It also makes it unclear whether there is a positive assertion that this item has no license and when the license is just missing or unknown.

justinlittman:
So I think this will fix the problem. However:
1. Needs a test for `ToCocina::Collection::Mapper` to verify that fixes.
2. `.selected_license` is only used in ``ToCocina::Collection::Mapper`. I'd recommend taking the logic out of the form and put it in the mapper.

justinlittman (approved):

PR #953, opened 2025-03-05
Changes related works to radio buttons.

### LINKED ISSUE #667: Related works behavior should match designs
The design for this shows that when the user ticks the checkbox to provide the full citation, the top box for the link is deactivated. However, it is not working that way. 
https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1116-5013&t=OJWk8ENl8jskT5Xe-4

Image

In this case, I entered content in both the link AND the text citation option fields. I initially had content in the top field. I then ticked the box and entered content in the full citation text field without removing the other text.

The validation error for that looks like this:

Image

Which makes me think that I have to leave both fields blank, which is not correct. I think this would not be a problem if the behavior matched the designs.

justinlittman:
closes #667

https://github.com/user-attachments/assets/a4bb79e7-6205-41d2-8d92-7c47de374c33


app/components/works/edit/citation_component.html.erb
```diff
@@ -1,15 +1,14 @@
-<%= tag.div data: { controller: 'citation' }, class: 'pane-section' do %>
-  <%= render Elements::Forms::FieldsetComponent.new(label: t('works.edit.fields.citation.label')) do %>
+<%= tag.div class: 'pane-section' do %>
+  <%= render Elements::Forms::FieldsetComponent.new(label: t('works.edit.fields.citation.label'), data: { controller: 'radio' }) do %>
```
justinlittman, line 2: Replaced citation stimulus controller with new radio controller.

aaron-collier (approved):

PR #961, opened 2025-03-05
[SPIKE] Ditch support class to explore whether we prefer inherited methods over method buckets

mjgiarlo:
A quick and harmless spike for @aaron-collier @thatbudakguy and @justinlittman to take a gander at. Which pattern feels better?


app/components/application_component.rb
```diff
@@ -17,4 +17,27 @@ def merge_classes(*)
   def merge_actions(*)
```
mjgiarlo, line 17: I like that we didn't plop `#merge_classes` and `#merge_actions` into a `MergeSupport` class so I thought it worth exploring parity here.

app/views/admin/druid_search/new.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="row align-items-start">
       <%= render Elements::Forms::LabelComponent.new(form:, field_name: :druid, label_text: 'Search for DRUID', classes: 'col-md-2') %>
       <div class="col-md-6">
-        <%= form.text_field :druid, class: 'form-control', aria: Elements::Forms::InvalidFeedbackSupport.arias_for(form:, field_name: :druid) %>
+        <%= form.text_field :druid, class: 'form-control', aria: invalid_feedback_arias_for(form:, field_name: :druid) %>
```
mjgiarlo, line 6: Ah, nevermind! This isn't going to work here because this is a view, not a component. 😆  Walking away from the spike.

justinlittman: And that is what led me to extract it in the first place ...

PR #962, opened 2025-03-05
Record participant names when creating users.

justinlittman:


app/controllers/accounts_controller.rb
```diff
@@ -15,7 +15,7 @@ def search
 
   # Does a lookup from the account service in production mode, otherwise examines local database for users
   def lookup_account
-    return AccountService.call(sunetid: params[:id]) if Rails.env.production?
+    return AccountService.call(sunetid: params[:id]) unless Rails.env.development?
```
justinlittman, line 18: Going to explicitly mock out the Account Service in tests, so the fake Account Service is only needed in development.

app/forms/depositor_form.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
 # Form for collection depositors
-class DepositorForm < ApplicationForm
-  attribute :sunetid, :string
+class DepositorForm < ParticipantForm
```
justinlittman, line 4: Since there is more than one attribute, having all participant forms subclass a common form.

app/forms/participant_form.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Form for collection participants
+class ParticipantForm < ApplicationForm
+  attribute :sunetid, :string
+  attribute :name, :string
```
justinlittman, line 6: `name` is a new attribute.

app/jobs/deposit_collection_job.rb
```diff
@@ -75,9 +75,8 @@ def assign_participants(role)
       participant = participant.attributes if participant.respond_to?(:attributes)
       next if participant['sunetid'].blank?
 
-      user = User.find_or_initialize_by(email_address: sunetid_to_email_address(participant['sunetid']))
-      user.update!(name: participant['sunetid']) if user.name.blank?
-      user.save!
+      user = User.create_with(name: participant['name'] || participant['sunetid'])
```
justinlittman, line 78: Now populating the name when creating a User.

app/services/to_collection_form/mapper.rb
```diff
@@ -45,12 +45,15 @@ def params
...
-    # @param [Symbol] role :managers or :depositors
+    # @param [Symbol] role :managers, :reviewers, or :depositors
     # @return [Array<Hash>] an array of participant attributes
     # Creates the attribute hash of sunetids for the given role to be used in the CollectionForm
     def participant_attributes(role)
       collection.send(role).map do |participant|
-        { sunetid: participant.sunetid }
+        {
+          sunetid: participant.sunetid,
+          name: participant.name
```
justinlittman, line 55: Now populating the form with name too.

mjgiarlo (approved):

PR #963, opened 2025-03-05
[SPIKE] Pass in table headers as objects in lieu of relying on parallel arrays

mjgiarlo:


app/components/collections/show/deposits_component.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= render Elements::Tables::TableComponent.new(id: 'deposits-table', classes: 'mb-5', label: 'Deposits for this collection', show_label: false) do |component| %>
-  <% component.with_header(headers: ['Deposit', 'Owner', 'Status', 'Modified', t('sharing_link.label')], each_tooltips: [nil, nil, nil, nil, t('sharing_link.tooltip_html')]) %>
+  <% component.with_header(headers:) %>
```
justinlittman, line 2: Can these be components/slots instead of a model?

PR #969, opened 2025-03-06
[SPIKE] Refactor components for table head.

justinlittman:


app/components/admin/collections_list_component.html.erb
```diff
@@ -1,9 +1,9 @@
 <%= render Elements::Tables::TableComponent.new(
       id:, label:, empty_message:
     ) do |component| %>
-  <% component.with_header(
-       headers: %w[Collection Druid Roles]
-     ) %>
+  <% component.with_header(label: 'Collection') %>
+  <% component.with_header(label: 'Druid') %>
+  <% component.with_header(label: 'Roles') %>
```
mjgiarlo, line 6: How would you feel about a small tweak to this PR, changing `with_header` to `with_headers` canonically?

```suggestion
  <% component.with_headers([
       { label: 'Collection' },
       { label: 'Druid' },
       { label: 'Roles' }
  ]) %>
```

justinlittman: I've never loved that that pattern, but it definitely is idiomatic view component. Which is to say, I would not be opposed.

mjgiarlo: In this case, I tend to think of the headers as a set of things, not as individual things and using the collection slot pattern reinforces that way of thinking about them. Either way, I'm popping an approval on this. Thanks for spiking on it!

mjgiarlo (approved):

app/components/admin/collections_list_component.html.erb
justinlittman: Change made everywhere.

PR #970, opened 2025-03-06
Change participants list on collection view page.

### LINKED ISSUE #834: UseSUNets to list participants
We are now using actual names across the UI everywhere but for participants on the view page, where we are listing emails. 

What should we display here? 

- Whatever was entered by the user (email or SUNet or whatever)
- Whatever was entered by the user (email or SUNet or whatever) + the actual name
- Just the actual name

@astridu do you have thoughts on this?

justinlittman:
closes #834

image

### DIFF
--- a/app/components/elements/tables/list_cell_component.rb
+++ b/app/components/elements/tables/list_cell_component.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Elements
+  module Tables
+    # Component for rendering a list in a table cell.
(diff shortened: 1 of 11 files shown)


app/components/elements/tables/list_cell_component.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
```
justinlittman, line 1: The ListCellComponent was split out of the RowComponent.

aaron-collier (approved):

PR #980, opened 2025-03-06
Standardize on druid as param for collection and work paths

### LINKED ISSUE #958: Fix link in invitation to deposit email
The link in email that invites users to deposit in a collection (app/views/collections_mailer/invitation_to_deposit_email.html.erb) has a bad link in it. It says "Start your deposit at [https://sul-h3-stage.stanford.edu/collections/173](https://sul-h3-stage.stanford.edu/collections/173#deposits)." This produces a 404 error. I think we should be using links with druids?

It would be good to make sure this isn't a problem in any of the other emails either.

mjgiarlo:
Fixes #958

It remains possible to link directly to a collection or work database ID, which we do in a handful of situations where we want to go directly to the "wait" page. Instead of special-casing the druid paths/URLs which are the norm in H3, switch the behavior to special-case the oddballs.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -55,4 +55,8 @@ class Collection < ApplicationRecord
   def first_version?
     version == 1
   end
+
+  def to_param
+    druid
+  end
 end
--- a/app/components/dashboard/show/drafts_list_component.rb
+++ b/app/components/dashboard/show/drafts_list_component.rb
@@ -22,7 +22,7 @@ module Dashboard
       def values_for(work)
         [
           link_to(work.title, work_or_wait_path(work)),
-          link_to(work.collection.title, collection_path(druid: work.collection.druid)),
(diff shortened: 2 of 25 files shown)


app/models/collection.rb
```diff
@@ -55,4 +55,8 @@ def reviewers_and_managers
   def first_version?
     version == 1
   end
+
+  def to_param
+    druid
+  end
```
mjgiarlo, line 61: The changes to the models in this commit are the key change that gets us back to being able to be more Rails-y in our path and URL helpers, e.g., `collection_path(collection)`, without losing the ability to craft special-case URLs such as the explicit "wait" paths, in which case we want `wait_collections_path(collection.id)`.

justinlittman (approved):

justinlittman:
Love it.

PR #981, opened 2025-03-06
Adds bulk entry of participants.

### LINKED ISSUE #906: Allow for bulk participant entry
Figma: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=789-4316&t=mTnNx6Dknc3uoVn1-4

This should be implemented for collection manager and depositor entries on this page.

- User enters a list in a text area, **either one entry per line (new-line delimited) OR a comma delimited list**, of Stanford email addresses and/or SUNet IDs.
-- Users would like to be able to enter email addresses that use Stanford aliases, but it is yet to be determined if we can look these up or not.
- User clicks the "Add managers" or "Add depositors" button
-- If a SUNet ID is found, the entry is removed from the list in the text area and the SUNet ID with the name of the person are displayed with a trash can icon below the text area box.
-- If we aren't able to identify someone from the entry, we leave the entry in the text area and include an error message below the text entry box indicating which entries have failed and that they do not appear to be valid SUNet IDs
-- If we are unable to lookup aliases, these will fail validation (like they do in H2)
-- The text area should be cleared if everything succeeds.
- User can then correct entries left in the box or add other entries to the box and click the button again to add more participants.

Possible issues to evaluate whether we need to address now or keep an eye on for later:
- We may need to wait and add the alias lookup later.
- Lookup for a big list of people may be slow. We may need to add a UI indicator/spinner/etc or limiting the number you can enter at one time
- The list of participants may also get really long. We may want to put these in a scrolly box. Pagination would be complex to do.
- Currently, newly added depositors are just appended to the end, and so it’s ordered by adds (not alphabetically); newest at end. Users have never said anything about this; presumably it’s fine?

justinlittman:
closes #906

https://github.com/user-attachments/assets/337045b8-bfc6-4c5a-b211-2cccc7d18633

### DIFF
--- a/app/components/collections/edit/participant_component.rb
+++ b/app/components/collections/edit/participant_component.rb
@@ -15,14 +15,6 @@ module Collections
 
       delegate :object, to: :form
 
-      def lookup_id
-        form.field_id(:lookup)
-      end
-
-      def label
-        'SUNet ID'
-      end
-
       def value
         return unless sunetid
 
--- a/app/javascript/controllers/participants_controller.js
+++ b/app/javascript/controllers/participants_controller.js
@@ -1,0 +1,53 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['input', 'invalidFeedbackContainer', 'invalidFeedbackTemplate']
+
+  static values = { url: String }
+
+  static outlets = ['nested-form']
+
+  async lookup (event) {
+    event.preventDefault()
+
+    if (this.inputTarget.value === '') return
+
+    this.invalidFeedbackContainerTarget.innerHTML = ''
+    const ids = this.inputTarget.value.split(/[,;]? |\n/)
+    const errorIds = await Promise.all(ids.map(id => this.lookupId(id)))
+    this.inputTarget.value = errorIds.filter(id => id).join('\n')
+  }
+
+  async lookupId (id) {
+    const sunetid = id.replaceAll('@stanford.edu', '')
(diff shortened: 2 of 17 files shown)


app/components/collections/edit/participant_component.rb
```diff
@@ -15,13 +15,13 @@ def initialize(form:)
...
+      # def lookup_id
+      #   form.field_id(:lookup)
+      # end
 
-      def label
-        'SUNet ID'
-      end
+      # def label
+      #   'SUNet ID'
+      # end
```
mjgiarlo, line 24: Cruft?

app/javascript/controllers/participants_controller.js
```diff
@@ -0,0 +1,53 @@
...
+      })
+      .then(data => {
+        if (!data) {
+          this.addError(id)
+          return id
+        }
+        this.addParticipant(data)
+        return null
+      })
+      .catch(error => console.dir(error))
```
mjgiarlo, line 37: Is this worth honeybadgering or not so much?

mjgiarlo (approved):
Approve pending comments

app/javascript/controllers/participants_controller.js
justinlittman: We haven't HBed from JS before when accessing our own APIs. I'd say "no" for now, but "definitely" if we encounter any problems.

app/components/collections/edit/participant_component.rb
justinlittman: Thanks.  Removed.

PR #982, opened 2025-03-07
Omit form in cocina when no type selected.

### LINKED ISSUE #973: Omit form in cocina when no type selected
When saving a draft of a work without a type, the following is recorded in cocina:
```
"form": [
      {
        "structuredValue": [
          {
            "type": "type"
          }
        ],
        "type": "resource type",
        "source": {
          "value": "Stanford self-deposit resource types"
        }
      }
    ],
```

Since there is no type, there should be no entries under form.

justinlittman:
closes #973

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -48,6 +48,9 @@ class WorkForm < ApplicationForm
   validates :license, presence: true, on: :deposit
 
   attribute :work_type, :string
+  before_validation do
+    self.work_type = work_type.presence
+  end
   validates :work_type, presence: true, on: :deposit
 
   attribute :work_subtypes, array: true, default: -> { [] }
--- a/app/services/to_cocina/work/types_mapper.rb
+++ b/app/services/to_cocina/work/types_mapper.rb
@@ -16,7 +16,7 @@ module ToCocina
       end
 
       def call
-        return [] unless work_type
+        return [] unless work_type.presence
 
         self_deposit_resource_types + genres + resource_types + datacite_types
       end
--- a/spec/services/to_cocina/work/types_mapper_spec.rb
+++ b/spec/services/to_cocina/work/types_mapper_spec.rb
@@ -13,6 +13,15 @@ RSpec.describe ToCocina::Work::TypesMapper do
     end
   end
 
+  context 'when blank types or subtypes' do
+    # Due to normalization in the work form this should never happen. Howver, testing just in case.
+    let(:work_form) { WorkForm.new(work_type: '', work_subtypes: ['', '']) }
+
+    it 'returns an empty array' do
+      expect(forms).to eq([])
+    end
+  end
+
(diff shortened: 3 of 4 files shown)


app/forms/work_form.rb
```diff
@@ -46,6 +46,7 @@ def persisted?
   validates :license, presence: true, on: :deposit
 
   attribute :work_type, :string
+  before_validation ->(work_form) { work_form.work_type.presence }
```
mjgiarlo, line 49: Can you say more about what this is doing? Like, if the block returns a `nil` does that short-circuit validation or something?

justinlittman: It just changes a `''` to a `nil` so that code using this form doesn't have to remember to treat `''` as `nil`.

The mapper forgot to do this, which is what caused the bug. This is just preventative of future bugs.

mjgiarlo: @justinlittman Doesn't `#presence` return the value instead of mutating?

mjgiarlo: FWIW, when I comment out the `before_validation` the mapper and form specs still pass.

justinlittman:
```
hungry-hungry-hippo(dev)> ''.presence
=> nil
hungry-hungry-hippo(dev)> 'foo'.presence
=> "foo"
```

Yes, I would expect the tests to pass. For the sake of defensiveness, I fixed the same problem in 2 places.

justinlittman: Oh, you're right. Duh.

app/services/to_cocina/work/types_mapper.rb
```diff
@@ -16,7 +16,7 @@ def initialize(work_form:)
       end
 
       def call
-        return [] unless work_type
+        return [] unless work_type.presence
```
mjgiarlo, line 19: With the change you just made to the validation callback in the work form, do we still need this check here, or are blank string work types still a thing we'll expect to see coming from other sources?

app/forms/work_form.rb
```diff
@@ -46,6 +46,7 @@ def persisted?
   validates :license, presence: true, on: :deposit
 
   attribute :work_type, :string
+  before_validation ->(work_form) { work_form.work_type = work_form.work_type.presence }
```
mjgiarlo, line 49: You may want to consider a related tweak here, too: https://github.com/sul-dlss/hungry-hungry-hippo/pull/1000

app/services/to_cocina/work/types_mapper.rb
justinlittman: Just being defensive.

app/forms/work_form.rb
justinlittman: Done.

mjgiarlo (approved):

PR #996, opened 2025-03-07
Avoid race condition for marking persist complete

justinlittman:


mjgiarlo (approved):

mjgiarlo:
Seems to fix the race condition? Let's merge this and revisit if the problem crops up again. Thanks, @justinlittman

PR #999, opened 2025-03-07
Allow user to change DOI option.

### LINKED ISSUE #972: User should be able to change DOI option on draft object
When a collection is set for a user to be able to decide whether to get a DOI or not, they should be able to change their mind about that until the item is deposited. 

Here's what happens now when the collection is set for the depositor to choose the DOI:
1. I create a new item and add a title.
2. Here's what the DOI page looks like by default:

Image

3. I then save as draft.
4. When I go back in to edit, the DOI page looks like this, and I have no choice to change my mind about it. This is particularly disconcerting when the user hasn't even been to this page before saving as draft and never got to make the decision once.

Image

One thing we could do is set the default to "No," which may help, but it doesn't solve the issue of what happens when people want to change their minds at this point.

As a reminder, it has always been our policy that there are no take-backs once the item is deposited. Any DOI issued for a published item is permanent.

@astridu Thoughts from you on what we should do here? Thanks.

justinlittman:
closes #972

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -27,7 +27,11 @@ class Doi
   end
 
   def assigned?
-    @assigned ||= client.exists?(id:)
+    @assigned ||= begin
+      result = client.exists?(id:) # returns a Dry::Monads result
+      Honeybadger.notify("Datacite error for #{druid}: #{result.failure}") if result.failure?
+      result.value_or(false)
+    end
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/doi.rb
```diff
@@ -27,7 +27,11 @@ def url
   end
 
   def assigned?
-    @assigned ||= client.exists?(id:)
+    @assigned ||= begin
+      result = client.exists?(id:) # returns a Dry::Monads result
```
justinlittman, line 31: Ooops, this is a Dry-monad Result, not a boolean.

mjgiarlo (approved):

PR #1003, opened 2025-03-07
Disable subtype entries that are not shown.

### LINKED ISSUE #949: Subtype entries should only be listed in the metadata once
See https://sul-h3-stage.stanford.edu/works/druid:sc054dj0261.

Stella was testing and did many things before getting to this state.

justinlittman:
closes #949

### DIFF
--- a/app/javascript/controllers/work_type_controller.js
+++ b/app/javascript/controllers/work_type_controller.js
@@ -11,11 +11,9 @@ export default class extends Controller {
   showSubtypes (event) {
     const workType = event.target.value
     this.subtypeSectionTargets.forEach((section) => {
-      if (section.dataset.workType === workType) {
-        section.classList.remove('d-none')
-      } else {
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/work_type_controller.js
```diff
@@ -11,11 +11,9 @@ export default class extends Controller {
...
     const workType = event.target.value
     this.subtypeSectionTargets.forEach((section) => {
-      if (section.dataset.workType === workType) {
-        section.classList.remove('d-none')
-      } else {
-        section.classList.add('d-none')
-      }
+      const show = section.dataset.workType === workType
+      section.classList.toggle('d-none', !show)
+      section.querySelectorAll('input').forEach((input) => { input.disabled = !show })
```
justinlittman, line 16: This is the fix.

mjgiarlo (approved):

PR #1011, opened 2025-03-08
Only display whats changing input after deposit.

### LINKED ISSUE #1006: "What's changing?" box should not display until after an item has been deposited
The "What's changing" box is intended to provide a note about what has changed for items that have already been deposited. I'm seeing it on an item that has never been deposited. It's probably been this way the whole time and I've just never registered this.

Image

Also, I thought we had changed this so the box was always empty by default. Maybe that just hasn't been pushed out yet.

justinlittman:
closes #1006

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -8,6 +8,7 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
   before_action :set_work_form_from_cocina, only: %i[show edit review]
   before_action :set_content, only: %i[show edit review]
   before_action :set_presenter, only: %i[show edit review]
+  before_action :set_license_presenter, only: %i[edit]
 
   def show
     authorize! @work
@@ -26,7 +27,7 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
 
     @content = Content.create!(user: current_user)
     @work_form = new_work_form
-    @license_presenter = LicensePresenter.new(work_form: @work_form, collection: @collection)
+    set_license_presenter # Note this requires @collection and @work_form set above.
 
     render :form
   end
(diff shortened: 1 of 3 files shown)


app/controllers/works_controller.rb
```diff
@@ -26,7 +27,7 @@ def new
 
     @content = Content.create!(user: current_user)
     @work_form = new_work_form
-    @license_presenter = LicensePresenter.new(work_form: @work_form, collection: @collection)
+    set_license_presenter
```
justinlittman, line 30: Ended up doing a little refactoring to get the complexity down on `update`.

app/views/works/form.html.erb
```diff
@@ -124,7 +124,7 @@
 
       <% component.with_pane do %>
         <%= render Works::Edit::PaneComponent.new(tab_name: :deposit, label: t('works.edit.panes.deposit.label'), form_id:, discard_draft_form_id:, work_presenter: @work_presenter, help_text: t('works.edit.panes.deposit.help_text'), active_tab_name:, next_tab_btn: false) do |component| %>
-          <% if @work_form.persisted? %>
+          <% if @work_presenter && !@work_presenter.first_version? %>
```
justinlittman, line 127: This was the fix.

app/controllers/works_controller.rb
mjgiarlo: @justinlittman What do you think about adding a comment here remind us that we can't move this line to the `before_action` above because we need to set `@collection` and `@work_form` to special values in `#new` before that method is called?

mjgiarlo (approved):
Approve pending comment consideration

app/controllers/works_controller.rb
justinlittman: Makes sense. Fixing.

PR #1020, opened 2025-03-10
Only send deposit email when collection manager deposits to a collect…

### LINKED ISSUE #976: Only send deposit email when collection manager deposits to a collection
When a collection manager deposits an item in a collection, they are currently getting two emails -- the one to the depositor saying they're deposit is complete, and the one to the manager saying there has been a deposit to the collection.

They should only receive the one to the depositor.

justinlittman:
…ion.

closes #976

### DIFF
--- a/app/jobs/deposit_collection_job.rb
+++ b/app/jobs/deposit_collection_job.rb
@@ -4,9 +4,12 @@
 class DepositCollectionJob < ApplicationJob
   # @param [CollectionForm] collection_form
   # @param [Collection] collection
-  def perform(collection_form:, collection:)
+  # @param [User] current_user
+  def perform(collection_form:, collection:, current_user:)
     @collection_form = collection_form
     @collection = collection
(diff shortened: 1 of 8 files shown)


app/jobs/deposit_collection_job.rb
```diff
@@ -4,9 +4,12 @@
 class DepositCollectionJob < ApplicationJob
   # @param [CollectionForm] collection_form
   # @param [Collection] collection
-  def perform(collection_form:, collection:)
+  # @param [User] current_user
+  def perform(collection_form:, collection:, current_user:)
     @collection_form = collection_form
     @collection = collection
+    # Setting current user so that it will be available for notifications.
+    Current.user = current_user
```
justinlittman, line 12: This was working locally because jobs were being run in the same process; however, in production they're not so need to set the current user.

aaron-collier (approved):

PR #1033, opened 2025-03-10
[DRAFT] Improve field accessibility by adding aria-required

### LINKED ISSUE #909: [SODA] Form field validation (header)
*Note:* I'm unclear if this is a duplicate or a different issue at this point:

The form field with an error is not identified in text, or text alternative, or via programmatic association.
There are several issues with the form validation.

- When a field is in error, an error message appears visibly inline, but it is not programmatically associated with a form field and disappears when a user begins interacting with the form. 
- When the error is a required field not entered, the text is a generic "Please fill out this field."
- Fields that are in error do not have aria-invalid="true" set on them.

There is also the issue that there are different types of error management which can be very confusing for users.

Soda recommendation:
```
The form field must be associated with the error text via aria-describedby as outline in the guidance that follows.

The error message text must be descriptive regarding the error and the form field in error. 

When a form field is error, aria-invalid="true" should be added.

Ensure there is one type of error management across the site.

WebAIM - Usable and Accessible Form Validation and Error Recovery
https://webaim.org/techniques/formvalidation/

RULE : 
Text-based alerts that identify the field in error MUST be provided to users for form validation cues and errors.

HOW TO FIX: 
Fix this issue by doing AT LEAST ONE of the following:
1. Programmatically associate the error message and the field in error - typically using the aria-describedby attribute to reference a unique id attribute value on the error message container. (STRONGLY PREFERRED)
2. Include the field name in the error message.

REFERENCE: 
Deque University: https://dequeuniversity.com/class/forms/form-validation/error-identification
W3C-WAI tutorial: https://www.w3.org/WAI/tutorials/forms/notifications/

BACKGROUND: 
When form fields that have an error are not identified with a text-based error message people who are blind or have low vision may not be able to understand that an error has occurred and what went wrong. In order to provide feedback that can be accessed by all users, error messages must be provided in a text format. Error messages must indicate which input is in error - either by the text of the error message or programmatically - and the nature of the error. This will allow people to recover from mistakes while filling out a form.

```

See issue id 2228 in https://docs.google.com/spreadsheets/d/1x0uBBDe3T8O0HNCs74OdVKQc8tQQNvOV_f-C6IS44IQ/edit?gid=1050712151#gid=1050712151

aaron-collier:
Fixes #909 

I think this is the other part of improving the accessibility of forms, I'm opening this as DRAFT for the purposes of raising for discussion.

One issue reported by SODA is that required fields are not properly indicated both visually and in the aria values for screen readers.

This adds an asterisk and `area-required=true` when a `TextField` gets `required: true` but also removes `required` from the rendered component. The reason for this is that the `required` value on a field component prevents the form from being submitted when the value is missing which breaks our server side form validation. But including the `area-required=true` should work here without the bare `required` as SODA reports: 

```
Required fields need instructions or labels to visible indicate they are required and the form field must include "required" or aria-required="true"
```

### DIFF
--- a/app/components/elements/forms/field_component.rb
+++ b/app/components/elements/forms/field_component.rb
@@ -7,10 +7,11 @@ module Elements
       def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists, Metrics/MethodLength, Metrics/AbcSize
                      disabled: false, hidden: false, data: {}, input_data: {}, placeholder: nil, width: nil,
                      label_classes: [], container_classes: [], input_classes: [], tooltip: nil, error_classes: [],
-                     readonly: false)
+                     readonly: false, mark_required: false)
         @form = form
         @field_name = field_name
         @required = required
+        @mark_required = mark_required
         @hidden_label = hidden_label
         @label = label
         @help_text = help_text
@@ -36,9 +37,19 @@ module Elements
         @help_text_id ||= form.field_id(field_name, 'help')
       end
 
+      # Used to tell the label to draw a "*" to indicate required
+      def mark_required?
+        @mark_required || required
+      end
+
       def field_aria
         {}.tap do |arias|
-          arias[:required] = true if required
+          # Set aria-required if we want to indicate required, but the field
+          # does not actually have a required attribute
+          #
+          # This is used for collection/work forms where we do server-side
+          # validation and don't want to block form submission on empty fields
+          arias[:required] = true if @mark_required
         end.merge(error_aria)
       end
 
--- a/app/components/elements/forms/field_component.rb
+++ b/app/components/elements/forms/field_component.rb
@@ -37,6 +37,12 @@ module Elements
       end
 
       def field_aria
+        {}.tap do |arias|
+          arias[:required] = true if required
+        end.merge(error_aria)
+      end
+
+      def error_aria
         InvalidFeedbackSupport.arias_for(field_name:, form:).tap do |arias|
           arias[:describedby] = merge_actions(arias[:describedby], help_text_id) if help_text.present?
         end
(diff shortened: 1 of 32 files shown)


justinlittman:
@aaron-collier I think we need two types of required: (1) validate required (i.e., required on an input that prevents a form from being submitted) and (2) aria required (which indicates that to a user that at some point in the form life cycle (e.g., when depositing) the field will be required.

aaron-collier:
@justinlittman Thank you, this is why I opened this for discussion. SODA indicates "required OR area-reaquired=true", but they do behave differently. As you mention, "required" prevents the form from being submitted, which breaks our method of validation. So I'm not clear if this is sufficient for their review based on that "or", ... or if we need to go back to JS for our validation (which I think we all hope not).

app/components/elements/forms/field_component.rb
```diff
@@ -37,6 +37,12 @@ def help_text_id
       end
 
       def field_aria
+        {}.tap do |arias|
+          arias[:required] = true if required
+        end.merge(error_aria)
```
thatbudakguy, line 42: does this just set `aria-required` if the field is required? that would have no effect, if so – `aria-required` only exists for cases where you cannot set the ordinary required attribute. if `required` is set, the browser or SR will already know that without the need for aria.

aaron-collier: @thatbudakguy right, the bigger issue here is that setting `required` means the form will not submit, but we do validation on the server side not the client side to avoid javascript for this.

thatbudakguy: it's totally OK (imo) to just set `aria-required` everywhere we would otherwise want to set `required`. the point is that you never need both. `required` has behavior we don't want (it blocks submission of the form), so all we need to do is set `aria-required` to convey the same thing without blocking submission.

aaron-collier: @thatbudakguy I feel like I'm missing something, isn't that what this is doing then? Setting `aria-required: true` when we pass `required: true` to the component? This PR removes `required` from the text field rendered by the component in favor of `aria-required`...

thatbudakguy: Oh, then I misunderstood! I thought it additionally set aria-required everywhere there was already required, instead of replacing.

PR #1038, opened 2025-03-11
Set role and aria-label when a tab indicates an error

### LINKED ISSUE #908: [SODA] Text alternative for alert image is missing when creating a collection
The text alternative for the informative image is missing. When one of the alert images is present, this information needs to be conveyed to all users by including a text equivalent.

Soda recommendation:
```
RULE : 
Images that convey content MUST have programmatically-discernible alternative text.

HOW TO FIX: 
Fix this issue by using ONE of the following techniques:

1. Use the alt attribute on the <img> element to convey the intent, purpose, or meaning of the image.

<img src="sunset.jpg" alt="The sun setting over the Pacific Ocean, with a silhouette of a flying seagull in the foreground">

<img src="cat.svg" alt="cat" role="img"> (role is optional, for backwards compatibility with Safari)

2. Use aria-label and role="img" to convey the intent, purpose, or meaning of the image.

<p>Have a great day <span class="fonticon smiley" role="img" aria-label="Smiley face"></span></p>

<svg role="img" aria-label="cat">

3. Use CSS clipped text to provide screen reader accessible (non-visible) alternative text.

<a href="https://www.facebook.com/dequesystems" class="fb">
<span class="sr-text">Deque's Facebook page</span>
</a>

.sr-text {
position: absolute;
clip: rect(1px, 1px, 1px, 1px);
-webkit-clip-path: inset(50%) ;
clip-path: inset(50%);
padding: 0;
border: 0;
height: 1px;
width: 1px;
white-space: nowrap;
overflow: hidden;
}

REFERENCE: 
Deque University: https://dequeuniversity.com/class/images/alt-text/informative
W3C-WAI tutorial: https://www.w3.org/WAI/tutorials/images/

BACKGROUND: 
People who are blind cannot see images on a page. In order to give people who cannot see an image access to the information conveyed by the image, it must have a text alternative. The text alternative must describe the information or function represented by the image. Screen readers can then use the alternative text to convey that information to the screen reader user.

For informative images - images that convey information - the alternative text must communicate the intent, purpose, or meaning of the image in a way that serves as a true alternative for the image.

It is not possible to provide alternative text by adding an alt attribute directly to CSS background images. Images that convey content should not generally be placed in CSS. However, if it is necessary to do so, you can add alternative text to informative and actionable CSS images using other methods.
```

See issue id 2258 in https://docs.google.com/spreadsheets/d/1x0uBBDe3T8O0HNCs74OdVKQc8tQQNvOV_f-C6IS44IQ/edit?gid=1050712151#gid=1050712151

aaron-collier:
Fixes #908

### DIFF
--- a/app/javascript/controllers/tab_error_controller.js
+++ b/app/javascript/controllers/tab_error_controller.js
@@ -11,9 +11,10 @@ export default class extends Controller {
       const paneEl = this.paneTargets[index]
       if (paneEl.querySelector('.is-invalid')) {
         tabEl.classList.add('is-invalid')
-        tabEl.setAttribute('aria-invalid', 'true')
-        tabEl.setAttribute('aria-label', 'Error: required fields missing')
-        tabEl.setAttribute('role', 'img')
+        tabEl.setAttribute('aria-description', 'Error: required fields missing')
+      } else {
+        tabEl.classList.remove('is-invalid')
+        tabEl.removeAttribute('aria-description')
       }
     })
     // Show the first tab that has an invalid input.
--- a/app/javascript/controllers/tab_error_controller.js
+++ b/app/javascript/controllers/tab_error_controller.js
@@ -11,6 +11,9 @@ export default class extends Controller {
       const paneEl = this.paneTargets[index]
       if (paneEl.querySelector('.is-invalid')) {
         tabEl.classList.add('is-invalid')
+        tabEl.setAttribute('aria-invalid', 'true')
+        tabEl.setAttribute('aria-label', 'Error: required fields missing')
+        tabEl.setAttribute('role', 'img')
       }
     })
     // Show the first tab that has an invalid input.


app/javascript/controllers/tab_error_controller.js
```diff
@@ -11,6 +11,9 @@ export default class extends Controller {
       const paneEl = this.paneTargets[index]
       if (paneEl.querySelector('.is-invalid')) {
         tabEl.classList.add('is-invalid')
+        tabEl.setAttribute('aria-invalid', 'true')
+        tabEl.setAttribute('aria-label', 'Error: required fields missing')
+        tabEl.setAttribute('role', 'img')
       }
```
thatbudakguy, line 17:
```suggestion
      if (paneEl.querySelector('.is-invalid')) {
        tabEl.classList.add('is-invalid')
        tabEl.setAttribute('aria-description', 'Error: required fields missing')
      } else {
        tabEl.classList.remove('is-invalid')
        tabEl.removeAttribute('aria-description')
      }
```

I don't think we want to set `aria-invalid` or a `role` here. The `aria-invalid` attribute is a standin for HTML input validation behavior; it doesn't really have an effect if set on things that aren't inputs because they can't be "invalid". Setting `role=img` also doesn't make sense because this element is a tab, not an image.

Instead, we can just add an additional description signaling that this tab contains errored fields, which is what the red exclamation mark means.

aaron-collier: @thatbudakguy do you think that is sufficient based on SODA's "how to fix" suggestions? If so, awesome!

thatbudakguy: Yes – their suggested solutions are all for making the exclamation have an alt, but this just accomplishes the same thing by adding that text to the description of the tab itself. We can't add an alt to the exclamation because it's a background image added via CSS pseudoselector.

aaron-collier: Ok, cool, that's very confusing though as their suggestion includes the `aria-invalid` and `aria-required` values... This stuff is just clear as mud.

thatbudakguy:
So, apparently `aria-description` is [still not fully supported in JAWS as of 2023](https://www.darins.page/articles/aria-description-crash-course#results).

That just means we have to do a slightly more annoying technique of adding an additional element that contributes to the accessible name instead, but I can work on doing that.

thatbudakguy:
Yep, that's exactly what I was thinking of doing. Alternatively I think Joey alluded to the possibility of somehow making the background-image of the exclamation point have text content, but I'm not sure how that works.

thatbudakguy:
Fixed in #1088

PR #1043, opened 2025-03-11
Support collection defaults for types and subtypes.

### LINKED ISSUE #991: Required default work types and subtypes
We want to allow the collection manager to set a required work type and subtypes at the collection level. We will largely recreate the work type/subtype selection form on a new page in the collection form.

Figma: @astridu ?

Requirements:
- The collection manager will not be required to set any required work types or subtypes. This is entirely optional. If they don't select anything, the choices will be left entirely up to the depositor.
- The "Other" work type will not be available as an option at the collection level.
- The collection manager will be able to set a single work type that will be applied to every deposit in the collection. The depositor will not be able to change this work type or add another (we only ever allow 1 work type).
- The collection manager will also be able to select one or more subtypes. The work subtypes that are displayed should follow the same logic that we currently use for the deposit form, which is to say that the same 'short lists' will display at the top based on the work type selected and the same modified versions of the full lists we use in the deposit form will display.
- The collection manager will not be required to meet the requirements for the 'Music' and 'Mixed Materials' work types, but the Depositor must still meet these requirement. For example, the collection manager can choose 'Field recording' as a required subtype for the 'Music' work type, which does not meet the requirement to choose one item from the short list. However, the Depositor will be required to choose a subtype from the short list to meet the requirement.
- Depositor will not be allowed to alter the subtypes selected by the collection manager but will be able to add additional subtypes, and as noted above, may be required to in some cases.
- Collection managers will not be allowed to select a subtype without first choosing a work type.
- Work types and subtypes selected by the collection manager will display on the work form as checked but will not be alterable.

We also need to check when moving an item to a different collection that the work types and subtypes don't conflict with the new collection requirements.

justinlittman:
closes #991

https://github.com/user-attachments/assets/85bdaa2a-b1ef-4f5c-88cb-97e1b10d11e9

### DIFF
--- a/app/components/collections/edit/work_type_component.html.erb
+++ b/app/components/collections/edit/work_type_component.html.erb
@@ -1,0 +1,37 @@
+<div data-controller="work-type">
+  <div class="pane-section">
+    <%= render Elements::Forms::FieldsetComponent.new(label: t('collections.edit.fields.work_type.label'), tooltip: t('collections.edit.fields.work_type.tooltip_html')) do %>
+      <div class="types-container">
+        <%= render Elements::Forms::InputCollectionRadioButtonsComponent.new(form:, field_name: :work_type, input_collection:, value_method: :value, input_data: { work_type_target: 'workTypeRadio', action: 'work-type#showSubtypes click->tooltips#hideAll' }) %>
+      </div>
+    <% end %>
+  </div>
+
+  <%# Text subtypes %>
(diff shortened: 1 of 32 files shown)


justinlittman:
https://github.com/user-attachments/assets/fd1f12b0-5056-4c3c-8977-f88badc7adf4

aaron-collier (approved):

PR #1049, opened 2025-03-11
Address issues reported as serious by SODA on the homepage

aaron-collier:
This addresses, using their recommendations, all but a couple of the "serious" issues reported by SODA on the homepage. The 2 remaining we will need clarification from them on.

### DIFF
--- a/app/views/home/show.html.erb
+++ b/app/views/home/show.html.erb
@@ -1,7 +1,7 @@
 <% content_for :fluid_body do %>
   <%= tag.div class: 'hero', style: "background-image: url(#{image_path(@hero_image)});" do %>
-    <div class="row align-items-center justify-content-center h-100 gy-2">
-      <div class="col-md-3 col-sm-12">
+    <div class="row align-items-center justify-content-center h-100 gy-2" role="list">
+      <div class="col-md-3 col-sm-12" role="listitem">
         <%= render Elements::CardComponent.new(classes: 'text-center mx-4 mx-md-2 opacity-85') do |component| %>
           <% component.with_body do %>
             <div class="mb-3">
@@ -13,7 +13,7 @@
           <% end %>
         <% end %>
       </div>
-      <div class="col-md-3 col-sm-12">
+      <div class="col-md-3 col-sm-12" role="listitem">
         <%= render Elements::CardComponent.new(classes: 'text-center mx-4 mx-md-2') do |component| %>
           <% component.with_body do %>
             <div class="mb-2">
@@ -24,7 +24,7 @@
           <% end %>
         <% end %>
       </div>
-      <div class="col-md-3 col-sm-12">
+      <div class="col-md-3 col-sm-12" role="listitem">
         <%= render Elements::CardComponent.new(classes: 'text-center mx-4 mx-md-2 opacity-85') do |component| %>
           <% component.with_body do %>
             <div class="mb-3">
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.bootstrap.scss
```diff
@@ -337,4 +337,10 @@ fieldset.form-fieldset legend {
   input:not(:checked) + label  {
     @extend .btn-outline-primary;
   }
-}
\ No newline at end of file
+}
+
+.card-text {
+  a {
+    text-decoration: underline !important;
```
aaron-collier, line 344: https://dequeuniversity.com/class/visual-design/color/distinguish-links-from-text

app/views/home/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :fluid_body do %>
   <%= tag.div class: 'hero', style: "background-image: url(#{image_path(@hero_image)});" do %>
-    <div class="row align-items-center justify-content-center h-100 gy-2">
-      <div class="col-md-3 col-sm-12">
+    <div class="row align-items-center justify-content-center h-100 gy-2" role="list">
```
aaron-collier, line 3: https://dequeuniversity.com/class/semantic-structure/lists

app/views/home/show.html.erb
```diff
@@ -77,7 +77,7 @@
             </div>
             <label for="mce-EMAIL" class="visually-hidden">Sign up for our newsletter</label>
             <div class="input-group">
-              <input type="email" value="" name="EMAIL" class="form-control" id="mce-EMAIL" autocomplete="on" placeholder="Enter your email address" required>
+              <input type="email" value="" name="EMAIL" class="form-control" id="mce-EMAIL" autocomplete="email" placeholder="Enter your email address" required>
```
aaron-collier, line 80: https://www.w3.org/TR/WCAG22/#input-purposes

app/assets/stylesheets/application.bootstrap.scss
justinlittman: @aaron-collier I can't get to this page. Can you say more about what is happening? It seems surprising that we should be overriding `<a>`

aaron-collier: They are reporting that the color contrast between the link text and the non-link text is not high enough. One suggestion is to underline links, which felt a bit easier than shifting away from standard bootstrap colors.

justinlittman: This seems like it should be a component library issue. cc @thatbudakguy @astridu

aaron-collier: I removed this override in favor of giving the link the `card-link` class to keep it within the bootstrap idiom here. Not sure if that resolves the contrast issue however.

aaron-collier: While I agree these should be component library or design issues, I'm unclear if it's OK for us to wait for that. If it's not fixed before the app is in use, but it's on the horizon for being fixed in the component library, is that OK?

app/views/home/show.html.erb
thatbudakguy: this is a cool idea! I've never used `role=list` on a series of cards before, but it makes total sense – this is just a list of content that happens to be visually represented as cards.

PR #1050, opened 2025-03-11
Fix validation of whats changing.

### LINKED ISSUE #1048: User should not be able to deposit without comments in "What's changing" box
See https://app.honeybadger.io/projects/128495/faults/118281532.

- I opened an already-deposited item for editing.
- Clicked over to the deposit tab
- The "what's changing" box was there but I did not fill it in
- I clicked the deposit button and the validation did not catch the empty box and triggered the above alert.

justinlittman:
closes #1048

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -96,7 +96,7 @@ class WorkForm < ApplicationForm
   validates :create_date_type, inclusion: { in: %w[single range] }
 
   attribute :whats_changing, :string
-  validates :whats_changing, presence: true, if: -> { persisted? && version > 1 }
+  validates :whats_changing, presence: true
 
   def content_file_presence
     return if content_id.nil? # This makes test configuration easier.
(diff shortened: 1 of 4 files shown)


app/forms/work_form.rb
```diff
@@ -96,7 +96,7 @@ def persisted?
   validates :create_date_type, inclusion: { in: %w[single range] }
 
   attribute :whats_changing, :string
-  validates :whats_changing, presence: true, if: -> { persisted? && version > 1 }
+  validates :whats_changing, presence: true
```
justinlittman, line 99: The problem is that this didn't validate when the user was making a change to a deposited version 1 (since at that time the version was still 1 it bypassed the validation).

Unfortunately, this state can't be detected without having access to version status, so took a different approach to the validation by providing a dummy value when an open first version.

spec/forms/work_form_spec.rb
```diff
@@ -248,81 +262,4 @@
       expect(form.custom_rights_statement).to eq("This is a test.\r\n\r\nThis is a second paragraph.")
     end
   end
-
-  describe 'Whats changing validation' do
```
justinlittman, line 252: Good news is I could get rid of all of these tests.

mjgiarlo (approved):

PR #1058, opened 2025-03-12
Do not send the deposit first version email when updating a collection

### LINKED ISSUE #951: Do not send email to admins about new collection being created when it's only been updated
When I update an existing an collection, I'm getting the email from app/views/collections_mailer/first_version_created_email.html.erb

I should only get this the first time the collection is saved.

aaron-collier:
Fixes #951

### DIFF
--- a/app/jobs/deposit_collection_job.rb
+++ b/app/jobs/deposit_collection_job.rb
@@ -17,15 +17,13 @@ class DepositCollectionJob < ApplicationJob
 
     update_collection_record(druid:)
 
-    unless new_cocina_object
+    if new_cocina_object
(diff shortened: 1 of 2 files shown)


app/jobs/deposit_collection_job.rb
```diff
@@ -17,13 +17,13 @@ def perform(collection_form:, collection:, current_user:)
 
     update_collection_record(druid:)
 
+    ModelSync::Collection.call(collection:, cocina_object: new_cocina_object) if new_cocina_object
```
justinlittman, line 20: Feels like L20 and L22 should be combined into an `if ... else`

justinlittman (approved):

PR #1067, opened 2025-03-12
Update work subtype validation when music is selected

### LINKED ISSUE #1064: Music work type should require at least one subtype from the short list
When a user selects the work type "Music" they are required to choose one of the subtypes from the short list that displays above the full list:

Image

Right now deposit happens successfully if any subtype is selected. See https://sul-h3-stage.stanford.edu/works/druid:gk982fh6304, which I successfully deposited with one of the non-short-listed subtypes selected:

Image

Note that it *is* properly preventing me from depositing if I have only chosen the "Music" work type and no subtypes.

aaron-collier:
Fixes #1064 

Validates that one of the selected subtypes is one of the required music subtypes when the music type is selected.

Refactored into 2 conditional validation, this seems much clearer.

~~DRAFT: After thinking about the codecov failure (there was a single line uncovered) I decided I thought this would be clearer if I used affirmative language (`music_subifields?`) vs negative (`missing_music_subfields?`) which altered the logic a bit. Reworking that and the testing now and then I will undraft.~~

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -55,7 +55,8 @@ class WorkForm < ApplicationForm
 
   attribute :work_subtypes, array: true, default: -> { [] }
   before_validation { work_subtypes.compact_blank! }
-  validate :correct_work_subtype_length
+  validate :music_subtypes, if: -> { work_type == WorkType::MUSIC }
+  validate :mixed_materials_subtypes, if: -> { work_type == WorkType::MIXED_MATERIALS }
 
   attribute :other_work_subtype, :string
   # Other requires a work subtype string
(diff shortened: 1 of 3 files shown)


app/forms/work_form.rb
```diff
@@ -125,15 +125,27 @@ def create_date_range_sequence
     errors.add(:create_date_range_from, 'must be before end date')
   end
 
-  def correct_work_subtype_length
+  def validate_work_subtypes
```
justinlittman, line 128: This feels sorta awkward. I'd recommend 2 separate [conditional validations](https://guides.rubyonrails.org/active_record_validations.html#conditional-validations) (one for music and one for mixed materials).

aaron-collier: Agree, working through it. Thanks :-)

justinlittman (approved):

PR #1068, opened 2025-03-13
Slows spinner.

### LINKED ISSUE #1063: Slow down rosette spinner just a little more
It is looking better, but still spinning a bit too fast. Could we try half that speed? Thank you!

justinlittman:
closes #1063


justinlittman:
@astridu approved.

mjgiarlo (approved):

PR #1069, opened 2025-03-13
Pane-adjacent component refactors

mjgiarlo:
Recommend reviewing with whitespace diffs turned off.

### DIFF
--- a/app/components/edit/tab_form/tab_list_component.rb
+++ b/app/components/edit/tab_form/tab_list_component.rb
@@ -4,7 +4,7 @@ module Edit
   module TabForm
     # Component for rendering tabbed navigation for the provided form.
     class TabListComponent < ApplicationComponent
-      renders_many :tabs, 'Edit::TabForm::TabComponent'
+      renders_many :tabs, Edit::TabForm::TabComponent
       # The reason for the before_form_panes is to
       # avoid nested form tags, which aren't allowed in HTML.
       # Note that to reference the main form from elements in the before_form_pane,
@@ -12,8 +12,11 @@ module Edit
       # The form_id can then be provided as the form attribute to an input element to
       # connect it to the main form.
       # This approach is used to allow submit buttons to be outside the main form.
-      renders_many :before_form_panes
-      renders_many :panes
+      renders_many :before_form_panes, Works::Edit::PaneComponent
+      renders_many :panes, types: {
+        work: { renders: Works::Edit::PaneComponent, as: :work_pane },
+        collection: { renders: Collections::Edit::PaneComponent, as: :collection_pane }
+      }
 
       def initialize(model:, form_id: nil, hidden_fields: [], classes: [])
         @classes = classes
--- a/app/components/elements/breadcrumb_nav_component.rb
+++ b/app/components/elements/breadcrumb_nav_component.rb
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 module Elements
-  # Displays the top bread crumb navigation
+  # Displays the top breadcrumb navigation
   class BreadcrumbNavComponent < ApplicationComponent
     renders_many :breadcrumbs, types: {
-      breadcrump: { renders: BreadcrumbComponent, as: :breadcrumb },
-      title_breadcrumb: { renders: ->(**args) { BreadcrumbComponent.new(truncate_length: 75, **args) },
-                          as: :title_breadcrumb },
-      collection_breadcrumb: { renders: lambda { |**args|
-        BreadcrumbComponent.new(truncate_length: 50, **args)
+      default: {
+        renders: BreadcrumbComponent, as: :breadcrumb
       },
-                               as: :collection_breadcrumb }
-
+      title: {
+        renders: ->(**args) { BreadcrumbComponent.new(truncate_length: 75, **args) }, as: :title_breadcrumb
+      },
+      collection: {
+        renders: ->(**args) { BreadcrumbComponent.new(truncate_length: 50, **args) }, as: :collection_breadcrumb
+      }
     }
   end
 end
--- a/app/views/collections/form.html.erb
+++ b/app/views/collections/form.html.erb
@@ -31,75 +31,57 @@
     <%# The panes below need a form in order to render their contents. This provides a "fake" form to use for that purpose. %>
     <%# TabListComponent will render the actual form. %>
     <% form_with(model: @collection_form) do |form| %>
-      <% component.with_pane do %>
-        <%= render Collections::Edit::PaneComponent.new(tab_name: :details, label: t('collections.edit.panes.details.label'), active_tab_name:) do %>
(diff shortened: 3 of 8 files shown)


app/components/edit/tab_form/tab_list_component.rb
```diff
@@ -4,16 +4,19 @@ module Edit
...
       # The form_id can then be provided as the form attribute to an input element to
       # connect it to the main form.
       # This approach is used to allow submit buttons to be outside the main form.
-      renders_many :before_form_panes
-      renders_many :panes
+      renders_many :before_form_panes, Works::Edit::PaneComponent
+      renders_many :panes, types: {
+        work: { renders: Works::Edit::PaneComponent, as: :work_pane },
+        collection: { renders: Collections::Edit::PaneComponent, as: :collection_pane }
+      }
```
mjgiarlo, line 19: Being more explicit about which components these slots expect has trade-offs. The downside is that we are temporarily coupled to using these components in these slots. The upside is less machinery when creating panes. See below for examples of the payoff.

app/components/elements/breadcrumb_nav_component.rb
```diff
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 module Elements
-  # Displays the top bread crumb navigation
+  # Displays the top breadcrumb navigation
   class BreadcrumbNavComponent < ApplicationComponent
     renders_many :breadcrumbs, types: {
-      breadcrump: { renders: BreadcrumbComponent, as: :breadcrumb },
```
mjgiarlo, line 7: Typo: `breadcrump`

app/components/elements/breadcrumb_nav_component.rb
```diff
@@ -1,17 +1,18 @@
...
+        renders: BreadcrumbComponent, as: :breadcrumb
       },
-                               as: :collection_breadcrumb }
-
+      title: {
+        renders: ->(**args) { BreadcrumbComponent.new(truncate_length: 75, **args) }, as: :title_breadcrumb
+      },
+      collection: {
+        renders: ->(**args) { BreadcrumbComponent.new(truncate_length: 50, **args) }, as: :collection_breadcrumb
+      }
```
mjgiarlo, line 15: Mostly whitespace normalization. This block of code was formatted strangely to my eyes.

app/views/collections/form.html.erb
```diff
@@ -30,69 +30,53 @@
...
     <% form_with(model: @collection_form) do |form| %>
-      <% component.with_pane do %>
-        <%= render Collections::Edit::PaneComponent.new(tab_name: :details, label: t('collections.edit.panes.details.label'), active_tab_name:) do %>
-          <%= render Elements::Forms::TextFieldComponent.new(form:, field_name: :title, label: t('collections.edit.fields.title.label'), tooltip: t('collections.edit.fields.title.tooltip_html'), container_classes: 'pane-section') %>
-          <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :description, label: t('collections.edit.fields.description.label'), tooltip: t('collections.edit.fields.description.tooltip_html'), container_classes: 'pane-section') %>
-          <%= render NestedComponentPresenter.for(form:, field_name: :contact_emails, model_class: ContactEmailForm, form_component: Collections::Edit::ContactEmailsComponent, hidden_label: true, bordered: false, single_field: true, fieldset_classes: 'pane-section') %>
-        <% end %>
+      <% component.with_collection_pane(tab_name: :details, label: t('collections.edit.panes.details.label'), active_tab_name:) do %>
```
mjgiarlo, line 33: Now, instead of the pattern being:

```ruby
component.with_pane do # no context at all here
  render Collections::Edit::PaneComponent.new(lots:, of:, args:, here) # This is always the same pane component in the work context and always the same in the collection context, which suggests we have distinct pane component types already, just undeclared ones.
  # pane content here
end
```

It's now this:

```ruby
component.with_pane(lots:, of:, args:, here) do # no context at all here
  # pane content here
end
```

app/views/collections/form.html.erb
```diff
@@ -30,69 +30,53 @@
...
       <% end %>
 
-      <% component.with_pane do %>
-        <%= render Collections::Edit::PaneComponent.new(tab_name: :deposit, label: t('collections.edit.panes.deposit.label'), help_text: t('collections.edit.panes.deposit.help_text'), active_tab_name:) do |component| %>
-          <%# For spacing %>
-          <div class="pane-section"></div>
-          <% component.with_deposit_button do %>
-            <%= render Elements::Forms::SubmitComponent.new(label: 'Save', data: { action: 'unsaved-changes#allowFormSubmission' }) %>
-          <% end %>
+      <% component.with_collection_pane(tab_name: :deposit, label: t('collections.edit.panes.deposit.label'), help_text: t('collections.edit.panes.deposit.help_text'), active_tab_name:) do |pane_component| %>
```
mjgiarlo, line 75: We were shadowing the outer `component` variable here so I renamed it.

app/views/works/form.html.erb
```diff
@@ -35,107 +35,83 @@
...
+        </div>
       <% end %>
 
-      <% component.with_pane do %>
-        <%= render Works::Edit::PaneComponent.new(tab_name: :title, label: t('works.edit.panes.title.label'), form_id:, discard_draft_form_id:, work_presenter: @work_presenter, active_tab_name:) do %>
-          <%= render Elements::Forms::TextareaFieldComponent.new(form:, field_name: :title, label: t('works.edit.fields.title.label'), tooltip: t('works.edit.fields.title.tooltip_html'),
-                                                                 help_text: t('works.edit.fields.title.help_text'), width: 500, rows: 1, container_classes: 'pane-section') %>
-          <%= render NestedComponentPresenter.for(form:, field_name: :contact_emails, model_class: ContactEmailForm, form_component: Works::Edit::ContactEmailsComponent, hidden_label: true, bordered: false, single_field: true, fieldset_classes: 'pane-section') %>
-        <% end %>
+      <% component.with_work_pane(tab_name: :title, label: t('works.edit.panes.title.label'), form_id:, discard_draft_form_id:, work_presenter: @work_presenter, active_tab_name:) do %>
```
mjgiarlo, line 52: Same as above but for works instead of components.

app/views/collections/form.html.erb
mjgiarlo: An alternative we could also consider is instead of typed slot components, a lambda slot component could sniff the `model` in `TabListComponent` and render the appropriate component if we're worried about coupling.

app/components/edit/tab_form/tab_list_component.rb
justinlittman: This is totally reasonable since this component is specifically for work and collections (i.e., it is not generic).

app/components/elements/breadcrumb_nav_component.rb
justinlittman: Thanks for the comment before I tried to parse the differences.

justinlittman (approved):

PR #1078, opened 2025-03-14
Adds required contributors.

### LINKED ISSUE #1018: Allow for entry of required default contributor(s)
Collection manager should be allowed to enter one or more default author(s)/contributor(s) that will be included with every deposit in the collection.

- The manager-entered contributors will be required for the deposit.
- The deposit will not be able to remove the entry or change any of the details about the entry (e.g. text in the box, the role term applied, etc).
- The depositor will always be able to reorder the entry in the list. 
- The collection manager will not have a choice as to where/how the entry(ies) is(are) ordered in the list. 
- The entry will be autopopulated in the contributor form for new deposits in the collection, but not for works that are changed after deposit.

This should require recreation of the contributor entry form in the collection form. No modifications to the form are needed.

justinlittman:
closes #1018


app/components/edit/contributor_component.html.erb
```diff
@@ -0,0 +1,186 @@
+<% if collection_required? %>
```
justinlittman, line 1: Renders the read-only and form version for a contributor.

justinlittman: Also, this component was moved from `Works::Edit`.

app/components/show/contributors_show_component.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Show
```
justinlittman, line 3: This component was moved also.

aaron-collier (approved):

PR #1079, opened 2025-03-14
Add help text to collection default types tab

### LINKED ISSUE #1077: Add text instructions to Work type page on collection form
We need some more instructions visible on this page.

- Below the header "Select a work type for deposits in this collection" please put this text:

You can designate a required type of work that will be associated with every deposit to this collection. Depositors will not be able to change this or add other types.

- Below the header "Select which subtypes will be applied to each deposit" please put this text:

You can also designate required subtypes that will be associated with every deposit to this collection. Depositors will not be able to remove this subtype, but will be able to add more subtypes.

aaron-collier:
Fixes #1077 

Collection form update:

Screenshot 2025-03-14 at 9 49 42 AM

Confirming that this does not display on the Work form because help_text defaults to nil:

Screenshot 2025-03-14 at 9 50 04 AM


app/components/collections/edit/work_type_component.rb
```diff
@@ -19,6 +19,14 @@ def tooltip
         t('collections.edit.fields.work_subtypes.tooltip_html')
       end
 
+      def type_help_text
+        t('collections.edit.fields.work_type.help_text')
+      end
+
+      def subtype_help_text
+        t('collections.edit.fields.work_subtypes.help_text')
+      end
```
mjgiarlo, line 28: It looks like `subtype_help_text` is defined in the work type component but all it does here is get injected into the work subtypes component. How about moving this method directly into the work subtypes component where it's used?

app/components/works/edit/work_subtypes_component.rb
```diff
@@ -5,7 +5,7 @@ module Edit
     # Component for the work subtypes field
     class WorkSubtypesComponent < ApplicationComponent
       def initialize(form:, work_type:, label:, tooltip: nil, more_options: false, minimum_terms: 0, # rubocop:disable Metrics/ParameterLists
-                     error_field_name: nil, required_subtypes: [])
+                     error_field_name: nil, required_subtypes: [], help_text: nil)
```
mjgiarlo, line 8: Having this argument carry the help text suggests that we have different values of help text but we don't seem to. It may be clearer for the help text to live in this class and to change this param to `display_help_text: false`. And then change the injected help text params above in the collection work types component to `display_help_text: true`.

justinlittman: It has been our pattern to inject labels, help text, etc. in "generic" components. This seems consistent with that pattern.

I think part of the confusion is that `Works::Edit::WorkSubtypesComponent` should be moved to `Edit::WorkSubtypesComponent` since it is used both for works and collections.

aaron-collier: On it.

app/components/collections/edit/work_type_component.html.erb
```diff
@@ -1,35 +1,37 @@
 <div data-controller="work-type">
   <div class="pane-section">
     <%= render Elements::Forms::FieldsetComponent.new(label: t('collections.edit.fields.work_type.label'), tooltip: t('collections.edit.fields.work_type.tooltip_html')) do %>
+    <%= tag.p class: 'form-text mb-3' do %><%= type_help_text %><% end %>
```
justinlittman, line 4: Are you sure this needs the `form-text` class?

Also, per our (very loose) conventions, this should be a `<p>` instead of `tag.p` since there are no variables / other ruby.)

app/components/edit/work_subtypes_component.html.erb
```diff
@@ -3,6 +3,7 @@
     <%= content %>
   <% else %>
     <%= render Elements::Forms::FieldsetComponent.new(label:, tooltip:, aria: fieldset_aria) do %>
+      <%= tag.p class: 'form-text mb-3' do %><%= help_text %><% end %>
```
justinlittman, line 6: Same as above.

justinlittman (approved):

PR #1080, opened 2025-03-14
Add collection contact email for works deposited therein

### LINKED ISSUE #992: Allow collection manager to set a default email that is included with every deposit
Collection manager should be allowed to enter a contact email address that must be included with each deposit. This will be adequate to meet the deposit requirement for having a contact email.

The Depositor should see this contact email address in the form but should not be able to alter or delete it.

Seems like we need a little design around this one.

mjgiarlo:
Fixes #992


app/controllers/works_controller.rb
```diff
@@ -192,7 +192,8 @@ def new_work_form
       agree_to_terms: current_user.agree_to_terms?,
       contact_emails_attributes: [{ email: current_user.email_address }],
       work_type: @collection.work_type,
-      work_subtypes: @collection.work_subtypes
+      work_subtypes: @collection.work_subtypes,
+      deposits_contact_email: @collection.deposits_contact_email
```
mjgiarlo, line 196: Make sure the new work form has the collection's contact email for deposits on it.

app/forms/collection_form.rb
```diff
@@ -3,91 +3,71 @@
 # Form for a Collection
 class CollectionForm < ApplicationForm
   accepts_nested_attributes_for :related_links, :contact_emails, :managers, :depositors, :reviewers
-  before_validation do
-    blank_managers = managers_attributes.select(&:empty?)
-    next if blank_managers.empty?
-
-    self.managers_attributes = managers_attributes - blank_managers
-  end
-  validates :managers_attributes, length: { minimum: 1, message: 'must have at least one manager' } # rubocop:disable Rails/I18nLocaleTexts
```
mjgiarlo, line 12: I reorganized the form so it's a bit more readable to my eyes. I tried to group things with like things, i.e., attributes, validations, instance methods.

app/forms/collection_form.rb
```diff
@@ -3,91 +3,71 @@
...
 
   def self.immutable_attributes
     ['druid']
   end
 
+  attribute :access, :string, default: 'world'
+  attribute :custom_rights_statement_instructions, :string
+  attribute :custom_rights_statement_option, :string, default: 'no'
+  attribute :default_license, :string
+  attribute :deposits_contact_email, :string
```
mjgiarlo, line 15: Here's the new attribute. I chose this option over mucking about with the existing nested attributes (`contact_emails_attributes`).

app/forms/collection_form.rb
```diff
@@ -3,91 +3,71 @@
...
-  attribute :custom_rights_statement_option, :string, default: 'no'
+  validates :custom_rights_statement_instructions, presence: true,
+                                                   if: -> { custom_rights_statement_option == 'depositor_selects' }
   validates :custom_rights_statement_option, presence: true, inclusion: { in: %w[no provided depositor_selects] }
-
-  attribute :provided_custom_rights_statement, :string
+  validates :default_license, presence: true, if: -> { license_option == 'depositor_selects' }
+  validates :deposits_contact_email, format: {
+    with: URI::MailTo::EMAIL_REGEXP, allow_blank: true, message: I18n.t('contact_email.validation.email.invalid')
+  }
```
mjgiarlo, line 43: New validation for the new attribute.

app/forms/collection_form.rb
```diff
@@ -3,91 +3,71 @@
...
+  validates :doi_option, inclusion: { in: %w[yes no depositor_selects] }
+  validates :license, presence: true, if: -> { license_option == 'required' }
+  validates :license_option, inclusion: { in: %w[required depositor_selects] }
+  validates :managers_attributes, length: { minimum: 1, message: 'must have at least one manager' }
   validates :provided_custom_rights_statement, presence: true, if: -> { custom_rights_statement_option == 'provided' }
-  before_validation do
-    self.provided_custom_rights_statement = LinebreakSupport.normalize(provided_custom_rights_statement)
-  end
+  validates :release_duration, presence: { message: 'select a valid duration for release' },
+                               if: -> { release_option == 'depositor_selects' }
```
mjgiarlo, line 51: This used to be a `validate` call pointing at a method, but I didn't see a reason to deviate from the pattern we use for all the other validations.

app/services/to_cocina/work/description_mapper.rb
```diff
@@ -62,7 +62,14 @@ def related_resource_params
...
       def access_contact_params
-        CocinaGenerators::Description.contact_emails(contact_emails: work_form.contact_emails_attributes)
+        CocinaGenerators::Description.contact_emails(
+          # Use `#dup` because mutating the passed-in form object would be unexpected
+          contact_emails: work_form.contact_emails_attributes.dup.tap do |contact_emails_attributes|
+            next if work_form.deposits_contact_email.blank?
+
+            contact_emails_attributes << ContactEmailForm.new(email: work_form.deposits_contact_email)
+          end
+        )
```
mjgiarlo, line 72: When mapping to Cocina, append the collection contact email to the list of work contact emails.

app/services/to_collection_form/mapper.rb
```diff
@@ -45,7 +45,8 @@ def params
         provided_custom_rights_statement: collection.provided_custom_rights_statement,
         custom_rights_statement_instructions: collection.custom_rights_statement_instructions,
         work_type: collection.work_type,
-        work_subtypes: collection.work_subtypes
+        work_subtypes: collection.work_subtypes,
+        deposits_contact_email: collection.deposits_contact_email
```
mjgiarlo, line 49: Map DB-persisted values back into the collection form so the user may edit these values.

app/services/to_form/contact_emails_mapper.rb
```diff
@@ -4,15 +4,25 @@ module ToForm
...
+        next if access_contact.value.blank? || access_contact.value == deposits_contact_email
 
-        { 'email' => access_contact[:value] }
+        { 'email' => access_contact.value }
       end.presence
     end
+
+    private
+
+    attr_reader :deposits_contact_email
```
mjgiarlo, line 26: When mapping Cocina into the H3 form representation, keep the work and collection contact emails distinct.

app/views/works/form.html.erb
```diff
@@ -53,6 +53,19 @@
...
+          <%= form.hidden_field :deposits_contact_email %>
+
+          <div class="pane-section">
+            <p>
+              Contact email provided by collection manager <%= render Elements::TooltipComponent.new(tooltip: t('works.edit.fields.deposits_contact_email.tooltip_html')) %>
+            </p>
+            <p>
+              <%= @work_form.deposits_contact_email %>
+            </p>
+          </div>
```
mjgiarlo, line 67: This is the work form UI for collection contact emails: keep the field in the form, but hidden, and let the user know their collection manager has set a contact email for their work that they cannot edit.

app/forms/collection_form.rb
justinlittman: Our pattern for forms has been to group an attribute with all things related to that attribute (validations, etc.). I'm not opposed to changing that pattern, but don't like changing it in just this form.

app/forms/collection_form.rb
justinlittman: Why isn't this just another nested attribute?

app/forms/collection_form.rb
justinlittman: This is also going to cause merge headaches with my PR that also touches this form.

app/forms/work_form.rb
```diff
@@ -31,6 +31,8 @@ def persisted?
 
   attribute :version, :integer, default: 1
 
+  attribute :deposits_contact_email, :string
```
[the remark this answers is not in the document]
justinlittman: Storing these separately is different than the approach used for work types / subtypes. I don't love having different approaches to doing the same thing. (I'm less concerned about the approach.)

justinlittman (commented):
Reviewing this made me realize a requirement that has been unstated but is highly likely: the ability for a user to re-order the contacts (like they can reorder the required / user entered contributors) for a work. This isn't really an issue now where the user enters all of the contact emails.

The approach in this PR would make add re-ordering difficult (whereas it comes for free if the required emails are just part of the work form nested contact emails).

app/controllers/works_controller.rb
mjgiarlo: Will do.

app/forms/collection_form.rb
mjgiarlo: I'll revert the form organization changes and may or may not consider making them again in a different PR later

app/forms/collection_form.rb
mjgiarlo: Why would it be a nested attr? The design seems to indicate that this is a single, simple field.

justinlittman: Oh, I totally missed that. Reconsidering.

app/forms/work_form.rb
```diff
@@ -31,6 +31,11 @@ def persisted?
 
   attribute :version, :integer, default: 1
 
+  attribute :works_contact_email, :string
+  validates :works_contact_email, format: {
```
justinlittman, line 35: Could this ever be invalid?

app/services/to_work_form/mapper.rb
```diff
@@ -31,24 +31,33 @@ def params # rubocop:disable Metrics/AbcSize
...
         agree_to_terms:,
+        works_contact_email:,
         whats_changing: version_description
       }.merge(ToWorkForm::WorkTypeMapper.call(cocina_object:))
         .merge(ToWorkForm::ReleaseDateMapper.call(cocina_object:))
         .merge(ToWorkForm::CreationDateMapper.call(cocina_object:))
     end
 
+    def works_contact_email
+      Collection.find_by(druid: collection_druid)&.works_contact_email
```
justinlittman, line 54: To be consistent with other mappers, this should be passed in instead of DB access within the mapper.

config/locales/en.yml
```diff
@@ -263,13 +263,16 @@ en:
           orcid_name:
             tooltip_html: >-
               This is the name we retrieved from ORCID and that will be included in the detailed information about this deposit. Editing this name will change it for this deposit only and will not alter your ORCID record.
+        works_contact_email:
+          tooltip_html: >-
+            The collection manager has provided an email address for all works deposited into this collection. Depositors can not change or remove this email address. It will be publicly visible and is intended to be a place where people can ask questions about the work you have deposited.
```
justinlittman, line 268: "cannot"?

config/locales/en.yml
```diff
@@ -361,14 +364,18 @@ en:
...
-          help_text: >
+          help_text: >-
             Enter a summary statement (max 600 words) describing the collection that you are creating.
             A detailed statement may improve the discovery of your collection in web searches.
           tooltip_html: >-
             Enter an approximately 2-3 sentence description of the works to be included in this collection. You may want to include information on who the creators of the works are/will be, general subject matter of the works, types of works (technical reports, datasets, video archives, etc), and the scholarly value of the content.
+        works_contact_email:
+          label: 'Contact email to be included in all deposits in this collection'
+          tooltip_html: >-
+            Enter one or more valid email addresses (one per box) that will be included for each deposit to this collection. Depositors will not be able to change or remove these emails. These email addresses will be publicly visible and are intended to be a place where people can ask questions about the content deposited.
```
justinlittman, line 375: I thought there is only one email?

amyehodge:
@justinlittman Yes, it does need to be added to the collection show page. 
@astridu do we have a mockup for that?

mjgiarlo:
@amyehodge No, that was not included in the designs for #992. Happy to work on that when it's ready. Or we could ship this and iterate.

config/locales/en.yml
mjgiarlo: Rebase casualty, sigh.

amyehodge:
@mjgiarlo I think you should ship this when it's ready and let's do the collection show page separately.

justinlittman:
We should still have system tests for the effects on the works.

mjgiarlo:
@justinlittman 💬 
> We should still have system tests for the effects on the works.

I'd like to add that as follow-up work too. Trying to land big PRs ASAP while we're trying to wrap this work cycle up.

justinlittman (approved):
As long as we complete this soonishly.

mjgiarlo:
@justinlittman 💬 
> As long as we complete this soonishly.

Definitely. Turning to that as soon as I click merge. 🍻

justinlittman:
That's what I figure @mjgiarlo. I've learned that you and I have different philosophies on this with exactly the same outcome.

app/forms/work_form.rb
mjgiarlo: With the current design, I don't think so? Unless someone patches in bad data on the Cocina side? Is that possible?

justinlittman: Then the validation should be in cocina. If the validation failed, then the user couldn't fix it. Also, I think the form would fail validation but the user wouldn't be told why it failed, so I really don't think we want this.

Issue #1061, opened 2025-03-12
Indicate on file upload page when each file was uploaded


amyehodge:
This should display a date and time for each file of when it was uploaded.

astridu:
On the manage files page, here are the measurements for the table. (please note the leading on the time/date is a little tight bc it's a quick and dirty mockup, but it shouldn't be that tight) 

We'd like to display a separate column named 'Uploaded' and then display the time and date: hh:mm:ss dd/mm/yy

<img width="876" alt="Image" src="https://github.com/user-attachments/assets/963732b2-4d16-4a4f-808e-3f85420e3b50" />

astridu:
Could we also display the uploaded time/date on the Item view page?

<img width="1174" alt="Image" src="https://github.com/user-attachments/assets/eab8b981-a1ab-41e0-b5bf-c69f3044692a" />

justinlittman:
What does uploaded mean on the item view page? Isn't uploaded an attribute of a file, not a work?

amyehodge:
@astridu when we talked about displaying the 'uploaded' info on the item view page, I meant for each file, so that if a user wanted to check this info for a file they wouldn't need to open the form to do that. @justinlittman is right that this is not about the work. If we can display this info, it needs to be next to each file name like it is in the file upload form.

astridu:
Sorry, my bad. I'll update that screenshot above.

amyehodge:
@astridu From standup today, we aren't going to be able to display this info, so you should hold off on this until we see what @aaron-collier comes up with to do for this.

PR #1084, opened 2025-03-14
[SPIKE] Show updated time for files

aaron-collier:
This is a spike on one possible approach to #1061. There is a bit of a challenge in showing meaningful date information with regard to files on a work in H3 as we do not persist them in the database permanently, they are transient for display on the work form.

This approach sets the `content_file.updated_at` value to the `updated_at` value of the work. This at least allows the files table to display relatively accurate date content for a file based on the last time the work was deposited or edited.

Then when uploading new files, the `updated_at` value is accurate to the time the individual file record is created during the work edit session.

The attached screen recording shows that an existing file is displaying a date from yesterday, and then the new file gets the current date and time... And then uploading a file with the same name as the existing file updates it's time to show that it was altered.

This may be an incomplete approach at this time, but shows how we may be able to provide meaningful/helpful information in this way. 

https://github.com/user-attachments/assets/ccf09b2b-8c56-4a40-ac4f-c5cd9b93f506

### DIFF
--- a/app/components/works/edit/content_file_hierarchy_component.html.erb
+++ b/app/components/works/edit/content_file_hierarchy_component.html.erb
@@ -8,6 +8,9 @@
 
 <% component.with_row_leaf(level: path_parts.length + 1, label: filename) do |row_component| %>
   <% row_component.with_cell do %>
+    <%= render Works::Edit::ContentFileUploadedTimeComponent.new(content_file:) %>
+  <% end %>
+  <% row_component.with_cell do %>
     <%= render Works::Edit::ContentFileDescriptionComponent.new(content_file:) %>
   <% end %>
   <% row_component.with_cell do %>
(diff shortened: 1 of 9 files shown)


justinlittman:
I'd suggest that claiming that the time a file was uploaded is the last time the work was updated is likely to be confusing.

aaron-collier:
Yeah, agree. I'm thinking along the lines of using something at the work level to identify files loaded during rendering of the show page and only display the time for files uploaded during this fit session. 

I'm not sure that wouldn't be confusing as well though to only show the time for some files... pondering...

aaron-collier:
I'm thinking if we don't have a file timestamp in cocina we can't really do this without some confusion.

justinlittman:
One option is to not show a timestamp for the newly uploaded files, just an indicator that it is newly uploaded.

image

aaron-collier:
Yeah, I like that. It avoids needing to do anything in the builder when loading from cocina and just adding an indicator on upload.

PR #1086, opened 2025-03-17
Make required subtypes more accessible.

### LINKED ISSUE #1057: Accessiblity of user view when collection manager has selected work/subtype
The color contrast of the text needs to be bumped up to 6E7272. Also, how does the screen reader go through the first section? Does it say which one is selected and that the section is disabled?

Image

justinlittman:
closes #1057

image


app/components/elements/forms/input_collection_checkboxes_component.html.erb
```diff
@@ -2,8 +2,8 @@
   <%= tag.div class: div_classes do %>
 
     <% disabled = disabled_values.include?(builder.object) %>
-    <%= builder.check_box(class: 'form-check-input', data: input_data, disabled:, readonly: disabled) %>
-    <%= builder.label(class: 'form-check-label') %>
+    <%= builder.check_box(class: disabled ? disabled_checkbox_classes : checkbox_classes, data: input_data, disabled:, readonly: disabled) %>
+    <%= builder.label(class: disabled ? disabled_label_classes : label_classes) %>
```
mjgiarlo, line 6: What would you think about moving this logic to the Ruby class within the component? I.e., having the Ruby class check `disabled` and set the various classes accordingly?

mjgiarlo (approved):
Approve pending consideration of component logic comment, non-blocking

mjgiarlo (approved):

PR #1098, opened 2025-03-17
Add works contact email to collection show page

### LINKED ISSUE #1097: Display required default email address on collection show page
@astridu Can you mock this up?

mjgiarlo:
Fixes #1097

Include also some code review bits that I left out of #1080.

### DIFF
--- a/app/controllers/admin/move_controller.rb
+++ b/app/controllers/admin/move_controller.rb
@@ -43,7 +43,8 @@ module Admin
       @work_form ||= ToWorkForm::Mapper.call(cocina_object:,
                                              doi_assigned: DoiAssignedService.call(cocina_object:, work:),
                                              agree_to_terms: current_user.agree_to_terms?,
-                                             version_description: 'Moving collection')
+                                             version_description: 'Moving collection',
+                                             collection: work.collection)
     end
 
     def version_status
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -2,7 +2,7 @@
 
 # Controller for a Work
 class WorksController < ApplicationController # rubocop:disable Metrics/ClassLength
-  before_action :set_work, only: %i[show edit update destroy review]
+  before_action :set_work_and_collection, only: %i[show edit update destroy review]
   before_action :check_deposit_registering_or_updating, only: %i[show edit]
   before_action :set_status, only: %i[show edit destroy review update]
   before_action :set_work_form_from_cocina, only: %i[show edit review]
@@ -137,7 +137,7 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
     work_params.merge(druid: params[:druid])
   end
 
-  def set_work
+  def set_work_and_collection
     @work = Work.find_by!(druid: params[:druid])
     @collection = @work.collection
   end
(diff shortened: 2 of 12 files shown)


app/controllers/admin/move_controller.rb
```diff
@@ -43,7 +43,8 @@ def work_form
       @work_form ||= ToWorkForm::Mapper.call(cocina_object:,
                                              doi_assigned: DoiAssignedService.call(cocina_object:, work:),
                                              agree_to_terms: current_user.agree_to_terms?,
-                                             version_description: 'Moving collection')
+                                             version_description: 'Moving collection',
+                                             collection: work.collection)
```
mjgiarlo, line 47: Instead of coupling the work form mapper to the DB, inject the collection where the mapper is called (controllers, primarily).

app/controllers/works_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 # Controller for a Work
 class WorksController < ApplicationController # rubocop:disable Metrics/ClassLength
-  before_action :set_work, only: %i[show edit update destroy review]
+  before_action :set_work_and_collection, only: %i[show edit update destroy review]
```
mjgiarlo, line 5: I was surprised that `set_work` was setting a collection ivar as well, so I thought this would be a painless tweak.

db/structure.sql
```diff
@@ -816,3 +816,4 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20241115001126'),
 ('20241111223829'),
 ('20241106143736');
+
```
mjgiarlo, line 819: Rails did this.

app/controllers/admin/move_controller.rb
justinlittman: With work types / subtypes, the required values are only set on the NEW work. The advantage of that approach is that the collection isn't always needed to perform mapping. Are you sure that it is worth the overhead of needing the collection to map?

justinlittman (approved):

app/controllers/admin/move_controller.rb
mjgiarlo: @justinlittman I'll yank it out and see how much trouble it causes.

mjgiarlo: @justinlittman Tried yanking it and things break, understandably so. With the designs for collection contact email, the new form *and* the edit form need to make clear which contact emails are under the depositor's control and which is set by the collection manager, if any. They are all mapped as regular ol' contact emails in Cocina-land. When mapping back from Cocina in an edit scenario, we need a mechanism to be able to tell which are associated with the work and which come from the collection. How can I do that without asking the collection?

justinlittman: For work types / subtypes, since this is needed by components for rendering, determining which are required and which aren't are handled in the component: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/components/works/edit/work_type_component.rb#L27-L29

mjgiarlo: @justinlittman OK, I see. The approach with collection contact emails is different. At a high level:

* When going from form to cocina, multiplex what are separate attributes of the form (`works_contact_email` and `contact_emails_attributes`) into Cocina contact emails on the work.
* When going from cocina to form, de-multiplex the Cocina contact emails into the separate form attributes so we can disambiguate them in the UI.

It'd be surprising (to me, anyway) to replace only one of those by shifting responsibility to the component level. And I'm not sure it's worth reworking the whole approach at this point. I am ready to declare collection contact emails done enough for this work cycle, personally. 😆 (However if you feel more strongly about that and want to toss a follow-up refactor at this, though, please have at it!)

justinlittman: I noticed that this required injecting collections and raised alternative for consideration. It was considered, so moving on ...

PR #1109, opened 2025-03-18
[SPIKE] A collection of refactoring ideas for how we might promote service classes

mjgiarlo:
- **Extract model sync classes to app/synchronizers/**
- **Extract cocina generator classes to app/generators/**
- **Extract importer classes to app/importers/**
- **Extract subscription action classes to app/messengers/**
- **Extract roundtrippers to app/roundtrippers/**
- **Extract collection form builders to app/builders/**
- **Extract work form mapper to app/builders/**
- **Finish extracting form builders to app/builders/ and reconcile name conflicts by explicitly naming builders**
- **Extract Cocina mappers to app/mappers/**


config/initializers/autoloading.rb
```diff
@@ -0,0 +1,30 @@
...
+# Import hash data (froma JSON export) to DB-persisted models
+module Importers; end
+# Map Cocina objects to form objects
+module Mappers; end
+# Send messages to the expected recipients at the expected time
+module Messengers; end
+# Roundtrip Cocina objects to ensure they can be converted to a form instance and back without lossiness
+module Roundtrippers; end
+# Synchronize Cocina to DB-persisted models
+module Synchronizers; end
```
mjgiarlo, line 18: If we want to take any ideas forward at some point, I think we'd probably want to amend the README to include this information and make clear what these entities are in H3.

mjgiarlo:
@thatbudakguy 💬 
> I'm broadly in favor of moving this direction, but I do have one question as someone who still considers themselves "new to rails": is this what `lib/` is for?

This may unfortunately be one of those questions where every developer will give you a different answer. So it's a good question. 🙂 Here's my answer anyway: I tend to use `lib/` for library code that an app depends upon but isn't necessarily core to the app---code that you can reasonably imagine extracting to a gem and re-using in two or three or more other codebases.

If you do a deep dive into how Rails handles auto-loading and reloading of code, `lib/` is handled differently from `app/`. This is the default behavior... and of course there are knobs and dials to tweak Rails' loading defaults. It's a question of what approach feels more like the Rails happy path.
 
> That is, these all feel like well-encapsulated and specific concepts that are nevertheless specific to H3, that also don't fit into the regular MVC hierarchy.

👍🏻 
 
> Ultimately it probably matters little where they live, but I think as someone coming to the app for the first time, I expect to see a few common things in `app/`, but not much in the way of custom logic.

I'm afraid you'd also find many of the other Infra-side apps surprising! Argo has components, forms, presenters, search builders, and services. H2 has many of the same and adds policies and validators. At the same time, you are 💯 right that this approach doubles down on that situation in an attempt to extract like things out of `app/services/`.
 
> If I saw lots of references to `lib/` and looked there, and saw such things as `builders` and `mappers`, I'd think "oh, these are just h3-specific concepts".
> 
> (Also evidently `app/lib/` is a thing, and that would feel just as appropriate to me)

And in fact some Infra apps do use `app/lib/`, e.g. the google-books and pre-assembly apps. Once you start putting code in both `app/lib/` and `app/services/`, you have a question to answer: what code goes in one vs the other?

The change here sidesteps that question and confidently asserts that building form objects from Cocina is a real thing in H3, so let's treat them as a first-class entity in the domain model. Those go in `app/builders/`. Generating Cocina from scalars is a real thing, so let's keep them together.

And so on and so on. That way, it's clear whether a thing goes in `app/mappers/` or can comfortably reside in the `app/services/` soup pot.

app/generators/description.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module CocinaGenerators
+module Generators
```
justinlittman, line 3: `CocinaGenerators` is clearer to me then `Generators` since they generate Cocina.

app/messengers/collection_deposit_persist_completed.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module SubscriptionActions
+module Messengers
```
justinlittman, line 3: `Messengers` doesn't mean anything to me.

app/builders/collection.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-module ToCollectionForm
-  # Maps Cocina Collection to CollectionForm
-  class Mapper
+module Builders
```
justinlittman: `Builders` doesn't make sense to me, since it is a mapping from one specific thing to another. My expectation is that a builder helps you build things (like XML documents).

Also, that this is mapping to a form is lost in this naming.

app/builders/work_builder/work_type_builder.rb
```diff
@@ -16,7 +16,7 @@ def call
     # @return [Cocina::Models::DescriptiveValue] descriptive value for Stanford self-deposit resource types
     def self_deposit_form
       @self_deposit_form ||= cocina_object.description.form.find do |form|
-        form&.source&.value == ToCocina::Work::TypesMapper::RESOURCE_TYPE_SOURCE_LABEL
+        form&.source&.value == WorkMapper::TypesMapper::RESOURCE_TYPE_SOURCE_LABEL
```
justinlittman, line 19: Now I'm totally confused what's a mapper and what's a builder.

thatbudakguy:
> The change here sidesteps that question and confidently asserts that building form objects from Cocina is a real thing in H3, so let's treat them as a first-class entity in the domain model. Those go in `app/builders/`. Generating Cocina from scalars is a real thing, so let's keep them together.

This is a strong statement! I'm a fan of that. I think for some reason I had the idea that the top-level directories inside `app/` are (conventionally) only for classes that correspond directly to core rails libs (models, controllers, jobs, mailers) or are brought in by one of the other dependencies of the app.

Of course, "services" isn't one of those things, but for whatever reason it feels like a core MVC thing to me – specifically, a service is a singleton that wraps access to something that multiple controllers need. That definition is way less specific than than the categories established by this PR, though, and probably less useful as well.

I think @justinlittman's comments are pointing to the fact that, if we did want to elevate these concepts to "first-class status", we probably want to explicitly establish what each of them means (as you did in `autoloading.rb`, but also maybe in the main README – again, as you mention doing).

lwrubel:
`app/services` has definitely grown a lot. Maybe there are Rails conventions I haven't quite absorbed yet to form a clear opinion on wha't best here. 

While `services` now holds code for varied purposes/usage, the organization there does give me some hints of the domain for that code. For example, if it's related to cocina generation or a collection form, or sdr integrations. I would need to understand better `builders` vs 'mappers` vs `synchronizers` to know where to look for the classes now grouped there. 

That said, there is some organization in `services` that could be reconsidered. Looking at the code for the first time in a few weeks, I don't quite understand, for example, `to_form` vs  `to_work_form` and `to_collection_form`.

justinlittman:
This is what I take a builder to be: https://nokogiri.org/rdoc/Nokogiri/XML/Builder.html

And just to remove some barriers, my primary concerns are with naming choices, not `app/XXX` vs. `app/services/XXX`.

app/generators/description.rb
mjgiarlo: Totally fine by me.

app/messengers/collection_deposit_persist_completed.rb
mjgiarlo: I feel like whatever these are, they're a novel concept in our work. Messengers is a blank slate and IMO so it SubscriptionActions. I'm looking for a noun that conveys an actor whose sole job is to reliably delivery messages. That's why I chose Messengers but there could be a dozen other nouns that are clearer.

app/builders/collection.rb
mjgiarlo: I was reluctant to use `FormBuilders` because those are already a Rails built-in thing. I do like the idea of conveying forminess with this name since that's the output. I was partial to builders only because we're building up the form object as we pick out the bits of Cocina that correspond to our forms, but I don't love the name. (I'm not wedded to any of the names in this PR. In fact, if we're actively discussing what's the best name for any of these, the PR is already a success.) Form factories? Form mappers? 

Whatever choices we make, if we decide to take this stuff forward, I'm committing to documenting in the README so that we can make sense of this in the months and years to come.

app/builders/work_builder/work_type_builder.rb
mjgiarlo: A builder builds a form object. A mapper maps to Cocina. Very arbitrary and wholly changeable!

mjgiarlo:
@justinlittman 💬 
> This is what I take a builder to be: [nokogiri.org/rdoc/Nokogiri/XML/Builder.html](https://nokogiri.org/rdoc/Nokogiri/XML/Builder.html)

Totally reasonable, IMO, since we broad use of Nokogiri.
 
> And just to remove some barriers, my primary concerns are with naming choices, not `app/XXX` vs. `app/services/XXX`.

Good to know! And FWIW, I'm not seeing any barriers at the moment. This is all good discussion. For me at least, this is way preferable over me grumbling vaguely about services in a retro. 🙂

mjgiarlo:
Discussed today and I will take another hack at this after the current H3 work cycle ends so as not to cause unnecessary rebase pain. We also arrived at the following names for the seven concerns:

* Input: DB-persisted models; Output: email notifications to expected recipients (`SubscriptionMailers`)
* Input: hash data (from an H2 JSON export); Output: DB-persisted models (`Importers`)
* Input: Cocina objects & DB-persisted values; Output: Form instances (`FormMappers`)
* Input: form instance & Cocina object; Output: boolean (`Roundtrippers`)
* Input: Cocina objects & DB-persisted values; Output: DB-persisted models (`ModelSynchronizers`)
* Input: form instance; Output: Cocina models (`CocinaMappers`)
* Input: Ruby primitives (strings, dates, arrays, hashes); Output: Ruby primitives in Cocina shapes (`CocinaBuilders`)

mjgiarlo:
All of this work has now been done in subsequent PRs. Closing.

PR #1121, opened 2025-03-18
Support adding aria-required to some fields

### LINKED ISSUE #909: [SODA] Form field validation (header)

### LINKED ISSUE #928: Indicate required fields
Currently there is no indication on the collection form which fields are required to save or the work form which fields are required to save OR deposit.

Per SODA:
> Fields that are required are not indicated as being required visually or programmatically.

thatbudakguy:
Fixes #928
Fixes #909 

This PR adds the option to several form components to pass `mark_required`, which adds `aria-required` to the component's corresponding input(s). Note that this is distinct from `required`, since we do not want the browser default behavior that will cause the form to not be able to submit.

This is then used to mark all of the fields that are actually required for submission of the work and collection forms.

For cases like the license selector, the Stimulus controller was updated so that the `<select>` associated with each radio button is marked required if that radio button is selected, since it's invalid to pick an option but not also choose a license.

This PR does not result in any visual changes to the app.

### DIFF
--- a/app/components/edit/label_component.rb
+++ b/app/components/edit/label_component.rb
@@ -3,15 +3,16 @@
 module Edit
   # Component for rendering a form label outside of a form field.
   class LabelComponent < ApplicationComponent
-    def initialize(label_text:, hidden_label: false, classes: [], required: false)
+    def initialize(label_text:, hidden_label: false, classes: [])
       @label_text = label_text
       @hidden_label = hidden_label
       @classes = classes
-      @required = required
       super()
     end
 
-    attr_reader :required
+    def call
+      tag.label(@label_text, class: classes)
+    end
 
     def classes
       merge_classes(%w[form-label fw-bold], @hidden_label ? 'visually-hidden' : nil, @classes)
--- a/app/components/edit/label_component.rb
+++ b/app/components/edit/label_component.rb
@@ -3,16 +3,15 @@
 module Edit
   # Component for rendering a form label outside of a form field.
   class LabelComponent < ApplicationComponent
-    def initialize(label_text:, hidden_label: false, classes: [])
+    def initialize(label_text:, hidden_label: false, classes: [], required: false)
       @label_text = label_text
       @hidden_label = hidden_label
       @classes = classes
+      @required = required
       super()
     end
 
-    def call
-      tag.label(@label_text, class: classes)
-    end
+    attr_reader :required
 
     def classes
       merge_classes(%w[form-label fw-bold], @hidden_label ? 'visually-hidden' : nil, @classes)
--- a/app/components/elements/forms/field_component.rb
+++ b/app/components/elements/forms/field_component.rb
@@ -7,10 +7,11 @@ module Elements
       def initialize(form:, field_name:, required: false, hidden_label: false, label: nil, help_text: nil, # rubocop:disable Metrics/ParameterLists, Metrics/MethodLength, Metrics/AbcSize
                      disabled: false, hidden: false, data: {}, input_data: {}, placeholder: nil, width: nil,
                      label_classes: [], container_classes: [], input_classes: [], tooltip: nil, error_classes: [],
-                     readonly: false)
+                     readonly: false, mark_required: false)
         @form = form
         @field_name = field_name
         @required = required
+        @mark_required = mark_required
         @hidden_label = hidden_label
         @label = label
         @help_text = help_text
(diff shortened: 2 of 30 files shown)


app/components/edit/label_component.rb
```diff
@@ -3,16 +3,15 @@
...
       @label_text = label_text
       @hidden_label = hidden_label
       @classes = classes
+      @required = required
       super()
     end
 
-    def call
-      tag.label(@label_text, class: classes)
-    end
```
mjgiarlo, line 15: It looks like this component changes from an inline component to a template-y one. Can you say more about that change? I'm not sure if we've been 💯 consistent about this, but I do recall early on trying to use inline components if justifiable.

app/components/elements/forms/field_component.rb
```diff
@@ -36,7 +37,22 @@ def help_text_id
...
+
       def field_aria
+        {}.tap do |arias|
+          # Set aria-required if we want to indicate required, but the field
+          # does not actually have a required attribute
+          #
+          # This is used for collection/work forms where we do server-side
+          # validation and don't want to block form submission on empty fields
+          arias[:required] = true if @mark_required
+        end.merge(error_aria)
```
mjgiarlo, line 52: I'm guessing here that we don't want the following:

```suggestion
        error_aria.tap do |arias|
          # Set aria-required if we want to indicate required, but the field
          # does not actually have a required attribute
          #
          # This is used for collection/work forms where we do server-side
          # validation and don't want to block form submission on empty fields
          arias[:required] = true if @mark_required
        end
```

because we want the `error_aria[:required]` to override the `true` here?

app/components/edit/label_component.rb
thatbudakguy: Ah – this is leftover from when I needed to do more actual rendering (to render the `*`) and switched to a template, but is no longer necessary now that visual changes have been removed. I'll revert it.

mjgiarlo (approved):
I have a few relatively minor questions/comments to better understand some parts, but otherwise happy to approve this.

app/components/edit/label_component.rb
```diff
@@ -3,16 +3,15 @@
 module Edit
   # Component for rendering a form label outside of a form field.
   class LabelComponent < ApplicationComponent
-    def initialize(label_text:, hidden_label: false, classes: [])
+    def initialize(label_text:, hidden_label: false, classes: [], required: false)
       @label_text = label_text
       @hidden_label = hidden_label
       @classes = classes
+      @required = required
```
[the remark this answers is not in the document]
thatbudakguy: Good catch – same thing as above; was from when labels could display a required indicator. No longer required...haha.

app/components/elements/forms/field_component.rb
thatbudakguy: This bit was originally written by @aaron-collier I think, but I think your suggestion is somewhat clearer – I don't think `error_aria` ever cares about `required`, so no need to start from an empty hash.

mjgiarlo: :shipit:

app/components/edit/label_component.rb
thatbudakguy: Removed.

app/components/edit/label_component.rb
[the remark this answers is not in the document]
thatbudakguy: Removed.

app/components/elements/forms/field_component.rb
```diff
@@ -36,7 +37,22 @@ def help_text_id
         @help_text_id ||= form.field_id(field_name, 'help')
       end
 
+      def mark_required?
+        @mark_required || required
```
mjgiarlo, line 41: @thatbudakguy This looks like the line causing CodeCov to bark at us.

PR #1137, opened 2025-03-19
Allow manager/reviewer to deposit rejected work with no changes

### LINKED ISSUE #1134: Collection manager or reviewer should be able to submit a work returned for revision
When working on #1041, the solution we went with was under the condition that the reviewer be able to submit the work without changes. That's not working.

Here's what currently happens:

1. Depositor submitted work for review
2. Manager returned the work
3. Manager then opened the work to edit, did not make any changes, and then clicked the button to deposit.
4. The view page reloaded and the work still says "Returned." No buttons display on the page for the depositor or the manager so there's no way to get it deposited.

The work we were testing on is https://sul-h3-stage.stanford.edu/works/druid:zk221sc9318

aaron-collier:
Fixes #1134

### DIFF
--- a/app/jobs/deposit_work_job.rb
+++ b/app/jobs/deposit_work_job.rb
@@ -34,6 +34,8 @@ class DepositWorkJob < ApplicationJob
     if accession?(new_cocina_object:)
       work.accession!
       Sdr::Repository.accession(druid:)
+      # If a collection manager or reviewer has rejected a previous review, we need to approve the work again
+      work.approve! if work.rejected_review?
     else
       work.deposit_persist_complete!
     end
--- a/spec/jobs/deposit_work_job_spec.rb
+++ b/spec/jobs/deposit_work_job_spec.rb
@@ -189,4 +189,27 @@ RSpec.describe DepositWorkJob do
       expect(Sdr::Repository).not_to have_received(:accession)
     end
   end
+
+  context 'when review rejected but a manager or reviewer deposits' do
+    let(:druid) { druid_fixture }
+    let(:cocina_object) { dro_with_metadata_fixture }
+    let(:work_form) { WorkForm.new(druid:, content_id: content.id) }
+    let(:work) { create(:work, :rejected_review, druid:, collection:) }
+    let(:version_status) { build(:draft_version_status) }
+    let(:collection) { create(:collection, druid: collection_druid_fixture, reviewers: [current_user]) }
+
+    before do
+      allow(Sdr::Repository).to receive(:status).and_return(version_status)
+      allow(ToCocina::Work::Mapper).to receive(:call).and_return(cocina_object)
+      allow(RoundtripSupport).to receive(:changed?).and_return(false)
+    end
+
+    it 'approves the work and deposits' do
+      expect(work.rejected_review?).to be true
+      described_class.perform_now(work_form:, work:, deposit: true, request_review: false, current_user:)
+
+      expect(work.review_not_in_progress?).to be true
+      expect(Sdr::Repository).to have_received(:accession)
+    end
+  end
 end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -238,10 +238,14 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
     nil
   end
 
+  def initial_review?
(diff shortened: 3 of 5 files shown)


justinlittman (commented):
This is something we've had trouble getting right. Can you add a test?

justinlittman (changes requested):
I don't think this change accomplishes what the ticket is asking for. The change affects when a work is marking as pending review. The ticket is about a deposit being performed.

aaron-collier:
@justinlittman I'm open to suggestion, when I test this manually it allows the deposit to put the work back to pending_review, without it when saving with no changes, it keeps the work in rejected.

justinlittman:
The ticket is about performing a deposit, not changing the review status. It might be helpful if @amyehodge clarified the intended behavior for the ticket.

aaron-collier:
Ah, ok... I wonder if there are multiple bugs here. It seems like we talked about the need to be able to kick back a reviewed work without changes which doesn't work without this change (or a similar change), as well as doing the deposit perhaps.

I'll see if I can setup testing as a manager locally.

justinlittman:
At one point we had talked about kicking back a reviewed work, but I think we determined that that wasn't necessary if it could be deposited by the reviewer.

aaron-collier:
Hmm... Ok, I'll keep poking.

amyehodge:
@justinlittman is correct above. That was what we discussed.

spec/jobs/deposit_work_job_spec.rb
```diff
@@ -189,4 +189,28 @@
...
+
+    before do
+      allow(Sdr::Repository).to receive(:status).and_return(version_status)
+      allow(ToCocina::Work::Mapper).to receive(:call).and_return(cocina_object)
+      allow(RoundtripSupport).to receive(:changed?).and_return(false)
+    end
+
+    it 'approves the work and deposits' do
+      expect(work.rejected_review?).to be true
+      work.approve!
```
justinlittman, line 209: Why is `approve!` here?

aaron-collier: Ah, I could remove this from the test I suppose as it's just a bit extra. Since the approve is in the controller and not the job, I put it in here to verify that it happened without the transition error, but maybe including it in the test is a bit confusing?

justinlittman: If the change is in the controller, then this test in of the DepositJob doesn't actually test the fix.

aaron-collier: Should I do a system test for this small change? I'm not sure I entirely agree that this isn't sufficient as it verifies the approval failure doesn't fail after the job, which is what is happening in the controller. But I'm open to suggestion.

justinlittman: Yes, this needs a small system test. Otherwise, it isn't clear that this solved the problem. (That is, the state of work is changed so that it is in a depositable state when passed to the DepositJob.)

The other alternative is to consider moving the change to the DepositJob where it can be tested.

justinlittman: And I'd suggest that it is better to have this sort of logic in the job than the controller.

justinlittman: I think your `if` would cover that.

app/jobs/deposit_work_job.rb
```diff
@@ -36,6 +36,8 @@ def perform(work_form:, work:, deposit:, request_review:, current_user:) # ruboc
       Sdr::Repository.accession(druid:)
     else
       work.deposit_persist_complete!
+      # If a collection manager or reviewer has rejected a previous review, we need to approve the work again
```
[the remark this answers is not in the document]
aaron-collier: Do you think it should be outside of the if? I ended up here when walking through the test with a debugger and was trying to make sure I didn't conflict with check on line 42, but I'm ok moving, this just made the most sense to me.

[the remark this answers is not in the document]
justinlittman: Let's start with the specific case of the ticket: When a collection manager deposits a work that is rejected, the work should be approved. Does that sound right?

justinlittman (approved):

PR #1138, opened 2025-03-19
Clear form validation errors on fields and associated tabs when value changes

mjgiarlo:


app/javascript/controllers/tab_error_controller.js
```diff
@@ -9,8 +9,17 @@ export default class extends Controller {
...
+      const invalidElements = paneElement.querySelectorAll('.is-invalid')
+      if (invalidElements.length > 0) {
+        invalidElements.forEach(invalidElement => {
+          invalidElement.addEventListener('change', event => {
+            const invalidFeedbackElement = paneElement.querySelector(`#${event.currentTarget.id}_error`)
+            event.currentTarget.classList.remove('is-invalid')
+            if (invalidFeedbackElement) invalidFeedbackElement.remove()
+            if (!paneElement.querySelector('.is-invalid')) tabEl.classList.remove('is-invalid')
+          })
+        })
```
mjgiarlo, line 22: For each invalid element in the current pane, attach a change event listener whose responsibility is to:

1. Mark the field itself as no longer invalid
2. Destroy the associated invalid feedback `div`, if present
3. Mark the tab as no longer invalid if there are no remaining invalid elements in the pane

app/javascript/controllers/tab_error_controller.js
```diff
@@ -24,4 +33,20 @@ export default class extends Controller {
...
+      invalidElements.forEach(invalidElement => {
+        if (invalidElement.tagName === 'DIV') {
+          invalidElement.remove()
+        } else {
+          invalidElement.classList.remove('is-invalid')
+        }
+      })
+      tabElement.classList.remove('is-invalid')
+    }
+  }
```
mjgiarlo, line 51: This allows other controllers to specify which pane/tab should be cleared of invalid status due to an action taken in a controller other than this one, e.g., when a file is uploaded via dropzone or when the work type is twiddled.

thatbudakguy (commented):
This works how I expect when testing locally!

Re: failing system specs, the behavior that I see is that the validation doesn't seem to be taking effect in the first place – the "without required types" block in `manage_work_type_spec.rb` ultimately chooses "Music" as its type (after fiddling with all the options a bunch), but it then doesn't select a subtype, even though the helptext indicates at least one subtype is required. It then saves as draft, and we don't get a validation error about the missing subtype (although we do get one about the missing title).

aaron-collier (approved):

PR #1139, opened 2025-03-20
Scope works on collection works page. Fix collection permissions for …

### LINKED ISSUE #1124: Clicking on deposit in collection view of deposit list takes user to H3 homepage
I was on a zoom with Kathy Kerns (kkerns), and in the deposit list for the collection "Welcome to your review-required fake capstone collection", she clicked on the first deposit - "2024 Consolidated AI Inventory", and landed on the H3 Homepage. We were able to replicate this a few times.

Image

justinlittman:
…collection owner.

closes #1124

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -82,7 +82,9 @@ class CollectionsController < ApplicationController
   def works
     authorize! @collection
 
-    @works = @collection.works.order(:title).page(params[:page])
+    @works = authorized_scope(@collection.works, as: :collection,
+                                                 scope_options: { collection: @collection })
+             .order(:title).page(params[:page])
     @work_statuses = Sdr::Repository.statuses(
       druids: @works.where.not(druid: nil).pluck(:druid)
     )
--- a/app/policies/collection_policy.rb
+++ b/app/policies/collection_policy.rb
@@ -9,7 +9,7 @@ class CollectionPolicy < ApplicationPolicy
   end
 
   def manage?
-    collection_manager? || record.user_id == user.id
+    collection_manager?
   end
 
   def new?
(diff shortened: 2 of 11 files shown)


app/controllers/collections_controller.rb
```diff
@@ -82,7 +82,9 @@ def wait
   def works
     authorize! @collection
 
-    @works = @collection.works.order(:title).page(params[:page])
+    @works = authorized_scope(@collection.works, as: :collection, type: :relation,
```
justinlittman, line 85: This is where the scope is set for the collection works.

app/policies/collection_policy.rb
```diff
@@ -9,7 +9,7 @@ def show?
   end
 
   def manage?
-    collection_manager? || record.user_id == user.id
+    collection_manager?
```
justinlittman, line 12: The collection owner shouldn't be a manager.

app/policies/collection_policy.rb
```diff
@@ -24,10 +24,7 @@ def review?
     collection_reviewer? || collection_manager?
   end
 
-  # TODO: Add a rule for collection managers & depositors
```
justinlittman, line 27: cruft

app/controllers/collections_controller.rb
mjgiarlo: @justinlittman Did you find you needed to pass the `type: :relation` kwargs? For some reason I thought that was the default value but I could be mistaken.

app/policies/work_policy.rb
```diff
@@ -21,6 +21,14 @@ def review?
     collection_reviewer? || collection_manager?
   end
 
+  scope_for :relation, :collection do |relation, collection:|
```
mjgiarlo, line 24: Since we're likely to be primarily using AR scopes, this could be shortened to:

```suggestion
  relation_scope(:collection) do |relation, collection:|
```

https://actionpolicy.evilmartians.io/#/scoping?id=active-record-scopes

mjgiarlo (approved):
Approve pending consideration of minor feedback

app/controllers/collections_controller.rb
justinlittman: If I made both of your suggested changes, this worked. Thanks!

PR #1160, opened 2025-03-21
Shows validation alert for nested errors.

### LINKED ISSUE #1155: Nested errors don't cause "Required fields" alert to be shown
Image

To reproduce: Try to save a draft of a work with a title and a contributor with a first name (but not last).

justinlittman:
closes #1155

`valid?` checks nested errors. However, `errors` does not include nested errors. Thus, this PR uses the results of `valid?` to determine whether to show validation message instead of checking for the presence of errors.

### DIFF
--- a/app/views/collections/form.html.erb
+++ b/app/views/collections/form.html.erb
@@ -14,7 +14,7 @@
     <% end %>
   <% end %>
   <%= render Elements::HeadingComponent.new(level: :h1, variant: :h2, text: @collection_form.title || 'Untitled collection', classes: 'mb-3') %>
-  <%= render Elements::AlertComponent.new(variant: :danger, value: t('messages.validation')) if @collection_form.errors.present? %>
+  <%= render Elements::AlertComponent.new(variant: :danger, value: t('messages.validation')) if defined?(@valid) && !@valid %>
   <% discard_draft_form_id = dom_id(@collection_form, 'discard_form') %>
   <%= render Edit::DiscardDraftFormComponent.new(presenter: @collection_presenter, id: discard_draft_form_id) if @collection_presenter %>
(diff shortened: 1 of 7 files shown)


spec/system/validate_work_deposit_spec.rb
```diff
@@ -101,6 +107,12 @@
     # Make the abstract valid
     fill_in('Abstract', with: abstract_fixture)
 
+    # Keywords are marked invalid
+    expect(page).to have_field('Keywords (one per box)', class: 'is-invalid')
```
justinlittman, line 111: Oops. There was a bug in this test -- keywords were never made valid.

app/views/collections/form.html.erb
```diff
@@ -14,7 +14,7 @@
     <% end %>
   <% end %>
   <%= render Elements::HeadingComponent.new(level: :h1, variant: :h2, text: @collection_form.title || 'Untitled collection', classes: 'mb-3') %>
-  <%= render Elements::AlertComponent.new(variant: :danger, value: t('messages.validation')) if @collection_form.errors.present? %>
```
mjgiarlo, line 17: I wonder if we might want to ticket investigating how our nested attrs implemention could make this better. For instance, if there are errors in a nested form (like the contributor form), should the errors be propagated to the top-level form (the work or collection)? If this is how AR errors would work with nested attrs (I can't recall), it may be worth the time the next time we touch H3.

mjgiarlo (approved):
LGTM, good catch!

app/views/collections/form.html.erb
justinlittman: In spent some time on this and came to the conclusion that this is the same behavior as AR. However, I agree it could take more investigation because this feels awkward.

Issue #898, opened 2025-03-03
[SODA] Tooltip content not accessible by screen readers.


aaron-collier:
Tooltip content is not accessible to screen readers. The presence of the tooltip is not conveyed to  screen reader users and it cannot be accessed by keyboard users.

SODA Recommendation:

```
RULE : 
The name, role, value, states, and properties of user interface components MUST be programmatically determinable by assistive technologies.

HOW TO FIX: 
A tooltip is a popup that displays information related to an element. There are several common patterns for tooltips including tooltips that appear on mouse hover or keyboard focus of the tooltip trigger and tooltips that appear on activation of the tooltip trigger. Tooltips may contain only static text or they may contain functional elements.

Fix this issue by ensuring all of the following are true:
1. The tooltip trigger must be focusable and operable with a keyboard.
2. If the tooltip content contains functional elements - such as links, buttons or form fields - they must be focusable and operable with a keyboard.
3. The content of the tooltip must be readable with a screen reader.
4. The keyboard focus (Tab key) order and screen reader reading order must be logical. The simplest way to ensure this is to insert the tooltip content immediately after the tooltip trigger in the DOM.

REFERENCE: 
Deque University: https://dequeuniversity.com/class/custom-widgets/examples/tooltip
Deque University: https://dequeuniversity.com/class/custom-widgets/examples/tooltip-dialog
WAI-ARIA Authoring Practices: https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

BACKGROUND: 
Every user interface control must have a role along with any applicable states and properties so that screen reader users know how to interact with the control. Native HTML elements - such as <button>, <a>, <input>, <select> - already have a role, and their necessary states and properties - such as the checked/unchecked state of a checkbox - are automatically conveyed so nothing more needs to be done. If you create a custom version of a native HTML element or a custom control or widget that does not have a native HTML equivalent, you must add the relevant role(s) and any applicable states and properties using ARIA as well as expected keyboard interactions.
```

See Issue ID 2252 in https://docs.google.com/spreadsheets/d/1x0uBBDe3T8O0HNCs74OdVKQc8tQQNvOV_f-C6IS44IQ/edit?gid=1050712151#gid=1050712151

thatbudakguy:
After investigating with a screen reader, I've found that it does not read the content of the tooltip, and the SR itself does not focus the tooltip when moving through the page (it is only focusable via tab).

[...]

PR #1172, opened 2025-03-21
Change tooltips to buttons with generated accessible labels

thatbudakguy:
SODA identified four issues with our tooltips:
1. The tooltip trigger must be focusable and operable with a keyboard.
2. If the tooltip content contains functional elements - such as links, buttons or form fields - they must be focusable and operable with a keyboard.
3. The content of the tooltip must be readable with a screen reader.
4. The keyboard focus (Tab key) order and screen reader reading order must be logical.

This PR fixes 1 and 4 by making the tooltips into buttons, which automatically makes them focusable and tabbable in a logical order. It also ensures a reasonable accessible name for the tooltip is generated by passing in the label of whatever the tooltip is tied to in order to generate a name like "More information on my field".

For 3, I've confirmed that you actually _can_ get the tooltip to be read, it's just a little weird in VoiceOver: you have to wait until the title of the tooltip is read ("More information on my field") and then you get informed that you can press a special key combination for more info, at which point it will open a _new_ menu where you can ask for the tooltip. This is...weird, but at the end of the day you can get at the tooltip text, so I'm calling it OK for now.

The only place we have a problem with 2 is the links inside a tooltip here:
https://github.com/sul-dlss/hungry-hungry-hippo/blob/b7998cd4dff6ca5e4d32d20c30d400a7fbe06820/config/locales/en.yml#L262

...but theoretically an SR user still hears the actual link addresses when the tooltip is read and could just navigate there if they needed to go get an ORCID, so it's maybe not as bad as it could be.

All that being the case, I think this PR can be considered to complete #898.

### DIFF
--- a/app/assets/stylesheets/application.bootstrap.scss
+++ b/app/assets/stylesheets/application.bootstrap.scss
@@ -280,6 +280,8 @@ i.icon-xl {
 }
 
 .tooltip-info {
+  background: none;
+  border: none;
   color: var(--stanford-digital-blue);
 }
 
--- a/app/components/edit/tab_form/pane_component.html.erb
+++ b/app/components/edit/tab_form/pane_component.html.erb
@@ -4,7 +4,7 @@
       <% if label || help_text || help? %>
         <div class="pane-header">
           <% if label %><h2 class="h4 tooltip-header"><%= label %></h2><% end %>
-          <%= render Elements::TooltipComponent.new(tooltip:) %>
+          <%= render Elements::TooltipComponent.new(target_label: label, tooltip:) %>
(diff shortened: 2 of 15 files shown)


thatbudakguy:
Screen from FF accessibility inspector showing generated label for a tooltip button:

Screenshot 2025-03-21 at 13 37 59

mjgiarlo (approved):

peetucket:
Note in case this PR gets looked at again in the future.  This PR https://github.com/sul-dlss/hungry-hungry-hippo/pull/1532 changes the tooltips from a button to an `<a>` tag in order to get them to work in Safari (to address #1519)

PR #1181, opened 2025-03-21
Do not include pending review or review rejected works in drafts list

### LINKED ISSUE #1133: Items submitted for review should not show up on dashboard under "Drafts - please complete"
Only draft works that have not been deposited or submitted for review should show up here. This is confusing some testers.

Please let me know if you need more info.

aaron-collier:
Fixes #1133

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -26,8 +26,14 @@ class DashboardController < ApplicationController
     @draft_works = []
     @status_map = current_user.your_works.to_h do |work|
       version_status = druid_to_status_map[work.druid] || VersionStatus::NilStatus.new
-      @draft_works << work if version_status.draft? && work.user == current_user
+      @draft_works << work if draft?(work, version_status)
       [work.id, version_status]
     end
+  end
+
+  def draft?(work, version_status)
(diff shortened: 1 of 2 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -26,8 +26,14 @@ def set_draft_works_and_status_map
     @draft_works = []
     @status_map = current_user.your_works.to_h do |work|
       version_status = druid_to_status_map[work.druid] || VersionStatus::NilStatus.new
-      @draft_works << work if version_status.draft? && work.user == current_user
+      @draft_works << work if draft?(work, version_status)
```
mjgiarlo, line 29: Before we used to check that the work user was the current user. Is that no longer needed?

mjgiarlo (approved):

app/controllers/dashboard_controller.rb
aaron-collier: Oh - good catch, I think we do still need it, I'll restore. it's not needed for the pending check, but it is for the draft check because otherwise works that aren't yours but are in your collections might show up here. I was going back and forth on some refactoring and missed bringing this back.

PR #1219, opened 2025-03-25
Include returned reviewed object in draft list on dashboard

aaron-collier:
Works that are in a rejected review state should be included on the users dashboard as drafts.


mjgiarlo (approved):
LGTM. How do you feel about the test coverage surrounding this area of code? Is it worth testing in a system spec or request spec? Or is that a fruitless rabbithole?

PR #1259, opened 2025-04-10
Only send new collection once to admins.

### LINKED ISSUE #1247: Required default email should be adequate for email validation
If the collection manager provides a required email address, the user should not be required to enter one. 

I'm currently seeing that with a required email and the email box left blank, I'm getting a validation error.

Image

justinlittman:
closes #1247

The code only sends the email when the version is 1. However, when the collection settings are updated, the version remains 1 since accessioning isn't performed. This moves the subscription to accessioning, which should only be triggered when the version is changed.


app/subscription_mailers/collection_accessioning_started_subscription_mailer.rb
```diff
@@ -1,13 +1,14 @@
...
+# Sends the appropriate emails when accessioning is started for a collection.
+class CollectionAccessioningStartedSubscriptionMailer
   def self.call(...)
     new(...).call
   end
 
   # @param [Work|Collection] object the work or collection that was persisted
-  def initialize(object:)
+  # @param [User] current_user the user that performed the action
+  def initialize(object:, current_user:) # rubocop:disable Lint/UnusedMethodArgument
```
mjgiarlo, line 11: Can you say more about how/where this value is used? Approving in the meantime.

mjgiarlo (approved):

PR #1287, opened 2025-04-23
Fix roles and changes roles to lookup by value.

justinlittman:


app/cocina_builders/contributor_role_cocina_builder.rb
```diff
@@ -8,24 +8,15 @@ class ContributorRoleCocinaBuilder # rubocop:disable Metrics/ClassLength
...
-    },
-    # person roles
-    composer: {
-      value: 'composer',
-      code: 'cmp',
-      uri: 'http://id.loc.gov/vocabulary/relators/cmp',
-      source: SOURCE
-    },
     contributing_author: {
-      value: 'contributing author',
```
justinlittman, line 28: There were a handful of incorrect values fixed here a below. These should correspond with the MARC relator.

app/mappers/form/work_contributors_mapper.rb
```diff
@@ -69,7 +69,9 @@ def organization?
       end
 
       def role
-        contributor.role.first&.value&.tr(' ', '_')
+        @role ||= ContributorRoleCocinaBuilder::ROLES.keys.find do |key|
```
justinlittman, line 72: Looking up by relator value instead of existing hack.

jmartin-sul (approved):

PR #1294, opened 2025-04-25
Encode user-provided keyword strings before sending to FAST API

mjgiarlo:
Do this because it is a best practice and because it eliminates HB exceptions about unhelpful 400 Bad Request responses from the FAST API when we send it improperly escaped query strings.

### DIFF
--- a/spec/support/stub_fast_connection.rb
+++ b/spec/support/stub_fast_connection.rb
@@ -2,8 +2,15 @@
 
 RSpec.shared_context('with FAST connection') do
   before do
-    stub_request(:get, "#{Settings.autocomplete_lookup.url}?query=#{query}#{other_params}")
-      .with(headers: lookup_headers)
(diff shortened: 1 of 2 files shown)


peetucket (approved):
pending CI passing

spec/support/stub_fast_connection.rb
```diff
@@ -2,8 +2,15 @@
...
-      .with(headers: lookup_headers)
+    stub_request(:get, Settings.autocomplete_lookup.url)
+      .with(headers: lookup_headers, query: {
+              query:,
+              rows: 20,
+              sort: 'usage desc',
+              queryIndex: 'suggestall',
+              queryReturn: 'idroot,suggestall,tag',
+              suggest: 'autoSubject'
+            })
```
mjgiarlo, line 13: Avoid double-encoding of params in tests.

PR #1331, opened 2025-05-09
Support globus

justinlittman:


app/jobs/deposit_work_job.rb
```diff
@@ -97,4 +94,36 @@ def accession?(new_cocina_object:)
     # If work already opened, then the work has changes and can be accessioned.
     deposit? && (new_cocina_object || status&.open?)
   end
+
+  def globus?
+    @globus ||= content.content_files.exists?(file_type: 'globus')
+  end
+
+  def endpoint_client_for(path)
+    GlobusClient::Endpoint.new(user_id: user.email_address, path:, notify_email: false)
```
mjgiarlo, line 103: Does this class need to reach into the implementation details of the client or can it be rewritten to use the public interface we've set up on the `GlobusClient` class? I fear this is a slippery slope to the SdrClient situation we're still trying to get out of.

app/jobs/deposit_work_job.rb
```diff
@@ -97,4 +94,36 @@ def accession?(new_cocina_object:)
...
+    new_endpoint_client = endpoint_client_for(GlobusSupport.new_path(user: current_user, with_uploads_directory: true))
+    work_endpoint_client = endpoint_client_for(GlobusSupport.path(work:))
+    # rename if not persisted and globus?
+    if !work_form.persisted? && new_endpoint_client.exists? && !work_endpoint_client.exists?
+      new_endpoint_client.rename(new_path: GlobusSupport.path(work:, with_uploads_directory: true))
+    end
+    # remove permissions
+    work_endpoint_client.delete_access_rule
+  rescue StandardError => e
+    raise unless e.message.include?('Access rule not found')
```
mjgiarlo, line 127: Do we expect `Access rule not found` to happen in certain occasions? Can we document them? Do they need logging or HB alerts?

app/jobs/globus_list_job.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def endpoint_client
+    @endpoint_client ||= if work.present?
+                           GlobusClient::Endpoint.new(user_id: user.email_address, path: GlobusSupport.path(work:),
+                                                      notify_email: false)
+                         else
+                           GlobusClient::Endpoint.new(user_id: user.email_address, path: GlobusSupport.new_path(user:),
+                                                      notify_email: false)
+                         end
```
mjgiarlo, line 51: Same question as above.

app/jobs/deposit_work_job.rb
justinlittman: It is actually a normal occurrence that should just be ignored. This is a terrible way to ignore it, so: https://github.com/sul-dlss/hungry-hungry-hippo/issues/1377

mjgiarlo (approved):

PR #1380, opened 2025-06-02
Validate that there are not too many content files.

### LINKED ISSUE #1374: Reviewers cannot access files for deposits that are pending review
Reviewers cannot access files for deposits that are still pending review. In H2, the work show page has links to download files that have not yet been deposited yet:

> Image

In H3, there are no file links:

> Image

In order to evaluate submissions prior to deposit, reviewers need to be able to access both the files and the metadata.

justinlittman:
closes #1374

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -28,7 +28,8 @@ class WorkForm < ApplicationForm
     self.contact_emails_attributes = (contact_emails_attributes - blank_contact_emails).map(&:attributes)
   end
 
-  validate :content_file_presence, on: :deposit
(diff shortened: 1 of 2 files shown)


app/forms/work_form.rb
```diff
@@ -135,9 +136,10 @@ def persisted?
 
   def content_file_presence
```
mjgiarlo, line 137: Since this method is now handling not only presence of content files but the "too many" case, would you consider renaming this method?

mjgiarlo (approved):
Approve pending consideration of validation method rename

PR #1381, opened 2025-06-02
send sunetid on close event when accessioning

### LINKED ISSUE #1370: version_close event (in DSA eventing) should include "who"
It is currently null. This may be a problem in H3 or DSA.

peetucket:
Fixes #1370 

* Verified that this the correct param name on DSA (and thus with the dor-services-client)
* Updated tests
* Ran an object through localhost and didn't see any errors
 
Have not tested on a server yet though

### DIFF
--- a/app/services/sdr/repository.rb
+++ b/app/services/sdr/repository.rb
@@ -50,11 +50,12 @@ module Sdr
 
     # @param [String] druid the druid of the object
     # @raise [Error] if there is an error initiating accession
-    def self.accession(druid:)
+    def self.accession(druid:, user_name:)
       # Close the version, which will also start accessioning
       # user_versions = mode for handling user versioning.
+      # user_name = the sunetid to use for the operation.
       # Setting to update_if_existing (the default in DSA) for now.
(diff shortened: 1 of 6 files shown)


app/services/sdr/repository.rb
```diff
@@ -50,11 +50,12 @@ def self.register(cocina_object:, assign_doi: false)
 
     # @param [String] druid the druid of the object
     # @raise [Error] if there is an error initiating accession
-    def self.accession(druid:)
+    def self.accession(druid:, user_name:)
```
justinlittman, line 53: Can you add user_name to the method docs?

justinlittman (approved):

app/services/sdr/repository.rb
mjgiarlo: @peetucket AFAIK, it's globally available within a given thread. `Current.user` is set in both jobs touched in this PR.

peetucket: pushed a new commit to do this, and tested in localhost without error, but not sure how to verify (in localhost) that DSA got the expected param

mjgiarlo (approved):

Issue #1372, opened 2025-05-30
Report new H3 events to DSA eventing


justinlittman:
* Submitted for review: include "who"
* Returned by reviewer: include "who" and "comments"
* Owner changed: include "who" (admin that performed change) and "description" ("Changed owner from [sunet1] to [sunet2]")
* Collection settings updated: "changes" (an array of the following:  Release settings modified, Download setting modified, DOI setting modified, License setting modified, Notification settings modified, Review workflow settings modified, Custom terms of use modified, Added depositors: [SUNet ID list], Removed depositors: [SUNet ID list])

The events should indicate that they are H3 specific, e.g., by prefixing the event names with "h3_" or including it in the details.

See https://github.com/sul-dlss/dor-event-client

justinlittman:
Owner changed split into https://github.com/sul-dlss/hungry-hungry-hippo/issues/1392 since currently blocked.

PR #1387, opened 2025-06-03
Report SDR events for reviews.

justinlittman:
refs #1372


app/services/sdr/event.rb
```diff
@@ -0,0 +1,18 @@
...
+    class Error < StandardError; end
+
+    # @param [String] druid
+    # @param [String] type the type of event
+    # @param [Hash] data the event data
+    # # @raise [Error] if there is an error creating the event
+    def self.create(druid:, type:, data:)
+      Dor::Event::Client.create(druid:, type:, data:) if Settings.rabbitmq.enabled
+    rescue StandardError => e
+      raise Error, "Failed to create event for #{druid}: #{e.message}"
```
mjgiarlo, line 15: Is there any other context in `e` that we care about other than the message? E.g., the class or the backtrace? If so, we may want to include that in the re-raised exception.

app/subscription_handlers/collection_accessioning_started_subscription_mailer.rb
mjgiarlo, line 1: How are subscription handlers related to or different from subscription mailers? Can you add a blurb about this to the software architecture section of the README?

app/models/work.rb
```diff
@@ -17,7 +17,7 @@ class Work < ApplicationRecord
     end
 
     before_transition any => :pending_review do |work|
-      Notifier.publish(Notifier::REVIEW_REQUESTED, work:)
+      Notifier.publish(Notifier::REVIEW_REQUESTED, work:, current_user: Current.user)
```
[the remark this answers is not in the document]
mjgiarlo: Ah, I see, it looks like not. Because this method takes a kwarg called `**payload`.

config/initializers/subscriptions.rb
```diff
@@ -45,12 +45,21 @@
   Notifier.subscribe(event_name: Notifier::REVIEW_REQUESTED) do |payload|
     ReviewsMailer.with(**payload).submitted_email.deliver_later
   end
+  Notifier.subscribe(event_name: Notifier::REVIEW_REQUESTED) do |payload|
+    ReviewRequestedEventSubmitter.call(**payload)
+  end
```
mjgiarlo, line 50: Just to confirm: it's preferable to create identical subscriptions that each handle a single concern rather than to create a single subscription that handles multiple concerns? If so, we may want to add a comment somewhere in here.

app/services/sdr/event.rb
justinlittman: I don't think so. In other code bases, we don't even rescue when calling `.create`. And Dor::Event::Client doesn't do any error handling. If we ever actually encounter errors, I think better error handling should be added to the client.

app/subscription_handlers/collection_accessioning_started_subscription_mailer.rb
justinlittman: Updated README.

app/subscription_handlers/review_approved_event_submitter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Submits an SDR event when a review is approved for a work.
+class ReviewApprovedEventSubmitter
```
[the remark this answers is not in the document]
justinlittman: I think it is OK to have 2 types of subscription handlers, mails and event submitters, which are differentiating by naming conventions. In the future, we might have other types of subscription handlers.

config/initializers/subscriptions.rb
justinlittman: I'm not aware of any reason to prefer one over the other. There was already examples of identical subscriptions, so I continued that convention.

app/subscription_handlers/review_approved_event_submitter.rb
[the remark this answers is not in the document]
mjgiarlo: Oh, I see now: you've changed the subscription mailer concern to a broader subscription handler concern, which each one can either be a mailer or an event submitter.

app/subscription_handlers/collection_accessioning_started_subscription_mailer.rb
justinlittman: I hadn't pushed because fixing tests. Pushed now.

PR #1401, opened 2025-06-04
Prefer using GlobusClient singleton API

### LINKED ISSUE #1376: Use GlobusClient public interface
Use GlobusClient::Endpoint methods instead of `GlobusClient::Endpoint` directly.

Per @mjgiarlo:
> Does this class need to reach into the implementation details of the client or can it be rewritten to use the public interface we've set up on the GlobusClient class? I fear this is a slippery slope to the SdrClient situation we're still trying to get out of. And if that's not feasible here, I wonder if we should consider making changes to the GlobusClient interface to accommodate these needs.

There are instances of this in `DepositWorkJob` and `GlobusListJob`.

mjgiarlo:
Fixes #1376


app/jobs/deposit_work_job.rb
```diff
@@ -13,7 +13,6 @@ def perform(work_form:, work:, deposit:, request_review:, current_user:) # ruboc
     @deposit = deposit
     @request_review = request_review
     @status = Sdr::Repository.status(druid: work_form.druid) if work_form.persisted?
-    @current_user = current_user
```
mjgiarlo, line 16: Unused.

app/jobs/deposit_work_job.rb
```diff
@@ -131,16 +126,43 @@ def accession_or_persist_complete(druid:)
...
+      path: GlobusSupport.work_path(work:)
+    )
+  end
+
+  def rename_globus_path
+    GlobusClient.rename(
+      new_path: GlobusSupport.work_path(work:, with_uploads_directory: true),
+      user_id: user.email_address,
+      notify_email: false,
+      path: GlobusSupport.user_path(user: Current.user, with_uploads_directory: true)
```
mjgiarlo, line 157: Without the GlobusSupport method renames, this call would have passed `GlobusSupport#path` as the `new_path` arg, and `GlobusSupport#new_path` as the `path` arg, which 😵

app/jobs/globus_list_job.rb
```diff
@@ -28,27 +28,24 @@ def perform(content:, cancel_check_interval: 100)
...
-  def list_files
-    endpoint_client.list_files.each_with_index do |file_info, index|
+  def list_files # rubocop:disable Metrics/AbcSize
+    GlobusClient.list_files(user_id: user.email_address, path: globus_path, notify_email: false)
+                .each_with_index do |file_info, index|
       # Only check for cancel every 100 files
       break if (index % @cancel_check_interval).zero? && canceled?
 
-      filepath = file_info.name.delete_prefix(globus_path)
+      filepath = Pathname.new(file_info.name).relative_path_from(globus_absolute_path).to_s
```
mjgiarlo, line 37: Defer path manipulation to the `Pathname` library instead of manually string munging.

app/jobs/globus_list_job.rb
```diff
@@ -28,27 +28,24 @@ def perform(content:, cancel_check_interval: 100)
...
-                                                      notify_email: false)
-                         else
-                           GlobusClient::Endpoint.new(user_id: user.email_address, path: GlobusSupport.new_path(user:),
-                                                      notify_email: false)
-                         end
+  def globus_path
+    return GlobusSupport.work_path(work:) if work.present?
+
+    GlobusSupport.user_path(user:)
   end
```
mjgiarlo, line 49: These are mere string/path manipulations, so don't bother memoizing.

justinlittman (approved):

PR #1412, opened 2025-06-05
[DRAFT] Add collection and item recent activity display to admin

aaron-collier:
NOTE: This is clearly not ready for review, I have code cleanup and testing to do... Opening now as draft mostly to view diffs, etc.


app/javascript/controllers/admin_days_limit_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
```
justinlittman, line 3: Can you name this controller more generically, e.g., `form_submit_controller.js`?

justinlittman (commented):
I'd recommend putting the work / collection differences in a separate work presenter and collection presenter. It would be cleaner to just pass around the correct presenter rather than litter the code with "if collection/work then ..." statements.

(Let me know if this suggestion isn't clear -- it's a pattern borrowed from @mjgiarlo.)

app/javascript/controllers/admin_days_limit_controller.js
```diff
@@ -0,0 +1,9 @@
...
+  static targets = ['daysLimit']
```
mjgiarlo, line 4: This target doesn't seem to be used anywhere in the controller. If so, can we dump it? Am I misreading? 😵‍💫

app/components/admin/recent_activity_list_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render Elements::Tables::TableComponent.new(id: 'recent-activity-table', classes: 'mb-5', label:, show_label: false) do |component| %>
```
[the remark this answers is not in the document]
aaron-collier: I think you mean `empty_message` - good call, I missed that param, but clearly provides a better experience.

app/components/admin/recent_activity_list_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render Elements::Tables::TableComponent.new(id: 'recent-activity-table', classes: 'mb-5', label:, show_label: false, empty_message:) do |component| %>
+  <% component.with_headers(headers) %>
+  <% items.each do |item| %>
+    <% component.with_row(values: presenter.new(item).values_for) %>
```
justinlittman, line 4: Can you map to presenters in the controller instead of here? (That's how it is done elsewhere.)

app/components/admin/recent_activity_list_component.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Admin
+  # Component for rendering a table of works on the admin user page.
+  class RecentActivityListComponent < ApplicationComponent
+    def initialize(items:, label:, presenter:, headers:)
+      @items = items
+      @label = label
+      @presenter = presenter
+      @headers = headers
+      @empty_message = "No #{label.downcase} activity for time period selected."
```
justinlittman, line 11: Make this a method instead of an instance variable?

app/controllers/admin/recent_activity_works_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Admin
+  # Controller for displaying recent activity for works
+  class RecentActivityWorksController < Admin::ApplicationController
+    def index
+      authorize!
+
+      @headers = headers
+      @label = label
```
justinlittman, line 10: Why are you passing in label and headers as variables from the controller?

app/views/admin/recent_activity/index.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :breadcrumbs do %>
```
justinlittman, line 1: Is this file used? From which controller?

app/components/admin/recent_activity_list_component.html.erb
aaron-collier: Good point, this is what happens when you try to wrap this up late on a tired day... :-)

PR #1414, opened 2025-06-05
Implement user versions.

### LINKED ISSUE #1382: Allow for creation of user versions, but no choices
We will be implementing user version in a way that does not allow the user to make any choices.
1. If they change one or more files, we will create a new user version.
2. If they do not change any of the files, we will not create a new user version.

The `Deposit` tab will display a single "What's changing" box all the time that the user will be required to fill in whether a new user version is being created or not.

For situation 1 above when one or more files have changed, they will also see a message alerting them that a new version will be created.

Design showing the alert for when files have changed: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1126-3223&t=kTmIXbJJPBWRwmm2-4

Image

justinlittman:
closes #1382

### DIFF
--- a/app/jobs/deposit_work_job.rb
+++ b/app/jobs/deposit_work_job.rb
@@ -13,6 +13,7 @@ class DepositWorkJob < ApplicationJob
     @deposit = deposit
     @request_review = request_review
     @status = Sdr::Repository.status(druid: work_form.druid) if work_form.persisted?
+    @original_cocina_object = Sdr::Repository.find(druid: work_form.druid) if work_form.persisted?
     # Setting current user so that it will be available for notifications.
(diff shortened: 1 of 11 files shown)


justinlittman:
As noted by @amyehodge, with the current approach changing an embargo or access will result in a new user version. (This is because both of these change the structural metadata.) Is this how it should work, or should these changes NOT result in a new user version?

One might argue that since the files on the PURL page are changing, the existing behavior is correct. However, it does feel slightly unintuitive.

@andrewjbtw Do you have thoughts on this?

mjgiarlo (approved):

PR #1423, opened 2025-06-06
Do not create empty creation date when create date has been cleared.

### LINKED ISSUE #1422: Removing creation date is not removing from cocina
https://sul-h3-qa.stanford.edu/works/druid:pt663jx0246 had a creation date:
```
{
        "type": "creation",
        "date": [
          {
            "structuredValue": [
              {
                "value": "2020-05-06",
                "type": "start"
              },
              {
                "value": "2022-09-09",
                "type": "end"
              }
            ],
            "type": "creation",
            "encoding": {
              "code": "edtf"
            }
          }
        ]
      }
```

when @amyehodge removed the creation date, this was left:
```
{
        "type": "creation"
      }
```

justinlittman:
closes #1422

### DIFF
--- a/app/mappers/cocina/work_events_mapper.rb
+++ b/app/mappers/cocina/work_events_mapper.rb
@@ -29,7 +29,7 @@ module Cocina
       date = if create_date_type == 'single'
                create_date_single.to_s
              else
-               "#{create_date_range_from}/#{create_date_range_to}"
+               [create_date_range_from, create_date_range_to].filter_map(&:to_s).join('/').presence
              end
 
       return if date.nil?
--- a/spec/mappers/cocina/work_events_mapper_spec.rb
+++ b/spec/mappers/cocina/work_events_mapper_spec.rb
@@ -3,9 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe Cocina::WorkEventsMapper do
-  describe 'publication events' do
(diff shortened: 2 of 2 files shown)


app/mappers/cocina/work_events_mapper.rb
```diff
@@ -29,7 +29,7 @@ def creation_date_event_params
       date = if create_date_type == 'single'
                create_date_single.to_s
              else
-               "#{create_date_range_from}/#{create_date_range_to}"
+               [create_date_range_from, create_date_range_to].filter_map(&:to_s).join('/').presence
```
mjgiarlo, line 32: Trivial (non-blocking) quibble/suggestion:

```suggestion
               [create_date_range_from, create_date_range_to].filter_map(&:presence).join('/').presence
```

mjgiarlo (approved):
Approve pending consideration of quibble

app/mappers/cocina/work_events_mapper.rb
justinlittman: I'm not convinced that change would work as intended. The forms will never be `nil`; they just might have `nil` attributes that when `to_s` is called might be `nil`. (There is logic in the forms about how to create a string.)

justinlittman: Since the quibble is minor and given ^^, I'm going to leave as is.

PR #1440, opened 2025-06-10
Display pending review alert when item is pending review

### LINKED ISSUE #1322: Next steps message should display after depositor submits an item for review
In H2 we have a specific message about next steps that displays when a user submits an item for review.

Image

Currently in H2 we do not display a message when a user submits an item for review. We only display the message below when the "deposit" button is clicked.

Image

mjgiarlo:
Fixes #1322

**NOTE**: After making the following screenshot, I changed the color & icon of the alert to match what's already displayed when depositing. I.e., the "green" success color and the check-mark icon instead of the info icon. This brings consistency with the depositing alert and parity with H2.

Screenshot from 2025-06-10 10-46-57

### DIFF
--- a/app/components/elements/spinner_component.rb
+++ b/app/components/elements/spinner_component.rb
@@ -44,7 +44,7 @@ module Elements
     private
 
     def message_position_classes
-      case @message_position # rubocop:disable Metrics/HashLikeCase
+      case @message_position # rubocop:disable Style/HashLikeCase
       when :bottom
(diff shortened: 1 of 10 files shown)


mjgiarlo:
Cannot get failing test to pass on CI and it never fails locally with the same rspec seed. Also, I've seen this same system test fail on other CI builds, so I think it's bogus. yay, system specs. Merging.

PR #1455, opened 2025-06-11
Show user version in show page instead of system version

### LINKED ISSUE #1420: Display user version on work show page
This should replace the system version for "Version details".

peetucket:
Fixes #1420 - fetch the current head user version for display and change the label in the show page (used to be system version from cocina object)

tested in QA, see https://argo-qa.stanford.edu/view/druid:jx716kz1859 for an example (https://sul-h3-qa.stanford.edu/works/druid:jx716kz1859)

Note that currently any existing H3 objects do not have a default user version, and the first time a user creates one (by changing files), User Version 1 will then be created.  The result is that the display will continue to show "Version 1" (the default) until they create a *second* change to the files, at which point it will become Version 2

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -39,7 +39,7 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
 
     unless editable?
       flash[:warning] = helpers.t('works.edit.messages.cannot_be_edited_html', support_email: Settings.support_email)
-      return redirect_to work_path(params[:druid]), status: :see_other
+      return redirect_to work_path(druid), status: :see_other
     end
 
     # This updates the Work with the latest metadata from the Cocina object.
@@ -142,6 +142,10 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
 
   private
 
+  def druid
+    params[:druid]
+  end
+
   def work_params
     params.expect(work: WorkForm.user_editable_attributes + [WorkForm.nested_attributes])
   end
@@ -151,11 +155,11 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
   end
 
   def update_work_params
-    work_params.merge(druid: params[:druid])
+    work_params.merge(druid:)
   end
 
   def set_work_and_collection
-    @work = Work.find_by!(druid: params[:druid])
+    @work = Work.find_by!(druid:)
     @collection = @work.collection
   end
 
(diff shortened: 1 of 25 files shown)


app/controllers/works_controller.rb
```diff
@@ -196,7 +200,8 @@ def editable?
   end
 
   def set_presenter
-    @work_presenter = WorkPresenter.new(work: @work, work_form: @work_form, version_status: @version_status)
+    @work_presenter = WorkPresenter.new(work: @work, work_form: @work_form, version_status: @version_status,
+                                        user_version: Sdr::Repository.latest_user_version(druid:))
```
peetucket, line 204: set the `user_version` for display in the presenter, value comes from DSA client call which occurs in the Sdr::Repository object

app/controllers/works_controller.rb
```diff
@@ -142,6 +142,10 @@ def history
 
   private
 
+  def druid
+    params[:druid]
+  end
+
```
peetucket, line 148: refactor into a method and use everywhere needed

app/services/sdr/repository.rb
```diff
@@ -24,6 +24,14 @@ def self.status(druid:)
...
     end
 
+    # @param [String] druid the druid of the object
+    # @return [Integer] the head user version or 1 as default
+    def self.latest_user_version(druid:)
+      Dor::Services::Client.object(druid).user_version.inventory.find(&:head)&.userVersion || 1
+    rescue Dor::Services::Client::NotFoundResponse
+      raise NotFoundResponse, "Object not found: #{druid}"
+    end
+
```
peetucket, line 34: get the user version from DSA client

app/views/works/show.html.erb
```diff
@@ -20,7 +20,7 @@
   <% component.with_row(label: 'DOI', values: [@work_presenter.doi_value]) %>
   <% component.with_row(label: 'Collection', values: [@work_presenter.collection_link]) %>
   <% component.with_row(label: 'Depositor', values: [@work_presenter.depositor]) %>
-  <% component.with_row(label: 'Version details', values: [@work_presenter.version]) %>
+  <% component.with_row(label: 'Version', values: [@work_presenter.user_version]) %>
```
peetucket, line 23: change label in form and show the user_version instead of the system version from cocina

app/services/sdr/repository.rb
peetucket, line 34: get the user version from DSA client

spec/services/sdr/repository_spec.rb
```diff
@@ -149,6 +149,59 @@
     end
   end
 
+  describe '#latest_user_version' do
+    context 'when the object is found' do
+      let(:object_client) { instance_double(Dor::Services::Client::Object, user_version: user_version_client) }
+      let(:user_version_client) { instance_double(Dor::Services::Client::UserVersion, inventory:) }
```
justinlittman, line 155: There's really no reason to use instance doubles for `Dor::Services::Client::UserVersion`.

PR #1465, opened 2025-06-12
dynamic user id

peetucket:
Noticed this in a spec -- it was working before, but it seems like it is better to reference the user.id that may exist (just like we do for the email address) instead of hardcoding it.


justinlittman (changes requested):
Those ids are from H2, so hardcoding them is correct.

PR #1466, opened 2025-06-12
Consistency in creating works

lwrubel:


spec/services/admin/work_report_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-
+# rubocop:disable RSpec/LetSetup
```
justinlittman, line 4: Do you need to use `let!` at all? Can the `create` go in `before` blocks?

PR #1469, opened 2025-06-12
add total files and file size to item report

### LINKED ISSUE #1430: Add number of files and total file size to item report
Completes requirements in #550 to add columns to the admin item report (work report):
* number of files
* total file size

peetucket:
Fixes #1430 - add files count and size to admin item report

Tested on localhost and QA

Example report attached from QA: 
[item_report.csv](https://github.com/user-attachments/files/20730960/item_report.csv)

### DIFF
--- a/app/services/admin/work_report.rb
+++ b/app/services/admin/work_report.rb
@@ -161,7 +161,9 @@ module Admin
           next unless druids.include?(druid)
 
           work_model = Work.find_by(druid:)
-          content_files = work_model.content.last.content_files
+          cocina_object = Sdr::Repository.find(druid:)
+          content = Contents::Builder.call(cocina_object:, user: work_model.user, work: work_model)
+          content_files = content.content_files
           row = [work_form.title,
                  work_model.id,
                  work_form.druid,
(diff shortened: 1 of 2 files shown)


app/services/admin/work_report.rb
```diff
@@ -139,18 +139,22 @@ def statuses
...
       @work_forms ||= druids.map do |druid|
         cocina_object = Sdr::Repository.find(druid:)
+        work = Work.find_by(druid:)
         # only cocina_object will be used; other arg assignments are irrelevant
-        Form::WorkMapper.call(cocina_object:, doi_assigned: true, agree_to_terms: true, version_description: nil,
-                              collection: nil)
+        work_form = Form::WorkMapper.call(cocina_object:, doi_assigned: true, agree_to_terms: true,
+                                          version_description: nil, collection: nil)
+        work_form.content_id = Contents::Builder.call(cocina_object:, user: work.user, work:).id
+        work_form
```
peetucket, line 147: Why all these shenanigans here to set the `content_id` in the work form?  Because if we need to use `Contents::Builder.call` to get the correct `content` model object, we need to have the cocina object to do this.  We already have it loaded here, so we might as well make use of it instead of loading it again when iterating below during the report.

peetucket: Though this approach has some potential performance considerations.  We build up an array of WorkForms in memory and then iterate over them below.  If we run a report over many many druids, which could potentially happen in the future, this could blow up.  Perhaps we should do this directly in the loop when generating the report instead of building it all in memory first.

justinlittman (approved):

PR #1476, opened 2025-06-13
Make work report spec more deterministic.

justinlittman:


app/services/admin/work_report.rb
```diff
@@ -113,7 +113,8 @@ def selected_states?(druid)
     # rubocop:enable Metrics/CyclomaticComplexity, Metrics/MethodLength
 
     def druids
-      @druids ||= query.pluck(:druid).compact
+      # Sorting makes this deterministic for testing.
+      @druids ||= query.pluck(:druid).compact.sort
```
mjgiarlo, line 117: The approach of using sorting for deterministic tests sounds good to me. I wonder if we want to push this into the database by either calling `.order` on `@query = Work.all` in `#initialize` above, or perhaps by adding a default scope to the work model.

justinlittman: I like that suggestion.

PR #1478, opened 2025-06-13
[SPIKE] Use capybara-lockstep gem to increase system spec reliability in CI

mjgiarlo:
- **Yank playwright out and use selenium for system specs**
- **Use capybara-lockstep gem to improve test reliability**


spec/system/show_discard_work_draft_spec.rb
```diff
@@ -70,7 +70,6 @@
         click_on('Discard draft')
       end
 
-      expect(page).to have_current_path(work_path(druid))
       expect(page).to have_no_button('Discard draft')
       expect(page).to have_css('.alert-success', text: 'Draft discarded.')
```
mjgiarlo, line 74: We expect the `Draft discarded` message on the work path page, so I removed the offending line above since it seemed to cause flapping for me locally. I don't think we're going to miss it.

spec/system/validate_collection_deposit_spec.rb
```diff
@@ -79,11 +79,8 @@
     find('.nav-link', text: 'Save your collection').click
     expect(page).to have_css('.nav-link.active', text: 'Save your collection')
     click_link_or_button('Save', class: 'btn-primary')
-    expect(page).to have_css('h1', text: collection_title_fixture)
-    expect(page).to have_current_path(collection_path)
-
-    # No Alert!
-    expect(page).to have_no_css('.alert-danger', text: 'Required fields have not been filled out.')
```
mjgiarlo, line 86: This test failed consistently for me. And it turns out these expectations... are not at all what we expect. They only passed prior because the tests ran too fast. What we *actually* expect is the page to be the collection wait page which is what we have below.

Turning on capybara-lockstep can cause tests that used to pass to fail for this reason, so this is not unexpected.

peetucket (approved):

justinlittman (approved):

PR #1492, opened 2025-06-17
Implement permissions for shared works.

### LINKED ISSUE #1484: Implement share permissions
* View only: User sees the item listed on their dashboard and is able to load the show page but does not see buttons or icons for loading the form. Or buttons/icons could be greyed out.
* View and edit: Same as above for view only, and also is able to edit anything on the form, including files, but only sees the "Save as draft" button and not the "Deposit" or "Submit for review" button.
* View, edit, and deposit: Same as above for View and edit, and also is able to actually deposit or submit for review.

This ticket is NOT for the UI for editing sharing.

justinlittman:
closes #1484

### DIFF
--- a/spec/components/works/show/header_component_spec.rb
+++ b/spec/components/works/show/header_component_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Works::Show::HeaderComponent, type: :component do
     WorkPresenter.new(work_form:, version_status:, work:)
   end
   let(:work_form) { WorkForm.new(druid: druid_fixture, title:) }
-  let(:work) { create(:work, review_state:) }
+  let(:work) { create(:work, review_state:, user: owner) }
   let(:version_status) do
     instance_double(VersionStatus, editable?: editable_version_status, discardable?: discardable_version_status,
                                    status_message:)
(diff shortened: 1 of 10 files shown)


spec/components/works/show/header_component_spec.rb
```diff
@@ -43,6 +47,16 @@
     end
   end
 
+  context 'when editable version status and user does not have permissions to edit' do
+    let(:editable_version_status) { true }
+    let(:permission) { 'view' }
+
+    it 'renders the edit button' do
```
jmartin-sul, line 54:
```suggestion
    it 'does not render the edit button' do
```

spec/components/works/show/header_component_spec.rb
```diff
@@ -79,6 +94,15 @@
     end
   end
 
+  context 'when discardable and user does not have permissions to discard' do
+    let(:discardable_version_status) { true }
+
+    it 'renders the discard button' do
```
jmartin-sul, line 100:
```suggestion
    it 'does not render the discard button' do
```

jmartin-sul (approved):
approve pending consideration of possible copypasta

PR #1479, opened 2025-06-13
Collection deposits search

jmartin-sul:
closes #1021 
Screenshot 2025-06-13 at 3 23 04 PM
Screenshot 2025-06-13 at 3 23 16 PM
Screenshot 2025-06-13 at 3 23 24 PM
Screenshot 2025-06-13 at 3 23 35 PM
Screenshot 2025-06-13 at 3 23 45 PM
Screenshot 2025-06-13 at 3 23 59 PM

Issue #1021, opened 2025-03-10
Add search for collection deposits list


amyehodge:
Collection deposits page should have a box to allow users to search the list of items in the collection.

<img width="1022" alt="Image" src="https://github.com/user-attachments/assets/a4ad7044-2e3d-4f24-aa39-975d21649b87" />

Figma: https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1114-2737&t=9kODmCFr9oSqITfS-4

justinlittman:
Can you say more about the search? Title only? Case insensitive?

amyehodge:
Search should be case insensitive and should search fields: title, owner, link.

I'm realizing that H2 also has the ability to sort the fields in this table. If we can search the status and the dates as well, sorting will be less important.

PR #1501, opened 2025-06-19
Fix collection deposit query

jmartin-sul:
It turns out that I got the `OR` clauses a bit wrong in #1479, which led to inadvertently including anything that matched the given search term, regardless of the collection it lives in.  this PR fixes things so that:
* only results for the given collection are returned
* we no longer search email (the original ticket, #1021, said to search for "owner", which i mistakenly assumed meant name and email)

This also adds specs that would've caught this issue in the first place.  I think I failed to notice the problem because the bit of manual testing I did for searches that could have results in multiple collections used the title field, which never had this issue, because it always filtered on the given collection.  And I didn't have any unit tests that covered this case.  You can see the new specs failing in the first commit.

@justinlittman had the suggestion for the query change.  The problem made sense when I looked at the underlying SQL for the old and new ones, note the lack of parens grouping the `OR`s in the original:

What was being generated before this PR:
```sql
SELECT "works".* FROM "works"
INNER JOIN "users" ON "users"."id" = "works"."user_id"
WHERE ("works"."collection_id" = 212 AND (title ILIKE '%kk%') OR (druid ILIKE '%kk%') OR (email_address ILIKE '%kk%') OR (name ILIKE '%kk%') OR (first_name ILIKE '%kk%'))
```

After this PR:
```sql
SELECT "works".* FROM "works"
INNER JOIN "users" ON "users"."id" = "works"."user_id"
WHERE "works"."collection_id" = 212 AND ((title ILIKE '%kk%') OR (druid ILIKE '%kk%') OR (name ILIKE '%kk%'))
```

### DIFF
--- a/spec/requests/show_collection_works_spec.rb
+++ b/spec/requests/show_collection_works_spec.rb
@@ -130,7 +130,7 @@ RSpec.describe 'Show collection works' do
         expect(response.body).not_to include(works.to_a[1].title)
         expect(response.body).not_to include(works.to_a[2].title)
 
-        query = works.to_a[2].user.first_name
+        query = works.to_a[2].user.name
         get "/collections/#{druid}/works", params: { q: query }
         expect(response.body).to include(works.to_a[2].title)
         expect(response.body).not_to include(works.to_a[1].title)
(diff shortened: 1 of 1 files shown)


justinlittman (approved):
Thanks for the quick fix!

PR #1504, opened 2025-06-19
user selection of days should be kept

### LINKED ISSUE #1503: Admin recent activity should retain user days ago selection
Visit https://sul-h3-stage.stanford.edu/admin/recent_activity_collections?admin_recent_activity%5Bdays_limit%5D=30

Select a different timeframe.  List shows results correctly but drop-down selector does not retain user selection.

peetucket:
Fixes #1503 - user selection of "days ago" should stick in admin collection recent activity report

### DIFF
--- a/app/controllers/admin/recent_activity_collections_controller.rb
+++ b/app/controllers/admin/recent_activity_collections_controller.rb
@@ -13,11 +13,11 @@ module Admin
     private
 
     def days_limit
-      params.fetch(:admin_recent_activity, {}).fetch(:days_limit, 7).to_i
+      recent_activity_params.fetch(:days_limit, Admin::RecentActivityForm.new.days_limit).to_i
     end
 
     def recent_activity_params
-      params.permit(:days_limit)
+      params.fetch(:admin_recent_activity, {}).permit(:days_limit)
     end
 
     def rows


app/controllers/admin/recent_activity_collections_controller.rb
```diff
@@ -13,11 +13,11 @@ def index
     private
 
     def days_limit
-      params.fetch(:admin_recent_activity, {}).fetch(:days_limit, 7).to_i
+      recent_activity_params.fetch(:days_limit, Admin::RecentActivityForm.new.days_limit).to_i
```
peetucket, line 16: the default value of "7" is provided in the form, let's use it here (else if you change in one spot and not the other it won't work as expected)

app/controllers/admin/recent_activity_collections_controller.rb
```diff
@@ -13,11 +13,11 @@ def index
...
     end
 
     def recent_activity_params
-      params.permit(:days_limit)
+      params.fetch(:admin_recent_activity, {}).permit(:days_limit)
```
peetucket, line 20: this allows the params to be set correctly so they can be used when rendering the form

aaron-collier (approved):

app/controllers/admin/recent_activity_collections_controller.rb
justinlittman, line 16: Can you make it a constant on the form and reference it here? This will make it clearer than instantiating a form to get the value.

PR #1505, opened 2025-06-19
Adds UI for editing shares.

### LINKED ISSUE #1468: [DRAFT] Allow depositor to assign permissions for an item to other users
[Proposal doc](https://docs.google.com/document/d/1jWiSQaDfDDcooB7QpcGW04CbEC0dI9kuCaWiErbj-gg/edit?tab=t.0#heading=h.dp2nitfpf5k0)
[Figma design](https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=5668-8740)

Requirements:

- User must be able to assign other users permission for access to a single work before the work is deposited in order to help with deposit, as well as after the work is deposited to manage the work in the future.
-- They do not need to be able to assign permissions to more than one work at a time.
-- Additional editors of the work should not be acknowledged in the metadata.
- User needs to be able to assign the level of permission: view only, view + edit, or view + edit + deposit.
- Permissions should be assigned until revoked by the user. Admins should be able to revoke permissions as well.
-- Not sure we have a use case for collection managers needing to be able to do this.
- Revoking permissions should be an easy task for the user.
- Users should be able to do this in a self-service UI that allows for the addition of users in bulk.

On the item view page, there will be a button for the depositor to manage who has access to this specific work. 

Only the person who "owns" this work can change these permissions. Other people that are added, even with depositor access, should not be able to add or remove people or change permissions. Basically, they shouldn't see this button.

[REPLACE]
Image

This will take the user to a new page where they can give access to additional people just to this work.

[REPLACE]
Image

The text box here operates exactly the same as the participant entry for collections with the exception that when the "Add people" button is clicked, identified users will show up in the list below the header "List of people with access" and by default will be assigned "View only" access. 

If a user is not found, the behavior will be the same as the existing implementation for collection participants with the error message and the unfound entries remaining in the entry box.

The three access options in the drop-down menus will be:
- View only: User sees the item listed on their dashboard and is able to load the show page but does not see buttons or icons for loading the form. Or buttons/icons could be greyed out.
- View and edit: Same as above for view only, and also is able to edit anything on the form, including files, but only sees the "Save as draft" button and not the "Deposit" or "Submit for review" button.
- View, edit, and deposit: Same as above for View and edit, and also is able to actually deposit or submit for review. 

The depositor of this item will be able to select the correct level of access for each person using the dropdown menu.
The depositor of this item will be able to remove individuals using the trash can icon.
None of these users can change the permissions for who else has access to the deposit.

If a user already has other permissions that let them do things, e.g. they are a collection manager, then those should prevail. I hadn't considered this and am wondering if we should do anything to make that clear, like detect that someone is already a manager and then not let them add them here and display a message like "this person can already do all this stuff" (excepted worded better)?

justinlittman:
closes #1468


app/components/collections/edit/workflow_component.html.erb
```diff
@@ -14,6 +14,6 @@
   <% end %>
 
   <%= tag.div data: { workflow_target: 'reviewersSection' }, class: 'd-none' do %>
-    <%= render Collections::Edit::BulkParticipantsComponent.new(form:, field_name: :reviewers, model_class: ReviewerForm) %>
+    <%= render Edit::BulkParticipantsComponent.new(form:, field_name: :reviewers, model_class: ReviewerForm, form_component: Collections::Edit::ParticipantComponent) %>
```
justinlittman, line 17: The existing `Collections::Edit::BulkParticipantsComponent` was moved and tweaked so that it could be used for editing shares.

app/components/edit/bulk_participants_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Edit
```
justinlittman, line 3: This is just a move with a little tweak (adding `form_component` as an argument).

app/components/elements/forms/fieldset_component.html.erb
```diff
@@ -1,12 +1,14 @@
 <%= tag.fieldset class: classes, data:, id:, aria: do %>
-  <%= tag.legend class: legend_classes do %>
-    <% if legend? %>
-      <%= legend %>
-    <% else %>
-      <%= tag.label class: label_classes do %><%= label %><% end %>
+  <% if legend? || label %>
```
justinlittman, line 2: This is an opportunistic cleanup to avoid rendering unnecessary html.

app/components/works/show/manage_sharing_button_component.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Works
```
justinlittman, line 3: Encapsulating this button in a component made it a little cleaner and easier to test.

app/policies/account_policy.rb
```diff
@@ -3,12 +3,16 @@
 # Authorization policy for lookup of accounts
 class AccountPolicy < ApplicationPolicy
   def search?
-    collection_creator? || manages_any_collection?
+    collection_creator? || manages_any_collection? || owns_any_work?
```
justinlittman, line 6: So that the account service can be used in this context also.

app/controllers/shares_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+  def share_params
+    if params.include?(:work_share)
+      params.expect(work_share: [WorkShareForm.nested_attributes])
+    else
+      { shares_attributes: [] }
+    end
+  end
+
+  # Updates the shares of a work based on the provided form.
+  class ShareUpdater
```
peetucket, line 41: can we move this elsewhere so we don't have a class definition embedded within the controller?  like maybe into the service folder ?

justinlittman: We could, but it is only used here. This keeps the services directory slim and puts it near the code in which it is used.

peetucket: Understood.  We don't have a shared practice on this since we have other cases of a class being used in only one place but defined in it's own file with a filename matching the class, which is what I'd typically expect.  It's not a big deal, but my first instinct when seeing a reference to a class is to go looking for it's definition in the typically expected spots and not to scroll down in the same file (particularly in a controller).

lwrubel (approved):
Approved but there is a lot going on here, so would understand if you want another person to read through before merging.

PR #1512, opened 2025-06-20
Send emails when a work is shared

### LINKED ISSUE #1487: Implement share emails
Send emails to individuals with whom a depositor has shared a work.

Subject line: Someone has shared a work with you in the Stanford Digital Repository

Text:
[Name of user sharing work] has shared with you the work "[work title linked to the work show page]" in the Stanford Digital Repository. You have permission to [permission level] the work.

You an access the work by logging into the SDR at https://sdr.stanford.edu or clicking the linked work title above.

<p>Please be sure to review the <%= link_to 'Terms of Deposit', Settings.terms_url %>.

(and then I think there is a standard signature/sign-off that you put here)

The [permission level] should say one of the following:
- view but not edit or deposit
- view and edit but not deposit
- view, edit, and deposit

peetucket:
Fixes #1487 

~~Waiting on #1505 which implements the sharing UI and has methods we need to hook into the send the emails~~

Send email with `WorksMailer.with(work:, permission: share.permission, user_shared_with: user).share_added_email`

### DIFF
--- a/app/views/works_mailer/share_added_email.html.erb
+++ b/app/views/works_mailer/share_added_email.html.erb
@@ -2,7 +2,7 @@
 
 <p><%= @work.user.name %> has shared with you the work "<%= link_to @work.title, @work %>" in the Stanford Digital Repository. You have permission to <%= @share.permission %> the work.
 
-<p>You an access the work by logging into the SDR at <a href="https://sdr.stanford.edu">https://sdr.stanford.edu</a> or clicking the linked work title above.</p>
+<p>You can access the work by logging into the SDR at <a href="https://sdr.stanford.edu">https://sdr.stanford.edu</a> or clicking the linked work title above.</p>
 
 <p>Please be sure to review the <%= link_to 'Terms of Deposit', Settings.terms_url %>.</p>
 


spec/mailers/previews/works_mailer_preview.rb
```diff
@@ -0,0 +1,27 @@
...
+  def ownership_changed_email
+    WorksMailer.with(work: Work.first).ownership_changed_email
+  end
+
+  def share_added_email
+    WorksMailer.with(work: Work.first, permission: 'view',
+                     user_shared_with: User.new(first_name: 'Shared', name: 'With',
+                                                email_address: 'test@test.com')).share_added_email
+  end
+end
```
peetucket, line 27: adds previews for all work mailers, including this new one

spec/mailers/previews/works_mailer_preview.rb
```diff
@@ -0,0 +1,28 @@
...
+    WorksMailer.with(work: Work.first).ownership_changed_email
+  end
+
+  def share_added_email
+    work = Work.first
+    shared_with = User.new(first_name: 'SharedWith', name: 'User', email_address: 'shared_with@test.com')
+    share = Share.new(user: shared_with, work:, permission: Share::VIEW_PERMISSION)
+    WorksMailer.with(share:, work:).share_added_email
+  end
+end
```
peetucket, line 28: adds previews for all work mailers, including this new one

app/models/share.rb
```diff
@@ -9,6 +9,8 @@ class Share < ApplicationRecord
   belongs_to :user
   belongs_to :work
 
+  after_create -> { Notifier.publish(Notifier::SHARE_ADDED, share: self, work:) }
```
justinlittman, line 12: Do you need to provide the work?

app/views/works_mailer/share_added_email.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= render Emails::SalutationComponent.new(user: @share.user) %>
+
+<p><%= @work.user.name %> has shared with you the work "<%= link_to @work.title, @work %>" in the Stanford Digital Repository. You have permission to <%= @share.permission %> the work.
+
+<p>You an access the work by logging into the SDR at <a href="https://sdr.stanford.edu">https://sdr.stanford.edu</a> or clicking the linked work title above.</p>
```
justinlittman, line 5: Typo: "You can access"

justinlittman (approved):
Approve pending addressing comments.

app/models/share.rb
peetucket: no you don't, you can get it from the share... but if you don't pass in the work, these lines will cause trouble, because it tries to get the work from the params directly and then uses that to get the user: 

https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/mailers/works_mailer.rb#L5
https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/mailers/application_mailer.rb#L8-L9

We could skip the `before_action` for just this mailer method and set the work and user separately via the share, but that seemed like it would be harder to understand by a developer later than just passing in the work too.

PR #1515, opened 2025-06-20
search files in edit view when paginating

### LINKED ISSUE #402: Add search functionality for paginated file display in form
@astridu Can you provide a design for this? Thanks.

peetucket:
HOLD for PO approval on QA

Fixes #402 - allow users to search for files by filename, only in the edit view and only when the object has more than a specified number of files (set to 50, but could be any value we want).  

The search works for both hierarchical file display and flat file display (i.e. will be shown for either view as long as the total number of files in the object is greater than the specified number, which is independent of both the page size and the limit that triggers the shift from hierarchical display with no paging to flat list with paging).

Note: the "x" to cancel the current search is not in this PR, it is also not in the collection deposits search view that is merged, and doesn't not appear to be in the component library used.

Here are the possible variants of file displays in this PR:

1. An object with no files - **shows a message** indicating there are no files (identical to current)

Screenshot 2025-06-23 at 1 49 58 PM

2. An object with fewer than the hierarchical file display limit (set to 1000) and fewer than the minimum needed for search bar (set to 50) - **shows a hierarchical display of files with no search box, no pagination** (identical to current).  Two screenshots below, one actually has hierarchy, the other doesn't, but it's the same view generating both under the hood.

Screenshot 2025-06-23 at 1 51 15 PM

Screenshot 2025-06-23 at 1 52 10 PM

3. An object with fewer than the hierarchical file display limit (set to 1000) but more than the minimum needed for search bar (set to 50) - **shows a hierarchical display of files and includes a search box, no pagination**.    Two screenshots below, one is shown with no search term showing the full list, the second shows a search term applied (smaller list):

No search:
Screenshot 2025-06-23 at 1 54 42 PM

With search:
Screenshot 2025-06-23 at 1 55 46 PM

4. An object with more than the hierarchical file display limit (set to 1000) and more than the minimum needed for search bar (set to 50) - **shows a flat display of files and includes a search box, with pagination**.    If the number of search results exceed the page size, paging will work as normal. Two screenshots below, one is shown with no search term showing the paginated list, the second shows a search term applied (smaller list):

No search, full list with paging:
Screenshot 2025-06-23 at 1 57 35 PM
Screenshot 2025-06-23 at 1 58 16 PM

With search, smaller list (no paging needed):
Screenshot 2025-06-23 at 1 57 40 PM

5. Any object with more than the minimum needed for the search bar (set to 50), when the user searches for something that can't be found - show a message:

Screenshot 2025-06-23 at 2 00 48 PM

6. _Edge case that is theoretically possible if you adjust the settings in an odd way_ - An object with more than the hierarchical file display limit but less than the minimum needed for search bar  - **shows a flat display of files with no search box, with pagination**.    We don't expect this to happen, given that the number that triggers a flat view with paging is likely to always be bigger than the number needed to show the search bar, but it would work as you'd expect it to.

To do:
- [x] tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,7 +21,6 @@ AllCops:
     - 'config/environments/*.rb'
     - 'db/**/*'
     - 'vendor/**/*'
-    - 'tmp/**/*'
 
 Layout/LineLength:
   Max: 120
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,6 +21,7 @@ AllCops:
     - 'config/environments/*.rb'
     - 'db/**/*'
     - 'vendor/**/*'
+    - 'tmp/**/*'
 
 Layout/LineLength:
   Max: 120
--- a/app/components/works/edit/content_hierarchy_component.rb
+++ b/app/components/works/edit/content_hierarchy_component.rb
@@ -4,15 +4,12 @@ module Works
   module Edit
     # Component for rendering an editable file hierarchy table for a Content.
     class ContentHierarchyComponent < ApplicationComponent
-      # NOTE: content is a reserved word in a view component, so using content_obj instead.
-      def initialize(content_obj:)
-        @content_obj = content_obj
+      def initialize(content_files:)
+        @content_files = content_files
         super()
       end
 
-      def content_files
-        @content_obj.content_files.path_order
-      end
+      attr_reader :content_files
     end
   end
 end
--- a/app/components/works/edit/content_non_hierarchy_component.html.erb
+++ b/app/components/works/edit/content_non_hierarchy_component.html.erb
@@ -9,16 +9,16 @@
     <% content_files.each do |content_file| %>
         <% component.with_row do |row_component| %>
             <% row_component.with_cell do %>
-            <%= content_file.filepath %>
+                <%= content_file.filepath %>
             <% end %>
             <% row_component.with_cell do %>
-            <%= render Works::Edit::ContentFileDescriptionComponent.new(content_file:) %>
+                <%= render Works::Edit::ContentFileDescriptionComponent.new(content_file:) %>
             <% end %>
             <% row_component.with_cell do %>
-            <%= render Works::Edit::ContentFileHideComponent.new(content_file:) %>
+                <%= render Works::Edit::ContentFileHideComponent.new(content_file:) %>
             <% end %>
             <% row_component.with_cell do %>
-            <%= render Works::Edit::ContentFileDeleteComponent.new(content_file:) %>
+                <%= render Works::Edit::ContentFileDeleteComponent.new(content_file:) %>
             <% end %>
         <% end %>
     <% end %>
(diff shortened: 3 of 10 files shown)


app/components/works/edit/content_hierarchy_component.rb
```diff
@@ -4,15 +4,12 @@ module Works
...
-        @content_obj = content_obj
+      def initialize(content_files:)
+        @content_files = content_files
         super()
       end
 
-      def content_files
-        @content_obj.content_files.path_order
-      end
+      attr_reader :content_files
```
peetucket, line 12: we need to pass in the content_files specifically now, because it may have a where search filter applied to the results ... we didn't need the full content object anyway for this view

app/views/works/form.html.erb
```diff
@@ -53,7 +53,6 @@
           <%= render Elements::HorizontalRuleComponent.new(classes: 'my-3') %>
         <% end %>
 
-        <%= render Edit::SectionHeaderComponent.new(title: t('works.edit.panes.files.list')) %>
```
peetucket, line 56: rendering the section header is going to move into the files list component itself, since there is now some conditions on how it renders (with or without a search box next to it)

app/views/contents/_show.html.erb
```diff
@@ -1,36 +1,66 @@
 <%= turbo_frame_tag content, 'show' do %>
   <%= turbo_stream_from 'show', content %>
   <% content_files ||= content.content_files.path_order.page(1) %>
```
peetucket, line 3: i'm not sure why this is this here -- content_files should always exist coming from the controller?

.rubocop.yml
```diff
@@ -21,6 +21,7 @@ AllCops:
     - 'config/environments/*.rb'
     - 'db/**/*'
     - 'vendor/**/*'
+    - 'tmp/**/*'
```
peetucket, line 24: if you upload files that look like code to a test object in localhost, this get thrown into a subfolder under tmp... we do not want rubocop running on these files

estanfor:
It looks like the search returns items if the search string appears in either the child filename or the parent folder name. Is that how it's supposed to work? I can see why it works that way, but it might be annoying if there are folders with lots of files in them that all get returned if the folder name matches the search.

peetucket:
Will unhold PR.  It is easy to change to filenames only later if feedback warrants.  Or offer a checkbox in the UI and let the user decide.

app/components/works/edit/content_non_hierarchy_component.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <% component.with_headers([
+                                { label: 'File Name' },
+                                { label: 'Description', tooltip: t('content_files.edit.fields.description.tooltip_html'), style: 'width: 180px;' },
+                                { label: 'Hide', tooltip: t('content_files.edit.fields.hide.tooltip_html'), style: 'width: 85px;' },
+                                { label: 'Action', style: 'width: 85px;' }
+                              ]) %>
+    <% content_files.each do |content_file| %>
+        <% component.with_row do |row_component| %>
+            <% row_component.with_cell do %>
+            <%= content_file.filepath %>
```
justinlittman, line 12: Can you fix the indentation do that lines between do and end are indented?

justinlittman (approved):
Approve pending requested change.

app/components/works/edit/content_non_hierarchy_component.html.erb
peetucket: Fixed, will merge when PR is green.

PR #1528, opened 2025-06-25
show link on dashboard to open contact form with request a collection selected

peetucket:
Fixes #1352

Alternative implementation in #1530 

Borrow how it's done from H2 (mostly)

New link at bottom of dashboard

Screenshot 2025-06-25 at 12 10 22 PM

when clicked, opens  modal with appropriate option pre-selected

Screenshot 2025-06-25 at 12 09 23 PM

PR #1530, opened 2025-06-25
use two modals instead of stimulus to trigger request collection form

### LINKED ISSUE #1352: Add link for people looking for another collection to bottom of dashboard
This replicates an existing H2 feature.

The link text should open the "Contact SDR staff" form with the option to "Request access to another collection" selected from the "How can we help you?" menu.

Link above footer that says "Looking for a different collection?" Note this text is slightly different from what's in H2, shown below.
Image

peetucket:
Alternative implementation of #1528 which uses two modals instead of stimulus to show the two possible contact us forms

Fixes #1352

### DIFF
--- a/app/components/structure/header_component.html.erb
+++ b/app/components/structure/header_component.html.erb
@@ -73,4 +73,4 @@
 <%= render Structure::TermsOfDepositModalComponent.new %>
 <%# there are two contact us modals; the first is the regular one; the second pre-selects requesting a collection %>
 <%= render Structure::ContactFormModalComponent.new(form_id: 'contact-form-modal', path: new_contacts_path(modal: true)) %>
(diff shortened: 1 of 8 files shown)


app/components/structure/header_component.html.erb
```diff
@@ -71,4 +71,6 @@
   </div>
 </header>
 <%= render Structure::TermsOfDepositModalComponent.new %>
-<%= render Structure::ContactFormModalComponent.new %>
+<%# there are two contact us modals; the first is the regular one; the second pre-selects requesting a collection %>
+<%= render Structure::ContactFormModalComponent.new(form_id: 'contact-form-modal', path: new_contacts_path(modal: true)) %>
+<%= render Structure::ContactFormModalComponent.new(form_id: 'request-collection-form-modal', path: new_contacts_path(help_how: 'Request access to another collection', modal: true)) %>
```
peetucket, line 76: we now have two contact modals on the page instead of one

PR #1562, opened 2025-06-27
set page html title from breadcrumb

### LINKED ISSUE #1540: Be more specific about page titles in the browser
Currently, all pages seem to have the same page title when I look at the history in the back button in Chrome:

Image

H2 lists specific collection and work titles and dashboard, and may have other specific page titles. We should be more specific like this in H3 as well. I would expect this is needed for accessibility.

Image

peetucket:
Fixes #1540 

Uses the same logic as H2 - when there are breadcrumbs rendered, the page title is derived from these by joining them together, and prepending `SDR | `.  For example, if you were here, your page title would be `SDR | Dashboard | New Collection | Test Deposit - Hierarchy`:

Screenshot 2025-06-27 at 1 45 48 PM

When there are no breadcrumbs, you get the default page title (as current) of `SDR | Stanford Digital Repository`


app/components/elements/breadcrumb_nav_component.html.erb
```diff
@@ -1,3 +1,4 @@
+<% content_for(:title, page_title_from_breadcrumbs) # set the page's html title from the breadcrumb nav %>
```
peetucket, line 1: Note: the application layout is already setup to render the `:title` when it is set here.  See https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/views/layouts/application.html.erb#L13

justinlittman (approved):

PR #1578, opened 2025-07-01
Do not allow user to globus upload until draft saved.

### LINKED ISSUE #1575: Require saving as draft before Globus button is active
When creating an item or updating an existing deposited item, the "Use Globus to transfer files" button should be greyed out until the user has saved a draft of the item.

I'm thinking that perhaps below the greyed out button we could just say "Click "Save as draft" and return here to use Globus."

justinlittman:
closes #1575

### DIFF
--- a/app/jobs/globus_setup_job.rb
+++ b/app/jobs/globus_setup_job.rb
@@ -18,16 +18,12 @@ class GlobusSetupJob < ApplicationJob
 
   def mkdir
     # This will ignore if the directory already exists.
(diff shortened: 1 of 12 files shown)


app/jobs/globus_setup_job.rb
```diff
@@ -18,16 +18,12 @@ def perform(user:, content:)
 
   def mkdir
     # This will ignore if the directory already exists.
-    path = work.present? ? GlobusSupport.work_path(work:) : GlobusSupport.user_path(user:)
+    path = GlobusSupport.work_path(work:)
```
lwrubel, line 21: I think the comment no longer applies.

PR #1580, opened 2025-07-01
Make show collection works spec sorting determinate.

justinlittman:


spec/requests/show_collection_works_spec.rb
```diff
@@ -48,7 +48,7 @@
              :with_required_contact_email, works_count:, reviewers_count: 2, druid:, title: collection_title_fixture,
                                            contributors: [contributor], managers: [manager], depositors: [depositor])
     end
-    let(:works) { collection.works.order(:title) }
+    let(:works) { collection.works.order(:id) }
```
lwrubel, line 51: Should `other_works` be ordered similarly in L58 to ensure they line up when editing in L86?

PR #1594, opened 2025-07-01
Show no changes error instead of what's changing validation error.

### LINKED ISSUE #1225: Show no changes error before error for empty "what's changing" box
I don't know if there is a way to do this, but it's strange to be shown the validation error for needing to fill out the "what's changing" box and then after you do that be told that there aren't any changes and you need to cancel out.

https://github.com/user-attachments/assets/18c7d525-b8e0-429b-8029-4e17c51c37f8

justinlittman:
closes #1225


app/controllers/works_controller.rb
```diff
@@ -285,6 +285,18 @@ def add_max_release_date
     @work_form.max_release_date = [@collection.max_release_date, @work_form.release_date].compact.max
   end
 
+  def handle_no_changes_or_invalid
+    if @valid
+      handle_no_changes
+    elsif @work_form.errors.one? && @work_form.errors.first.attribute == :whats_changing
```
lwrubel, line 291: TIL about `Array.one?`

lwrubel (approved):

PR #1638, opened 2025-07-08
Indicate there are changes if they are only file changes

### LINKED ISSUE #1629: Globus error happens on deposit as well
I've just noticed that the exact same error happens with Globus when at the very end of the process described in https://github.com/sul-dlss/hungry-hungry-hippo/issues/1603 you click "deposit" instead of "save as draft" without having filled in the what's changing box. You get a message saying that nothing has changed.

Apologies for not realizing this when I filed the previous ticket.

aaron-collier:
Fixes #1629 

This fixes the bug that if the only changes to a work on deposit or save draft is uploading files but the "what's changing" isn't filled out the form indicates as such instead of "no changes"

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -298,7 +298,11 @@ class WorksController < ApplicationController # rubocop:disable Metrics/ClassLen
   end
 
   def invalid_for_whats_changing_only?
-    @work_form.errors.one? && @work_form.errors.first.attribute == :whats_changing
+    @work_form.errors.one? && @work_form.errors.first.attribute == :whats_changing && !attached_files_changed?
+  end
+
+  def attached_files_changed?
+    @content.content_files.any? { |content_file| content_file.file_type == 'attached' }
   end
 
   def handle_no_changes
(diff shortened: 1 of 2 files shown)


app/controllers/works_controller.rb
```diff
@@ -298,7 +298,11 @@ def handle_no_changes_or_invalid
   end
 
   def invalid_for_whats_changing_only?
-    @work_form.errors.one? && @work_form.errors.first.attribute == :whats_changing
+    @work_form.errors.one? && @work_form.errors.first.attribute == :whats_changing && !attached_files_changed?
+  end
+
+  def attached_files_changed?
+    @content.content_files.any? { |content_file| content_file.file_type == 'attached' }
```
justinlittman, line 305: What if there are files deleted? Renamed? Globus?

aaron-collier: Ah yeah, on it.

justinlittman:
@aaron-collier Do you have steps to reproduce the bug?

aaron-collier:
@justinlittman yes,
- Create a new work with at least one file and deposit/accession it.
- Edit the work and upload a new file, or delete a file but make no other changes
- Click deposit

You will get a "You haven't made any changes" error, not the "what's changing" validation error.

PR #1640, opened 2025-07-08
[DRAFT] Display error message on duplicate filename

### LINKED ISSUE #1634: Properly handle new files with the same name
So, here's what we are seeing:

- I created a new item and uploaded a file called test.txt. I then opened a new version, uploaded a different version of the file with the same name. It looked like the file was uploaded. I then went to the deposit page and tried to deposit without filling in the "what's changing" box. I was told that I had not made any changes to the item and couldn't deposit.
- I then cancelled out of the form and started over, but this time filled in the "what's changing" box before clicking deposit. The same thing happened -- I was told I had not made any changes. I then also changed the title so I could deposit and check what happened with the file. As you reported, the file did not change.
- As you described, if I delete the old file first and then upload the new one, I am allowed to deposit

There are a couple of ways to handle this. In H2, I think what we do is detect that the file name is the same and tell the user that they need to delete the old file first.

But we should not have the app behave as if the new file with the same has been uploaded but then NOT change the file. This is VERY BAD.

aaron-collier:
For consideration/discussion:

Fixes #1634 

Screenshot 2025-07-08 at 3 40 05 PM

### DIFF
--- a/app/controllers/contents_controller.rb
+++ b/app/controllers/contents_controller.rb
@@ -28,9 +28,13 @@ class ContentsController < ApplicationController
   def update
     authorize! @content
 
-    update_files
-
-    head :ok
+    if valid?
+      update_files
+      head :ok
+    else
+      render json: { error: 'Unable to upload file due to: Duplicate file. Delete the file and try again.' },
+             status: :unprocessable_entity
+    end
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/controllers/contents_controller.rb
```diff
@@ -28,9 +28,13 @@ def show_table
...
     authorize! @content
 
-    update_files
-
-    head :ok
+    if valid?
+      update_files
+      head :ok
+    else
+      render json: { error: 'Unable to upload file due to: Duplicate file. Delete the file and try again.' },
```
justinlittman, line 35: Multiple files will be uploaded at a time -- this doesn't indicate which file is the dupe.

justinlittman (commented):
The problem with the approach is that it doesn't account for files being uploaded in batches. I believe it will incorrectly indicate that all files in the batch are dupes if a single is a dupe.

PR #1649, opened 2025-07-09
Remove left/right padding from icon buttons

### LINKED ISSUE #335: Collections details page - viz design refinement
Can we move the trash icons closer to the field they delete?

On stage:
image

Proposed:
image

mjgiarlo:
Fixes #335

This commit effectively reduces the amount of horizontal padding between icon buttons and adjacent fields, having the desired effect of bringing the field delete button and the text field for repeatable components (e.g., contact email) closer. It also removes the horizontal padding of the move up and move down buttons, which I checked out in the work edit screen for authors, and it does not meaningfully impact that part of the UI as the delete and move buttons are arranged vertically there.

Image

### DIFF
--- a/app/components/elements/delete_button_component.rb
+++ b/app/components/elements/delete_button_component.rb
@@ -4,7 +4,7 @@ module Elements
   # Component for a delete button
   class DeleteButtonComponent < IconButtonComponent
     def initialize(label: 'Clear', **args)
-      super(icon: :delete, label:, **args)
+      super(icon: :delete, label:, classes: Array(classes) + ['px-0'], **args)
     end
   end
 end


justinlittman (changes requested):
I believe this will add the layout globally. The class should just be provided for this particular use of IconButtonComponent.

mjgiarlo:
@justinlittman 💬 
> I believe this will add the layout globally. The class should just be provided for this particular use of IconButtonComponent.

AFAICT, this will affect delete buttons, move up buttons, and move down buttons, which are used in date inputs, in orderable fields (like work authors), and in simple repeatable fields like contact emails. I've checked out all three kinds of fields in both the work and collection contexts, and they all appear fine to my eye. I could move the change directly to the `DeleteButtonComponent` if it makes a meaningful difference.

Maybe @astridu and @amyehodge could review the change in a deployed environment and evaluate the overall impact of the change?

mjgiarlo:
@astridu 💬 
> Which environment should I check?

It's in QA now. Thank you!

astridu:
I checked all the instances in the item and collection forms where there is a trash can and the ordering arrows, and all looks good. Thank you!

mjgiarlo:
@justinlittman 💬 
> Disagree. This component is already designed to have classes passed into it for exactly this purpose. Hard-coding in classes instead of passing them in is incurring debt and likely to confuse a future user of this component and/or create additional work. Better to pass in the classes now while the uses are understood rather than try to figure it out later.

Alrighty, I moved the `px-0` class into the `DeleteButtonComponent`.

app/components/elements/delete_button_component.rb
```diff
@@ -6,5 +6,9 @@ class DeleteButtonComponent < IconButtonComponent
     def initialize(label: 'Clear', **args)
       super(icon: :delete, label:, **args)
     end
+
+    def classes
+      merge_classes(super, @classes, %w[px-0])
```
mjgiarlo, line 11: @justinlittman Including `@classes` here is not sufficient to pull in the classes merged via the parent class, `IconButtonComponent`, and when I tested, `@classes` was an empty array. I added `super` here to make sure the `border` and `border-0` classes are injected properly via the parent. Not sure if this is a pattern we want to be using or not.

justinlittman: I'd be inclined to handle it in the initializer:
```
def initialize(label: 'Clear', classes: [], **args)
      super(icon: :delete, label:, classes: Array(classes) + ['px-0'], **args)
    end
```

How does that seem?

PR #1650, opened 2025-07-09
Do not disable dropzone if globus cancelled

### LINKED ISSUE #1642: Need a way to get back to drag and drop when Globus canceled
If the user decides not to use Globus after initially selecting it, they need to be able to get back to the drag and drop option. Instead they are left with the drag and drop area greyed out.

Steps to reproduce:
1. Open an existing deposited item.
2. Click "Use Globus to transfer files"
3. In the pop-up error message, click "Cancel"
4. Drag and drop stays greyed out and user can't upload files.

Image

aaron-collier:
Fixes #1642

### DIFF
--- a/app/components/works/edit/globus_button_component.rb
+++ b/app/components/works/edit/globus_button_component.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Works
+  module Edit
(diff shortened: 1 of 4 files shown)


aaron-collier:
@justinlittman for sure - on it. Thanks.

justinlittman (approved):

PR #1672, opened 2025-07-16
Prevent hitting enter in keywords from submitting form.

### LINKED ISSUE #1671: Exclude keyword entry field from enter=save HTML behavior
To reproduce problem:
1. Create or edit a deposit
2. Type text into the keyword entry field
3. Instead of selecting from the dropdown list or clicking elsewhere, hit enter to save the text you entered as a keyword
4. The deposit will be saved as a draft OR you'll be bumped to the deposit page with an error message if you didn't complete all the required fields.

Since hitting enter is a common way to submit text, the default enter = save as draft behavior should be suppressed for this field. Hitting enter should have the same effect as clicking away from the form field: the dropdown list disappears, the entered text value is preserved in the keyword field, and the field is no longer active.

I would argue that enter=save as draft is unexpected behavior generally for a multi-tabbed form, but keyword is the only field where it's tripped me up in testing.

justinlittman:
closes #1671


compose.yaml
```diff
@@ -27,6 +27,11 @@ services:
       DATABASE_PASSWORD: postgres
       DATABASE_HOSTNAME: db
       DATABASE_PORT: 5432
+      WORKFLOW_DATABASE_NAME: workflow-server
```
justinlittman, line 30: Unrelated, opportunistic fix.

PR #1678, opened 2025-07-17
move admin search for admin dashboard

### LINKED ISSUE #1658: Move search for user to Admin dashboard main page
If this isn't too complicated, I like to move the search for user to the main dashboard page with the collection and druid searches, and removed from the dropdown list. I'm finding that I'm doing this a lot and it would be convenient to have it there.

So, added to this page wherever is easiest:

Image

peetucket:
Fixes #1658 - move search of user so that it can be rendered on the admin dashboard home page

Renamed the controller and routes to make it consistent with the other search functions found on the admin dashboard

Some examples of how it looks:

Screenshot 2025-07-17 at 4 55 12 PM

Screenshot 2025-07-17 at 4 55 18 PM

Screenshot 2025-07-17 at 4 55 25 PM


app/forms/admin/druid_search_form.rb
```diff
@@ -24,7 +24,7 @@ def normalize_druid
     def collection_or_work_present
       return if collection.present? || work.present?
 
-      errors.add(:druid, 'not found')
+      errors.add(:druid, 'druid not found')
```
peetucket, line 27: made this change to match with the new `user not found` message i added -- seems clearer to me

app/views/admin/dashboard/show.html.erb
```diff
@@ -7,7 +7,6 @@
 
 <div class="d-flex justify-content-end">
   <%= render Admin::FunctionDropdownComponent.new do |component| %>
-    <% component.with_function(label: 'Search for user', link: admin_users_path) %>
```
peetucket, line 10: no longer in the admin menu

justinlittman (commented):
I'm not following why this required multiple turbo frames and the use of turbo streams. Wondering whether this can't be done with a single turbo frame and no streams?

Happy to discuss.

peetucket:
> I'm not following why this required multiple turbo frames and the use of turbo streams. Wondering whether this can't be done with a single turbo frame and no streams?
> 
> Happy to discuss.

i removed an extra turob-frame in the admin dashboard view that is no longer needed (i started out with it and then it's use was removed as work progressed but I forgot to take out it of the view)

justinlittman (approved):

PR #1679, opened 2025-07-18
Updates globus.

### LINKED ISSUE #1673: Globus permissions are inconsistent
In testing whether users besides the depositor can access the Globus shared folder, a difference was found in two distinct scenarios that seem like they should behave the same.

Ideally, collection managers and users with whom a work has been shared would all be able to get to the Globus shared folder.

Scenario 1: shared user -- the behavior seen here is the desired behavior

1. Create a work and share it (view, edit, and deposit permissions) with a user who does not have admin privileges in H3 or Globus. My example shared work was https://sul-h3-stage.stanford.edu/works/druid:dd822bk3521.
2. The person helping me with testing was able to view this work, open the form to edit, and click on the button to use Globus for the deposit.
3. This opened the Globus UI and she was able to see the files that I had already deposited to the work and delete a file.
4. She then returned to H3 and indicated Globus transfer was complete and the files were retrieved from Globus with the deleted file now removed. She was then able to save this as a draft.

Scenario 2: manager -- the behavior seen here is not the desired behavior

1. I had my tester create a new collection on stage and add me as a depositor. 
2. I then created this draft work in that collection: https://sul-h3-stage.stanford.edu/dashboard  
3. Tester then clicked to that item from the deposit list, opened the form, and clicked on the link to use Globus to transfer files. When Globus opened she saw this:

Image

On the off chance this is related to this somehow, she also could not see the "Create a new collection" button on stage in Chrome but could in Firefox. She could see it in Chrome in production though. Very odd.

justinlittman:
closes #1673

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -281,7 +281,7 @@ GEM
       raabro (~> 1.4)
     globalid (1.2.1)
       activesupport (>= 6.1)
-    globus_client (0.17.0)
+    globus_client (0.18.0)
       activesupport (>= 4.2)
       faraday
       faraday-retry
--- a/app/jobs/globus_setup_job.rb
+++ b/app/jobs/globus_setup_job.rb
@@ -18,12 +18,10 @@ class GlobusSetupJob < ApplicationJob
 
   def mkdir
     # This will ignore if the directory already exists.
-    path = GlobusSupport.work_path(work:)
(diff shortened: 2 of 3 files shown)


peetucket (approved):
I trust it works, but I'm not sure I understand how why `with_uploads_directory: true` is no longer needed.

PR #1684, opened 2025-07-18
add delete all files in a work link

### LINKED ISSUE #1614: Make it easier for users to delete all the files on a deposit at one time
Currently if a user wants to delete all the files, they need to click on the trash can for each one. There should be a mechanism to delete all the files, like a button.

peetucket:
Fixes #1614 - delete all files with one link.  I put a javascript confirmation dialog box in front of this even though it wasn't indicated in the ticket.

Note: one thing that confused me was that the delete doesn't actually "stick" unless you save the work.  This is also true of files deleted one by one.  I presume this is expected behavior (but it confused me during testing).

Screenshot 2025-07-18 at 3 40 25 PM

### DIFF
--- a/app/views/contents/_show.html.erb
+++ b/app/views/contents/_show.html.erb
@@ -1,6 +1,6 @@
 <%= turbo_frame_tag content, 'show' do %>
   <%= turbo_stream_from 'show', content %>
-  <% if total_files > Settings.search.file_search_box_min # there is at least this files in the object, so show search bar with section header %>
+  <% if total_files > Settings.search.file_search_box_min # there is at least this many files in the object, so show search bar with section header %>
       <div class="row">
         <div class="col-3">
           <%= render Edit::SectionHeaderComponent.new(title: t('works.edit.panes.files.list')) %>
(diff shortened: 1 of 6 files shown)


app/views/contents/_show.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= turbo_frame_tag content, 'show' do %>
   <%= turbo_stream_from 'show', content %>
-  <% if total_files > Settings.search.file_search_box_min # there is at least this files in the object, so show search bar with section header %>
+  <% if total_files > Settings.search.file_search_box_min # there is at least this many files in the object, so show search bar with section header %>
```
peetucket, line 3: fixed typo in comment

spec/system/manage_files_spec.rb
```diff
@@ -89,12 +90,14 @@
...
 
-      expect(page).to have_text('Your files will appear here once they have been uploaded.')
+      expect(page).to have_text(no_files_message)
 
       # Add one file
       find('.dropzone').drop('spec/fixtures/files/hippo.png')
       await_upload
 
+      expect(page).to have_no_text(no_files_message)
+
```
peetucket, line 100: the message about not having files goes away after a file is uploaded

spec/system/manage_files_spec.rb
```diff
@@ -156,6 +159,12 @@
       expect(page).to have_field('Hide this file', checked: true)
       sleep 0.25 # Wait for the form to submit.
       expect(content_file.reload.hide).to be true
+
+      # Delete all of the files
+      accept_confirm do
+        click_link_or_button('Delete all files')
+      end
+      expect(page).to have_text(no_files_message)
```
peetucket, line 167: now we delete the files and the message comes back

PR #1686, opened 2025-07-21
Allow only first name if coming from ORCID

### LINKED ISSUE #1130: Support an author/contributor with only one name
This came up in testing for a fake author entry. However, we know there at least was at one point a student in physics with one name (not sure if they are still here). If we could come up with a way to support this it would be nice. 

Proposal: do not require both names if the user has entered an ORCID iD. See more info in comment below.

aaron-collier:
Fixes #1130 

Screenshot 2025-07-21 at 1 29 39 PM


lwrubel:
sul_orcid_client can handle only a forename when creating contributor data for an ORCID work, so we're good there, too.

PR #1694, opened 2025-07-22
Show additional information from RoR

### LINKED ISSUE #1690: Show location information for ROR results
Because there are instances where multiple institutions share the same name, we should display additional information from ROR along with the primary name. Top priority would be location, but organizational type and name variations could be useful as well.

See the [Figma designs](https://www.figma.com/proto/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=1116-3906&node-type=frame&t=VohocZAmtVluyXwc-0&scaling=min-zoom&content-scaling=fixed&page-id=1092%3A2227&starting-point-node-id=1114%3A442&show-proto-sidebar=1) (scroll to the right) that include this:

Image

And info on this at ROR: https://ror.readme.io/docs/forms

peetucket:
fixes #1690 - show additional organizational info from RoR lookups in type ahead

We can play with the display.  Using the RoR website examples, I went with this for now:

**Primary Name**
City, Country
(org types)
_aliases_

Aliases are limited to the first three, during my testing I noticed some orgs (like SLAC) have a lot of long aliases, which makes the display look pretty bad.

HOLD for PO review

Some examples below:

Screenshot 2025-07-22 at 4 52 19 PM

Screenshot 2025-07-22 at 4 52 42 PM

Screenshot 2025-07-22 at 4 52 10 PM


app/models/ror_org.rb
```diff
@@ -0,0 +1,20 @@
...
+    @country = data.dig('country', 'country_name')
+    # some orgs have lots of aliases (eg. https://ror.org/02vzbm991), this keeps it sane for display
+    @aliases = data['aliases']&.first(3)&.join(', ')
+    @org_types = data['types']&.join(', ')
+  end
+
+  def location
+    [city, country].compact.join(', ')
+  end
+end
```
peetucket, line 20: created this new model which takes in RoR data that comes back from the API and maps to a simple object -- this makes the view simpler

app/services/ror_service.rb
```diff
@@ -15,7 +15,7 @@ def call
     results = organizations
     return [] if results['number_of_results'].to_i.zero?
 
-    results['items'].map { |org| org.slice('id', 'name') }
+    results['items'].map { |org| RorOrg.new(org.slice('id', 'name', 'aliases', 'addresses', 'country', 'types')) }
```
peetucket, line 18: we need some extra fields from RoR API now, and our service now returns an array of model objects instead of an array of hashes straight from the API

amyehodge:
Screenshots look good.

app/models/ror_org.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Model for RoR organization
```
justinlittman, line 3: Can you add a note about how/where this is used?

app/models/ror_org.rb
```diff
@@ -0,0 +1,20 @@
...
+class RorOrg
+  attr_accessor :id, :name, :city, :country, :aliases, :org_types
+
+  def initialize(data)
```
justinlittman, line 7: Can you document what data is / where it comes from?

app/models/ror_org.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
justinlittman, line 1: I'd make the case for this being an inline class in RorService:
* That is the only place it is used.
* It makes the context for this class clear. (In `models/`, it has no context as to why it exists and what it is used for.)
* It is not an active record model.

(If inline, you can ignore my comments below.)

justinlittman (approved):

app/models/ror_org.rb
peetucket: Inlined Org class with RoRService and also added a couple comments

Issue #1698, opened 2025-07-23
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "contributors" violates foreign key constraint


justinlittman:
## Backtrace

    line 116 of [PROJECT_ROOT]/app/jobs/deposit_collection_job.rb: DepositCollectionJob#assign_contributors
    line 63 of [PROJECT_ROOT]/app/jobs/deposit_collection_job.rb: DepositCollectionJob#update_collection_record
    line 19 of [PROJECT_ROOT]/app/jobs/deposit_collection_job.rb: DepositCollectionJob#perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/128495/faults/122245309)

amyehodge:
I was editing a new collection I just created to add a second affiliation for default author entry. Collection is 232 "End of July test collection." I was adding "Centre College" as the second affiliation.

PR #1699, opened 2025-07-23
foreign key constraint violation with clear

peetucket:
Ref #1698 and https://app.honeybadger.io/projects/128495/faults/122245309

Not sure how what the intent was with clearing the contributors from a collection before adding them again, but just breaking the association by removing the key (which is what `.clear` does) seems to lead to a foreign key violation.  Can you we just destroy them and then re-add them, or is this bad? (i.e. are these contributors potentially referenced by other collections?)

### DIFF
--- a/app/jobs/deposit_collection_job.rb
+++ b/app/jobs/deposit_collection_job.rb
@@ -113,7 +113,7 @@ class DepositCollectionJob < ApplicationJob
   end
 
   def assign_contributors
-    collection.contributors.clear
+    collection.contributors.destroy_all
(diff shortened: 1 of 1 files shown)


justinlittman:
Given https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/models/collection.rb#L17 and [this explanation](https://guides.rubyonrails.org/association_basics.html) of `.clear`, I would expect `.clear` to destroy.

Can you write a spec to reproduce the problem to verify that this fixes?

PR #1700, opened 2025-07-23
ignore whitespace in admin druid search

peetucket:
Because if you copy and paste a druid from a URL or webpage into the admin druid search textbox, you may get extra whitespace and then it'll tell you it can't find the druid, which is a LIE

### DIFF
--- a/app/forms/admin/druid_search_form.rb
+++ b/app/forms/admin/druid_search_form.rb
@@ -16,6 +16,8 @@ module Admin
     end
 
     def normalize_druid
+      druid.strip! # remove any trailing or leading whitespace the user may have inadvertently entered via a copy/paste
+
       return if druid.starts_with?('druid:') || druid.blank?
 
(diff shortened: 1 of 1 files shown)


app/forms/admin/druid_search_form.rb
```diff
@@ -16,6 +16,8 @@ def work
     end
 
     def normalize_druid
+      druid.strip! # remove any trailing or leading whitespace the user may have inadvertantly entered via a copy/paste
```
justinlittman, line 19: Typo: "inadvertently"

justinlittman (approved):
Approve pending fixing typo.

app/forms/admin/druid_search_form.rb
peetucket: your high school english teacher should be proud of this catch

peetucket:
typo fixed

PR #1745, opened 2025-08-13
raise standarderror

peetucket:
see https://app.honeybadger.io/projects/128495/faults/122702166

~~I think we want to raise `StandardError` instead to get better HB alerts?~~

Just let faraday exception go to HB if it happens

### DIFF
--- a/app/services/ror_service.rb
+++ b/app/services/ror_service.rb
@@ -25,8 +25,6 @@ class RorService
 
   def organizations
     conn.get('/v2/organizations', params, headers).body
-  rescue Faraday::Error => e
-    raise StandardError, "RoR Service connection error: #{e.message}"
   end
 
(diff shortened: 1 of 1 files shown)


app/services/ror_service.rb
```diff
@@ -26,7 +26,7 @@ def call
   def organizations
     conn.get('/v2/organizations', params, headers).body
   rescue Faraday::Error => e
-    raise Error, "Connection err: #{e.message}"
+    raise StandardError, "RoR Service connection error: #{e.message}"
```
justinlittman, line 29: If we're not going to handle this error in any way, then there is no reason to rescue the Faraday::Error at all.

justinlittman: Yup.

PR #1750, opened 2025-08-15
Remove duplicates from works list in dashboard collection table

lwrubel:
Dashboard query to populate the works in a collection was including some works multiple times in the result. This resulted in the user seeing the same work multiple times in the dashboard table. Only non-admins or people who are not managers/reviewers of the collection would see this result. The work also had to have been shared with others.

### DIFF
--- a/spec/policies/work_policy_spec.rb
+++ b/spec/policies/work_policy_spec.rb
@@ -240,7 +240,7 @@ RSpec.describe WorkPolicy do
       policy.apply_scope(target, name: :collection, type: :active_record_relation, scope_options: { collection: }).to_a
     end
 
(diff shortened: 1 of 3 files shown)


spec/policies/work_policy_spec.rb
```diff
@@ -240,7 +240,7 @@
       policy.apply_scope(target, name: :collection, type: :active_record_relation, scope_options: { collection: }).to_a
     end
 
-    let!(:unowned_work) { create(:work, collection:, title: 'unknowned') }
```
peetucket, line 243: i guess this means you got dumber?

peetucket (approved):
excellent!

jmartin-sul (approved):

PR #1804, opened 2025-11-25
Refactor ActiveModel::NestedAttributes

mjgiarlo:
This commit refactors our nested form model implementation to reduce complexity, to use/re-use more functionality that already exists in ActiveModel, and to optimize for the principle of least surprise, aligning the implementation with the behavior of ActiveRecord nested models.

### DIFF
--- a/app/components/elements/forms/repeatable_nested_component.rb
+++ b/app/components/elements/forms/repeatable_nested_component.rb
@@ -10,7 +10,7 @@ module Elements
       def initialize(form:, model_class:, field_name:, form_component:, hidden_label: false, bordered: true, # rubocop:disable Metrics/ParameterLists, Metrics/MethodLength, Metrics/AbcSize
                      reorderable: false, single_field: false, fieldset_classes: [], skip_tooltip: false,
                      fieldset_id: nil, hide_add_button: false, add_button_data: {}, column_classes: ['col'],
-                     separated: false, nested_buttons_classes: [])
+                     separated: false, nested_buttons_classes: [], render_empty: true)
         @form = form
         @model_class = model_class
         @field_name = field_name
@@ -33,6 +33,7 @@ module Elements
         @column_classes = column_classes
         @separated = separated
         @nested_buttons_classes = nested_buttons_classes
+        @render_empty = render_empty
         super()
       end
 
@@ -47,6 +48,10 @@ module Elements
         return if skip_tooltip?
 
         helpers.t("#{field_name}.edit.tooltip_html", default: nil)
+      end
+
+      def render_empty?
+        @render_empty
       end
 
       def skip_tooltip?
@@ -117,6 +122,12 @@ module Elements
       # A form component can provide a delete button label by providing a `form_button_label` method.
       def delete_button_label(form_component_instance:)
         form_component_instance.try(:delete_button_label) || 'Clear'
+      end
+
+      def template_model
+        model_class
+          .new
+          .then { |form_instance| form_instance.try(:prepopulate) || form_instance }
       end
     end
   end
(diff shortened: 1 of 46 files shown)


app/forms/collection_form.rb
```diff
@@ -1,30 +1,36 @@
 # frozen_string_literal: true
 
 # Form for a Collection
-class CollectionForm < ApplicationForm
+class CollectionForm < RefactoredApplicationForm
   accepts_nested_attributes_for :related_links, :contact_emails, :managers, :depositors, :reviewers, :contributors
+
   before_validation do
-    blank_managers = managers_attributes.select(&:empty?)
```
mjgiarlo: The prior implementation supported readers for nested attributes both with and without the `_attributes` suffix, which was a departure from ActiveRecord. Now the `..._attributes` form is only supported for writes (the `#..._attributes=` form instance method).

spec/system/manage_contributors_spec.rb
```diff
@@ -336,7 +336,10 @@
 
       # There is a single contributor form
       form_instances = all('.form-instance')
-      expect(form_instances.count).to eq(4)
+
+      # NOTE: I don't know why this number should be 4 (and is now 5), so I can't tell if this is problematic.
```
justinlittman: How do you think this should be handled?

spec/system/validate_work_deposit_spec.rb
```diff
@@ -115,7 +118,9 @@
 
     # Add a blank contributor. It will be ignored.
     click_link_or_button('Add another contributor')
-    expect(page).to have_css('.form-instance', count: 3)
+    # NOTE: I don't know why this number should be 3 (and is now 4), so I can't tell if this is problematic.
+    #
+    # expect(page).to have_css('.form-instance', count: 3)
```
justinlittman: Same.

app/forms/affiliation_form.rb
```diff
@@ -11,10 +11,6 @@ class AffiliationForm < ApplicationForm
   attribute :uri, :string
   attribute :department, :string
 
-  def empty?
```
mjgiarlo: A base implementation of `#empty?` is now provided by `ApplicationForm`.

app/presenters/collection_presenter.rb
```diff
@@ -90,7 +95,17 @@ def created_datetime
...
+
+  def related_links
+    return ['None provided'] if collection_form.related_links.blank?
+
+    collection_form.related_links.filter_map do |related_link|
+      next if related_link.url.blank?
+
+      link_to_new_tab(related_link.text || related_link.url, related_link.url)
+    end
   end
```
mjgiarlo: This method moved from the form presenter into the collection presenter because works no longer have related links.

lib/active_model/validations/associated_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+      end
+
+      private
+
+      def valid_object?(record, context)
+        (record.respond_to?(:marked_for_destruction?) && record.marked_for_destruction?) || record.valid?(context)
+      end
+    end
+  end
+end
```
mjgiarlo: Re-use ActiveRecord validations for nested ("associated") models and adapt it to ActiveModel. This lets us remove custom validation logic from the NestedAttributes module.

app/components/elements/forms/repeatable_nested_component.rb
```diff
@@ -118,6 +118,12 @@ def hide_delete_button?(form_component_instance:)
       def delete_button_label(form_component_instance:)
         form_component_instance.try(:delete_button_label) || 'Clear'
       end
+
+      def template_model
+        model_class
+          .new
+          .then { |form_instance| form_instance.try(:prepopulate) || form_instance }
```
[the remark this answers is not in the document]
mjgiarlo: @justinlittman Yes, because `#try` returns `nil` when the method is not invoked, and we need this to return a form instance.

PR #1811, opened 2025-12-04
Do not raise if deleting content AR doesn't succeed in DeportWorkJob

justinlittman:
Deleting isn't actually required and should not fail the job.

Related: https://app.honeybadger.io/projects/128495/faults/125442437


justinlittman:
@edsu It gets cleaned up by https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/config/schedule.rb#L23-L25

PR #1920, opened 2026-02-13
lookup and store name when adding participant

### LINKED ISSUE #1737: Display name of depositor added
Adan reported that when he added annisb as a depositor to the collection (tn741mc5549), the name does not display. We thought this might be an issue with the porting of old data (maybe she's already in there from H2?), but I just noticed in the history that he tried adding and removing this person several times and every time the history table lists her full name in the comments for who is added but only her SUNet ID in the comments for who was removed. If we have the name, I don't understand why it's not displaying the depositor list or in the list of names removed.

In the image below you can see the display of "annisb (annisb)" near the bottom of the depositors list, but then the history table entries as well. I have not gotten any report that this is causing the user any problems (yet).

Image

peetucket:
(sort of) fixes #1737

The code currently looks up and stores in the database both the name and sunetid of any user added as a participant.  This name should then appear in the history and on the webpage.

However, the name could be missing in a couple scenarios:

1. The user was first added as a participant to any collection when the account service was not working correctly.  This would add them to the database with no name, and once they are added to the database with no name, we do not currently ever go back to lookup their name again (unless they log into SDR themselves).
2. The user was ported from H2 and has never logged into H3.

In these cases (the user exists in our database without a name and is then added to a different collection), we won't know their name and won't look it up again, resulting in the described behavior.

This code change deals with this situation by storing their name in the database when they get added to a collection if we realize we don't currently have their name stored.  So it "heals" the situation the next time this participant is added anywhere.  It will also be "healed" if they log into H3 (no more action needed, we already do this).

Separately and in addition, we can run a query to find all users who seem to be missing a name, look them up immediately, and then store their name.  I will write that out as a one time remediation and place in a ticket comment.

### DIFF
--- a/spec/jobs/deposit_collection_job_spec.rb
+++ b/spec/jobs/deposit_collection_job_spec.rb
@@ -95,6 +95,54 @@ RSpec.describe DepositCollectionJob do
         expect(Sdr::Event).not_to have_received(:create)
       end
     end
+
+    context 'when a participant name is blank and a name is provided' do
+      let(:existing_user) { create(:user, email_address: 'jdoe@stanford.edu', name: '') }
+      let(:collection_form) do
+        CollectionForm.new(title: collection.title, managers_attributes:, depositors_attributes:)
+      end
+      let(:managers_attributes) { [{ sunetid: existing_user.sunetid, name: 'Jane Doe' }] }
+      let(:depositors_attributes) { [] }
+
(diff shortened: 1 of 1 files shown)


justinlittman (approved):
Approve pending green.

peetucket:
added to spec to get coverage patch to 100%

Issue #672, opened 2025-02-07
White vertical strip and photo credit on homepage


astridu:
Amy is reporting a white vertical strip appearing on the left side of the homepage in her browser (Mac Sequoia 15.3 and Chrome). I can't reproduce in Chrome or Safari, but it is persistent for her.

![Image](https://github.com/user-attachments/assets/50cce222-e788-482b-9e52-1bfde0700db9)

Also, the photo credit gets cut off when we make the browser window narrower. This happens for both Amy and me. 

![Image](https://github.com/user-attachments/assets/649fe833-306d-4b7e-840a-ead00d0d5fa3)

justinlittman:
@astridu The credit is part of the image. How would you like this handled?

astridu:
Oh sorry. I didn't realize Figma exported the text too. I will update the images w/o the text. Can we add the text so it's text and not part of the image?

justinlittman:
Not without moderate complexity. Right now all H3 needs to know is the number of images. Adding the text would mean we would need a mapping from each image to its photo credit.

amyehodge:
Noting here that Stella also sees this issue on Windows 10 with Firefox 135, and in Chrome on the same machine. This is the Firefox screenshot.

<img width="1156" alt="Image" src="https://github.com/user-attachments/assets/a439f793-c76a-4993-9b05-67323682d21e" />

justinlittman:
Can this ticket be split into separate tickets for vertical strip and photo credits?

amyehodge:
@andrewjbtw OMG I really hope this is the source of the problem and that it's an easy fix. THANK YOU!

justinlittman:
@amyehodge Can you verify if this is fixed?

amyehodge:
@justinlittman Thanks. The shifting on the deposits page is now not happening, which is awesome.

The one thing I still see is that when the scrollbar is set for "automatically based on mouse or trackpad" that there is a little shifting on other pages, specifically the Manage files and the Access settings pages. See the video below.

https://github.com/user-attachments/assets/9650b3c7-d74f-42ec-8e05-43b885a11b62

PR #1947, opened 2026-02-19
Fix container-fluid. Remove setting left margin on body.

justinlittman:
refs #672


justinlittman:
I can't reproduce on FF or Chrome.

mjgiarlo:
@justinlittman ship it

mjgiarlo (approved):

PR #1956, opened 2026-02-19
Adds license and terms of deposit to article form.

### LINKED ISSUE #1931: Implement license and terms of deposit for article deposit
Image

But not version of article.

Terms of deposit should only be rendered if user hasn't accepted terms of deposit.

justinlittman:
closes #1931

image


justinlittman:
Yes, unfortunately the layout is different.

Issue #1948, opened 2026-02-19
Article edit form


justinlittman:
The article edit form should be adjusted to account for the relaxed validation requirements. In particular, the following are required for a Work, but not required for an Article:
* Contact
* Authors / contributors
* Abstracts
* Keywords

When using an article edit form, the user should be able to submit the article in which these are absent.

In addition, the following tabs should be marked as optional (append "(optional)") for the article edit form:
* Authors / contributors
* Abstract and keywords

and the pane titles for those tabs should be marked as optional:
* Authors / Contributors
* Describe your deposit

Also, the following fields should be marked as optional:
* Contact email
* Abstract
* Keywords

In the current implementation, much of this is hard-coded (e.g., "(optional)" is in the label in `en.yml`). Ideally, whether tabs / panes / fields are required should be able to be specified in the form view (e.g., `article_form.html.erb`).

PR #1962, opened 2026-02-20
Make optional / required a configuration for tabs and panes.

justinlittman:
refs #1948


justinlittman:
Sort of. They have an aria set when required, but the label isn't automatically adjusted. There will be a separate PR for that, but wanted to keep the scope of this PR limited.

PR #1966, opened 2026-02-20
strip html/xml tags from crossref metadata

### LINKED ISSUE #1957: Remove tags and XML comments from CrossRef data
For example:

Image

Tags and comments should be removed from all fields.

peetucket:
Fixes #1957

### DIFF
--- a/app/services/crossref_service.rb
+++ b/app/services/crossref_service.rb
@@ -20,8 +20,8 @@ class CrossrefService
   def call
     Rails.cache.fetch(doi, namespace: 'crossref', expires_in: 1.month) do
       {
-        title: message['title']&.first,
-        abstract: normalize_abstract(message['abstract']),
+        title:,
+        abstract:,
         related_works_attributes: [{
           relationship: 'is version of record',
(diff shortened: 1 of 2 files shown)


app/services/crossref_service.rb
```diff
@@ -76,9 +94,14 @@ def publication_date_attrs
...
-            .gsub(%r{</?jats:.+?>\n?\s*}, '')
-            .strip
+  # Strips all HTML/XML tags and comments from text
+  # @param text [String, nil] the text to sanitize
+  # @return [String, nil] the sanitized text, or nil if input is nil
+  def strip_tags_and_comments(text)
+    return nil if text.nil?
+
+    text.gsub(/<!--.*?-->/m, '') # Remove XML/HTML comments
+        .gsub(/<[^>]+>/, '').squeeze(' ') # Collapse multiple spaces (but not newlines) into one
```
peetucket: this could result in actual user text with multiple spaces being collapsed to a single space, but if you don't do this, it could result in some odd extra spaces that come from stripping the tags

Issue #1837, opened 2026-01-21
Change form to specify which fields are required instead of which are optional


amyehodge:
@astridu and I reviewed this today and we think there are no longer any impediments to changing the form so that instead of indicating which fields are optional, we note with an asterisk which fields are required. Mockups for the deposit and collection forms are in progress, and we'll design the easy deposit/OA form and the abbreviated form for GitHub deposits to follow this pattern as well.

astridu:
We definitely need to test with users!

PR #2019, opened 2026-02-25
Switch to marking required.

justinlittman:
refs #1837

image


mjgiarlo (approved):

amyehodge:
Where: Article deposit longer form.
When: always
Change: authorls should not have the asterisk next to the header and in the navigation. But I'm wondering if we have implemented this requirement anywhere for this?


Screenshot 2026-02-26 at 1 12 22 PM

PR #2025, opened 2026-02-25
lookup doi with pmid/pmcid

### LINKED ISSUE #1992: Support article deposit by PMID, PMCID
Image

DOI lookup should be changed so that:
* Determine if a DOI or a PMID / PMCID. (DOI will have at least one "/" in it.)
* If a DOI, normalize by removing "https://doi.org/" prefix and perform DOI lookup.
* If a PMID / PMCID, normalize and perform PMID / PMCID lookup get DOI and then DOI lookup.
* If any lookups fail do to not being found, return the following validation error:
> We were unable to retrieve metadata for this identifier. Please double check the
identifier, and if you still receive an error, use the full form to deposit your item.

If possible, link "full form" to `new_work_path(collection_druid: @collection.druid)` (but might not be possible in a validation message).

For info on PMID / PMCID lookup, see https://github.com/sul-dlss/hungry-hungry-hippo/issues/1941

### LINKED ISSUE #2040: Improve error messages for DOI lookup
Currently, our error messaging for a failed DOI search looks like this:

Image

It would be more helpful to say "Unable to retrieve metadata for this DOI/PMID/PMCID. You will need to use the "Deposit to this collection" button to deposit this work."

peetucket:
Fixes #1992 and fixes #2040

- Allow article deposit to work with PMID and PMCID.  If entered, will lookup DOI with Pubmed API first, then proceed as normal.
- Altered UI and messaging to match design and improve error messaging a bit
- Altered display of resulting publication to show DOI
- Add separate form field to track identifier separately from DOI (since they can be different and are both shown to the user if different, even if we are only storing the DOI eventually)

Todo:
- [x] if user enters a PMID or PMCID, keep this value in the form and then show the resulting DOI in the result (already done); currently if a DOI is found, it will be replace the pmid/pmcid the user entered in the input box, which may be confusing
- [x] Add additional tests for the article deposit form
- [x] Verify that Pubmed API being down or throwing an error doesn't break our app (just causes a DOI lookup to fail, but still allows straight DOI lookup)

Screenshot 2026-02-25 at 3 51 22 PM

### DIFF
--- a/app/forms/article_form.rb
+++ b/app/forms/article_form.rb
@@ -71,11 +71,22 @@ class ArticleForm < ApplicationForm
   private
 
   def doi_article
-    return errors.add(:identifier, 'identifier was not found') unless doi_found?
+    use_full_form_message = 'You will need to use the "Deposit to this collection" button to deposit this work.'
 
-    return errors.add(:identifier, 'identifier is not a journal article') unless doi_journal_article?
+    unless doi_found?
+      return errors.add(:identifier,
+                        "Unable to retrieve metadata for this DOI/PMID/PMCID. #{use_full_form_message}")
+    end
 
-    errors.add(:identifier, 'identifier does not have a title') unless doi_has_title?
+    unless doi_journal_article?
+      return errors.add(:identifier,
+                        "The metadata for this identifier indicates it is not a journal article. #{use_full_form_message}") # rubocop:disable Layout/LineLength
+    end
+
+    return if doi_has_title?
+
+    errors.add(:identifier,
+               "The metadata for this identifier does not include a title. #{use_full_form_message}")
   end
 
   def doi_ok?
--- a/spec/forms/article_form_spec.rb
+++ b/spec/forms/article_form_spec.rb
@@ -6,6 +6,10 @@ RSpec.describe ArticleForm, type: :form do
   let(:form) { described_class.new(identifier:) }
 
   let(:identifier) { '10.1234/nonexistent' }
+  let(:base_error) { 'You will need to use the "Deposit to this collection" button to deposit this work.' }
+  let(:not_found_error) { "Unable to retrieve metadata for this DOI/PMID/PMCID. #{base_error}" }
+  let(:not_article_error) { "The metadata for this identifier indicates it is not a journal article. #{base_error}" }
+  let(:no_title_error) { "The metadata for this identifier does not include a title. #{base_error}" }
 
   describe 'identifier validations' do
     context 'when identifier is not present' do
@@ -46,7 +50,7 @@ RSpec.describe ArticleForm, type: :form do
 
       it 'is not valid' do
         expect(form).not_to be_valid
-        expect(form.errors[:identifier]).to include('identifier was not found')
+        expect(form.errors[:identifier]).to include(not_found_error)
       end
     end
 
@@ -57,7 +61,7 @@ RSpec.describe ArticleForm, type: :form do
 
       it 'is not valid' do
         expect(form).not_to be_valid
-        expect(form.errors[:identifier]).to include('identifier is not a journal article')
+        expect(form.errors[:identifier]).to include(not_article_error)
       end
     end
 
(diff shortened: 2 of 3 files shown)


app/forms/article_form.rb
```diff
@@ -61,11 +69,11 @@ def lookup_performed?
...
 
   def doi_article
-    return errors.add(:doi, 'not found') unless doi_found?
+    return errors.add(:identifier, 'identifier was not found') unless doi_found?
 
-    return errors.add(:doi, 'is not a journal article') unless doi_journal_article?
+    return errors.add(:identifier, 'identifier is not a journal article') unless doi_journal_article?
 
-    errors.add(:doi, 'does not have a title') unless doi_has_title?
+    errors.add(:identifier, 'identifier does not have a title') unless doi_has_title?
```
peetucket: tweaks to the error messages on the input form to make it a bit more clear

app/services/pubmed_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    response.body['records'].first['doi']
+  end
+
+  def doi_missing?(response)
+    response.body['records'].blank? || extract_doi(response).blank?
+  end
+
+  def new_conn
+    Faraday.new({ url: Settings.pubmed.url }) do |f|
+      f.request :json
```
justinlittman: Is this needed?

app/views/articles/form.html.erb
```diff
@@ -46,31 +46,30 @@
...
           </div>
         </div>
 
-        <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name: :doi) %>
+        <%= render Elements::Forms::InvalidFeedbackComponent.new(form:, field_name: :identifier) %>
       </div>
     </div>
 
-    <%= render Articles::Edit::LookupTableComponent.new(article_work_form: @article_work_form, classes: 'mt-3') %>
+    <%= render Articles::Edit::LookupTableComponent.new(article_work_form: @article_work_form, doi: @article_form.doi, classes: 'mt-3') %>
```
justinlittman: Why pass the DOI as an argument is you're passing the article work form?

app/services/pubmed_service.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def headers
+    {
+      'Accept' => 'application/json',
+      'User-Agent' => 'Stanford Self-Deposit (Hungry Hungry Hippo)'
+    }
+  end
+
+  def params
+    { ids: search, format: 'json', tool: 'my_tool', email: Settings.pubmed.email }
```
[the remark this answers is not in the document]
peetucket: nope, will make it more clear

app/services/pubmed_service.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def doi_missing?(response)
+    response.body['records'].blank? || extract_doi(response).blank?
+  end
+
+  def new_conn
+    Faraday.new({ url: Settings.pubmed.url }) do |f|
+      f.request :json
+      f.response :json
+      f.response :raise_error
```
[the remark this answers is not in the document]
peetucket: yup, will adjust custom error handling

spec/forms/article_form_spec.rb
```diff
@@ -3,72 +3,85 @@
...
       end
 
       it 'is not valid' do
         expect(form).not_to be_valid
-        expect(form.errors[:doi]).to include('does not have a title')
+        expect(form.errors[:identifier]).to include('identifier does not have a title')
+      end
+    end
+
+    context 'when identifier is a PMID that is not found' do
```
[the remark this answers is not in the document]
peetucket: or woops, was thinking this was the PubmedService class spec.  so maybe you are thinking an example where the PMID is found?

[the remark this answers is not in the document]
justinlittman: Yes.

app/views/articles/form.html.erb
peetucket: doi is on the article_form, not the article_work_form?

justinlittman: It is is the related works in the article_work_form, I believe.

peetucket: i want to access the DOI returned by the cross service call, not sure i follow how to get that from related_works, maybe we can look together tomorrow

justinlittman (approved):

PR #2149, opened 2026-03-04
Force world access and add note to form

### LINKED ISSUE #2104: Article form: adjust access and DOI settings
We had originally said that articles deposited via the new workflow should always be i

- not get a DOI
- immediately released (no embargoes allowed)
- downloadable by anyone (no "Stanford only")

We've managed to get this set up where we are doing the first two but not the last one.

Here's what we should do here:

- set this up so these are always downloadable by anyone
- remove the "access settings" page from the form
- somewhere in the short form, include a note that warns people that this deposit will be immediately available to anyone in the world once deposited, and will not receive a DOI.

I do see a statement about immediate release to anyone was included by Astrid in the mockup at https://www.figma.com/design/g46Nfl5EHG7DfwqupK3G4X/H3-2024?node-id=8031-5895&t=Gvg9qk688oRCUnjO-4.

We should also add. "It will not receive a DOI."

Image

aaron-collier:
Fixes #2104 

Screenshot 2026-03-04 at 2 39 34 PM


justinlittman:
Wow, it must have been a really long day. This time for an article ...

Still seeing a DOI getting assigned:
image

PR #2282, opened 2026-03-19
Fix bug in PubmedService. Fix VCR config.

justinlittman:


app/services/pubmed_service.rb
```diff
@@ -65,6 +65,6 @@ def params
   end
 
   def ids
-    search.upcase.delete_suffix('PMC')
+    search.upcase.delete_prefix('PMC')
```
aaron-collier: doh!

PR #2292, opened 2026-03-20
Mail tweaks.

### LINKED ISSUE #2291: Email updates
**1. Update manager invite email.** That's this one: https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/views/collections_mailer/manage_access_granted_email.html.erb

The line that currently says
`<p>To visit the collection where deposits are made, and for more information about SDR repository services,
   please visit https://<%= Settings.host %>.</p>`

should be updated to just say

> Visit this collection.

And that entire text should be linked to https://sdr.stanford.edu.

**2. Update the contact component**, which I believe is in https://github.com/sul-dlss/hungry-hungry-hippo/blob/main/app/components/emails/contact_component.rb#L5.

We would like it to say:

> Contact us with questions.

And that entire text should be linked to https://sdr.stanford.edu/contacts/new

**3. Global change to component order.** Whenever the contact component above appears in a message, move it below the signature. So, currently I think most messages currently have this at or near the end:

```
<%= render Emails::ContactComponent.new %>

<%= render Emails::SignatureComponent.new %>
```

and instead I want them to look like this:

```
<%= render Emails::SignatureComponent.new %>

<%= render Emails::ContactComponent.new %>
```

justinlittman:
closes #2291

with assist from copilot CLI.

### DIFF
--- a/app/views/collections_mailer/manage_access_granted_email.html.erb
+++ b/app/views/collections_mailer/manage_access_granted_email.html.erb
@@ -4,11 +4,10 @@
    Stanford Digital Repository. Managers have the ability to manage global settings for the collection,
    manage Depositors and other participants, and edit any deposits in the collection.</p>
 
-<p>To visit the collection where deposits are made, and for more information about SDR repository services,
-   please visit https://<%= Settings.host %>.</p>
-
-<%= render Emails::ContactComponent.new %>
+<p><%= link_to 'Visit this collection.', Settings.host %></p>
 
 <%= render Emails::SignatureComponent.new %>
+
+<%= render Emails::ContactComponent.new %>
 
 <%= render Emails::NewsletterComponent.new %>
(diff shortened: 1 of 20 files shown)


app/views/collections_mailer/manage_access_granted_email.html.erb
```diff
@@ -4,11 +4,10 @@
    Stanford Digital Repository. Managers have the ability to manage global settings for the collection,
    manage Depositors and other participants, and edit any deposits in the collection.</p>
 
-<p>To visit the collection where deposits are made, and for more information about SDR repository services,
-   please visit https://<%= Settings.host %>.</p>
-
-<%= render Emails::ContactComponent.new %>
+<p><%= link_to 'Visit this collection.', Settings.host %></p>
```
peetucket: I presume the dropping of the `https://` here is intentional and we are counting on mail clients to make the link correctly even if `Settings.host` is just `sdr.stanford.edu`: https://github.com/sul-dlss/shared_configs/blob/sul-h3-prod/config/settings/production.yml#L21

PR #2319, opened 2026-04-09
Avoid user version mismatch when sdr version has advanced

### LINKED ISSUE #2317: New public version should only be created for file changes
A depositor reported that he only made metadata changes to sp776gx3909, but a new public version (v3) was created. @andrewjbtw confirmed that no changes were made that should have triggered a new public version (add or remove files, hide or unhide files).

We should look into this. I've never seen this happen before, and I doubt I could replicate it.

aaron-collier:
Fixes #2317

This is a subtle bug that is triggering when there have been SDR version bumps that didn't create new user versions (e.g., technical corrections by an admin), the SDR version will be ahead of the last user version. For example:
  - User version 2 created at SDR v2 → files have version: 2
  - Admin technical correction opens SDR v3 (no new user version)
  - Depositor makes metadata-only change: work_form.version = 3 → mapped_cocina_object has files with version: 3
  - Comparison: version: 3 ≠ version: 2 → UserVersionChangeService returns true → new public version incorrectly created

Removing the version from the `contains_hash_for` should guard against this bug in the future.

### DIFF
--- a/app/services/user_version_change_service.rb
+++ b/app/services/user_version_change_service.rb
@@ -26,9 +26,11 @@ class UserVersionChangeService
     structural.to_h[:contains].tap do |contains_hash|
       contains_hash.each do |file_set_hash|
         file_set_hash.delete(:label)
+        file_set_hash.delete(:version)
         file_set_hash[:structural][:contains].each do |file_hash|
           file_hash.delete(:label)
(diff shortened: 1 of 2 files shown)


justinlittman (approved):
Nice find!

Issue #2308, opened 2026-03-31
Article abstraction UI updates


amyehodge:
See #2307 for details on the testing I was doing here.

I was testing different files repeatedly on the same object without starting over completely from the beginning, i.e. clicking the "Deposit article using DOI,PMID" button. This is likely not typical user behavior, but I did notice something we should consider addressing. 

Once the AI process has run, whether it's told you "content missing" or retrieved an abstract or given you an error message, there's no way to return to the button to try it again. Let's say you realize that you have uploaded the wrong file because you see the retrieved abstract is not the right one. You can delete the abstract using the trash can, but it leaves an empty box behind and the button doesn't show up again. I found the best way to reset this was to click the "Get publication info from DOI/PMCID" button again, though I doubt most users would try this. 

It seems like when the user clicks the trash can for the abstract that this should return to the state with the button to do the abstract extraction again. I'm not sure what this should look like if they get the "Content missing" error.

I'll put some screenshots below.

amyehodge:
This is what it looks like before you click the button to extract the abstract:

<img width="851" height="83" alt="Image" src="https://github.com/user-attachments/assets/e18eac61-5e95-4c35-bdea-91a3c26f9174" />

This is what it looks like after an abstract is extracted:

<img width="850" height="228" alt="Image" src="https://github.com/user-attachments/assets/2609d8d2-d2f9-4381-b5df-89e70232a23e" />

And this is what it looks like if you click the trash can icon:

<img width="861" height="234" alt="Image" src="https://github.com/user-attachments/assets/a0535003-db59-4c17-9a9d-7bc1f0360cfe" />

And this is the content missing message:

<img width="870" height="103" alt="Image" src="https://github.com/user-attachments/assets/b6b523de-807b-4588-9416-5d4e88c4bc82" />

justinlittman:
closed by https://github.com/sul-dlss/hungry-hungry-hippo/pull/2311

PR #2321, opened 2026-04-10
Resets article abstract when user clears.

justinlittman:
refs #2308

https://github.com/user-attachments/assets/f2a1a7d3-fdb4-4787-a506-43ce8a983ec7

### DIFF
--- a/app/components/elements/delete_button_link_component.rb
+++ b/app/components/elements/delete_button_link_component.rb
@@ -2,8 +2,8 @@
 
 module Elements
   # Component for a delete button
-  class DeleteButtonComponent < IconButtonComponent
-    def initialize(label: 'Clear', **args)
+  class DeleteButtonLinkComponent < IconButtonLinkComponent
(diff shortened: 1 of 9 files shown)


app/components/elements/delete_button_link_component.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Elements
+  # Component for a delete button
+  class DeleteButtonLinkComponent < IconButtonLinkComponent
+    def initialize(label: 'Clear', **args)
+      super(icon: :delete, label:, classes: Array(classes) + ['px-0'], **args)
```
aaron-collier: I would suggest moving classes into a method to match our usual `merge_classes` pattern.

aaron-collier (approved):

app/components/elements/delete_button_link_component.rb
justinlittman: Will do.

PR #2325, opened 2026-04-15
rollback deposit state when a github deposit fails

peetucket:
Follow on from #2324 

If something similar happens in the future (starting accessioning or other things fail in the job), this rolls back the deposit state so it can be cleanly retried without manual intervention.

### DIFF
--- a/app/jobs/deposit_github_release_job.rb
+++ b/app/jobs/deposit_github_release_job.rb
@@ -20,7 +20,7 @@ class DepositGithubReleaseJob < ApplicationJob
       return
     end
 
-    update_release_status(status: 'started', status_details: nil)
+    github_release.update!(status: 'started', status_details: nil)
 
     return if check_release_zip_exists!
 
@@ -32,7 +32,10 @@ class DepositGithubReleaseJob < ApplicationJob
 
     github_release.completed!
   rescue StandardError => e
-    update_release_status(status: 'failed', status_details: "error processing release: #{e.message}")
+    github_release.update!(status: 'failed', status_details: "error processing release: #{e.message}")
+
+    # rollback the repository status so it can be tried again
+    github_repository.deposit_clear_fail! if github_repository.can_deposit_clear_fail?
     raise
   end
 
@@ -54,7 +57,7 @@ class DepositGithubReleaseJob < ApplicationJob
     return false if skip_publish_wait
     return false unless github_release.published_at > publish_wait
 
-    update_release_status(status_details: "waiting #{time_ago_in_words(publish_wait)} after publishing")
+    github_release.update!(status_details: "waiting #{time_ago_in_words(publish_wait)} after publishing")
 
     true
   end
@@ -63,15 +66,15 @@ class DepositGithubReleaseJob < ApplicationJob
     Settings.github.publish_wait.seconds.ago
   end
 
-  def check_version_status!
+  def check_version_status! # rubocop:disable Metrics/AbcSize
     if version_status.open? && !version_status.closeable?
-      update_release_status(status_details: 'version is open but not closeable')
+      github_release.update!(status_details: 'version is open but not closeable')
       Honeybadger.notify('Version is open but not closeable')
       return true
     end
 
     if !version_status.open? && !version_status.openable?
-      update_release_status(status_details: 'version is not openable')
+      github_release.update!(status_details: 'version is not openable')
       Honeybadger.notify('Version is not openable')
       return true
     end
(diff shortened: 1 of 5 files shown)


app/jobs/deposit_github_release_job.rb
```diff
@@ -20,7 +20,7 @@ def perform(github_release:, skip_publish_wait: false) # rubocop:disable Metrics
       return
     end
 
-    github_release.update!(status: 'started', status_details: nil)
+    update_release_status(status: 'started', status_details: nil)
```
peetucket: refactor the update into it's own method, so a fail can also rollback the state

app/jobs/deposit_github_release_job.rb
```diff
@@ -163,4 +163,13 @@ def download_assets
...
     end
   end
+
+  def update_release_status(status_details:, status: github_release.status)
+    github_release.update!(status_details:, status:)
+
+    # if the release failed, rollback the repository status so it can be tried again
+    return unless status == 'failed'
+
+    github_repository.deposit_fail! if github_repository.can_deposit_fail?
```
peetucket: new state transition is defined below

app/models/concerns/deposit_state_machine.rb
```diff
@@ -31,17 +38,18 @@ module DepositStateMachine
...
       before_transition deposit_registering_or_updating: :accessioning do |object|
         Notifier.publish(Notifier::ACCESSIONING_STARTED, object:, current_user: Current.user)
       end
 
-      before_transition accessioning: :deposit_not_in_progress do |object|
+      before_transition on: :accession_complete do |object|
         Notifier.publish(Notifier::ACCESSIONING_COMPLETE, object:)
       end
     end
+    # rubocop:enable Metrics/BlockLength
```
peetucket, line 53: we should only notify when the transition was due to a success mode, not a failure mode

app/jobs/deposit_github_release_job.rb
```diff
@@ -163,4 +163,13 @@ def download_assets
       end
     end
   end
+
+  def update_release_status(status_details:, status: github_release.status)
```
justinlittman: Why add this additional abstraction if there is only one place where the status is set to failed? I found it harder to follow.

app/models/concerns/deposit_state_machine.rb
```diff
@@ -16,6 +17,12 @@ module DepositStateMachine
         transition deposit_not_in_progress: :deposit_registering_or_updating
       end
 
+      # On deposit failure, clear in-progress states so a retry can be attempted.
+      event :deposit_fail do
```
justinlittman: Can this be called `:deposit_clear_fail`? It is called when you're clearing a failure not when the failure occurs.

app/models/concerns/deposit_state_machine.rb
```diff
@@ -4,6 +4,7 @@
 module DepositStateMachine
   extend ActiveSupport::Concern
 
+  # rubocop:disable Metrics/BlockLength
   included do
```
justinlittman, line 8: Changes in DepositStateMachine should be tested in `work_spec.rb`.

app/jobs/deposit_github_release_job.rb
peetucket: there are are actually two places where a fail can occur: `check_github_repository_deposit_state!` and the rescue clause in `perform`

app/models/concerns/deposit_state_machine.rb
peetucket: sure

app/jobs/deposit_github_release_job.rb
justinlittman: Then I'd recommend just handling the failures with a separate method, not all updates.

peetucket: done

app/models/concerns/deposit_state_machine.rb
peetucket: added new spec

spec/requests/contact_us_spec.rb
```diff
@@ -63,7 +63,7 @@
       it 'renders the form with an error and does not send email' do
         expect { post contacts_path, params: contact_params }.not_to change(ActionMailer::Base.deliveries, :count)
 
-        expect(response).to have_http_status(:unprocessable_entity)
+        expect(response).to have_http_status(:unprocessable_content)
```
peetucket, line 66: deprecation warning fix when running specs: `warning: Status code :unprocessable_entity is deprecated and will be removed in a future version of Rack. Please use :unprocessable_content instead.`

justinlittman (approved):

PR #2343, opened 2026-05-05
Disable save and deposit buttons while files are uploading.

### LINKED ISSUE #2342: It should not be possible to start a deposit while file uploads are still in progress
It's possible to press the deposit button before file uploads have been completed. This results in files never getting uploaded and a "missing files" error in accessioning.

To reproduce:

1. Fill out all required metadata fields
2. Go to the deposit tab
3. Start uploading files
4. Go back to the final tab for deposit
5. Press the deposit button before all files have been uploaded
6. See the error in accessioning - some files will be missing, it depends on how much was uploaded before the deposit started.

You can also reproduce this way:

1. Upload some files
2. Fill out all required metadata fields
3. Go back to upload more files
4. Go to deposit tab and press the deposit button without waiting for the files to finish uploading

**Expected behavior**

Either the form should prevent the deposit while an upload is in progress or accessioning should not begin until uploading is completed.

justinlittman:
closes #2342

### DIFF
--- a/app/components/works/edit/pane_component.html.erb
+++ b/app/components/works/edit/pane_component.html.erb
@@ -17,8 +17,8 @@
       </div>
       <% if draft_btns? %>
         <div>
-          <%= render Elements::Forms::SubmitComponent.new(form_id:, label: 'Save as draft', variant: :'outline-primary', classes: 'me-2', data: { action: 'unsaved-changes#allowFormSubmission' }) %>
-          <%= render Edit::DiscardDraftSubmitButtonComponent.new(presenter: work_presenter, form_id: discard_draft_form_id) %>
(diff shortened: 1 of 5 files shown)


mjgiarlo (approved):
Maybe a naïve question, but would we expect the unsaved-changes controller to handle this? Looking at https://github.com/sul-dlss/hungry-hungry-hippo/pull/1190 and the issues that closed.

Approving anyway

justinlittman:
The unsaved-changes controller will prevent you from leaving the page when there are unsaved changes UNLESS the user clicks save / deposit buttons. So this is sort of the opposite -- not allowing the user to click the save / deposit buttons.

PR #2351, opened 2026-05-12
Unify account lookup through AccountService

### LINKED ISSUE #2349: Handle participant entry when angle brackets are present
## Backtrace

    line 32 of [PROJECT_ROOT]/app/services/account_service.rb: block in AccountService#params
    line 31 of [PROJECT_ROOT]/app/services/account_service.rb: AccountService#params
    line 19 of [PROJECT_ROOT]/app/services/account_service.rb: AccountService#call

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/128495/faults/130466878)

mjgiarlo:
While working on the PR that closed #2349, I saw an opportunity to do some refactoring. Here's the refactoring I had in mind.

- Move account lookup responsibility out of AccountsController so search uses a single service path and avoids environment-specific branching in controller code.
- Update AccountService to build Account data from `Person#to_h`, return `nil` when attributes are blank, and choose a dev-only fake MaIS client to keep local behavior consistent with production call sites.
- Add `MaisPersonClient::Person#to_h` and a `FakeMaisPersonClient` so both real and fake responses share one attribute contract and reduce duplicated mapping logic.
- Align autocomplete and search handling on `id` (instead of `sunetid`) and use `blank?` checks in views/controllers to treat empty results consistently.
- Refresh service specs to assert returned account attributes from real XML parsing, reflecting the new normalization/mapping behavior.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -7,29 +7,12 @@ class AccountsController < ApplicationController
   def search
     authorize! with: AccountPolicy
 
-    @account = lookup_account
-    return head :not_found if @account.nil?
+    @account = AccountService.call(id: params[:id])
+    return head :not_found if @account.blank?
 
     respond_to do |format|
       format.html { render layout: false }
       format.json { render json: @account }
     end
-  end
-
-  private
-
-  def id
-    @id ||= params[:id] || params[:sunetid]
-  end
-
-  # Does a lookup from the account service in production mode, otherwise examines local database for users
-  def lookup_account
-    return AccountService.call(id:) unless Rails.env.development?
-
-    user = User.find_by_sunetid(sunetid: id) || User.find_by(email_address: id)
-    return nil unless user
-
-    AccountService::Account.new(name: user.name || user.sunetid, sunetid: user.sunetid,
-                                description: 'Digital Library Systems and Services')
   end
 end
--- a/app/services/account_service.rb
+++ b/app/services/account_service.rb
@@ -4,43 +4,71 @@
 class AccountService
   Account = Struct.new(:sunetid, :name, :description)
 
+  # Fake the MAIS Person API client. Used for development environments
+  class FakeMaisPersonClient
+    Person = Struct.new(:sunetid, :name, :description)
+
+    def self.fetch_user(id)
+      user = User.find_by_sunetid(sunetid: id) ||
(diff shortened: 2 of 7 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -7,29 +7,12 @@ class AccountsController < ApplicationController
   def search
     authorize! with: AccountPolicy
 
-    @account = lookup_account
-    return head :not_found if @account.nil?
+    @account = AccountService.call(id: params[:id])
```
mjgiarlo, line 10: Standardize on the ID param being named `id` instead of allowing both `id` and `sunetid`.

app/services/account_service.rb
```diff
@@ -8,39 +8,38 @@ def self.call(...)
...
   # @return [Account, nil] the account or nil if not found
   def call
-    return if id.blank?
-    return if params.blank?
+    return if id.blank? || account_attributes.blank?
 
     # Using the sunetid returned from the Account API since it will perform some normalization
     # e.g., removing accidental delimiter characters
-    Account.new(**params)
+    Account.new(**account_attributes)
```
mjgiarlo, line 45: Rename method for clarity.

config/initializers/mais_person_client.rb
```diff
@@ -6,3 +6,22 @@
   base_url: Settings.person_api.url,
   user_agent: 'Stanford Digital Repository'
 )
+
+# Overrides to the MaIS Person Client.
```
justinlittman: This would surprise me. Can it be put in `account_service.rb`? Or at the very least, a comment added to `account_service.rb` pointing at this?

lib/fake_mais_person_client.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Fake the MAIS Person API client. Used for development environments
```
justinlittman: Can this be a nested class of AccountService? I wouldn't look for this here.

mjgiarlo: Done.

config/initializers/mais_person_client.rb
mjgiarlo: Added comments.

justinlittman (approved):
Approve pending green.

PR #2358, opened 2026-05-20
Remove no break spaces from participants entry

### LINKED ISSUE #2356: [hungry-hungry-hippo/prod] URI::InvalidURIError: URI must be ascii only "/doc/person/kristiep\u00A0"
## Backtrace

    line 57 of [PROJECT_ROOT]/app/services/account_service.rb: AccountService#person
    line 53 of [PROJECT_ROOT]/app/services/account_service.rb: block in AccountService#account_attributes
    line 53 of [PROJECT_ROOT]/app/services/account_service.rb: AccountService#account_attributes

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/128495/faults/130639789)

lwrubel:
Fixes #2356. Splits on any kind of whitespace (including no-break spaces) and removes them. 

Tested on stage

### DIFF
--- a/app/javascript/controllers/participants_controller.js
+++ b/app/javascript/controllers/participants_controller.js
@@ -11,6 +11,7 @@ export default class extends Controller {
     if (this.inputTarget.value === '') return
 
     this.invalidFeedbackContainerTarget.innerHTML = ''
+    // Split on comma, semicolon, pipe, or whitespace (including any no-break spaces), and remove leading/trailing empty strings
     const ids = this.inputTarget.value.split(/[,;|\s]+/).filter(Boolean)
     await Promise.all(ids.map(id => this.lookupId(id)))
       .then((errorIds) => {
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/participants_controller.js
```diff
@@ -11,7 +11,7 @@ export default class extends Controller {
     if (this.inputTarget.value === '') return
 
     this.invalidFeedbackContainerTarget.innerHTML = ''
-    const ids = this.inputTarget.value.split(/[,;| ]+|\n/)
+    const ids = this.inputTarget.value.split(/[,;|\s]+/).filter(Boolean)
```
justinlittman, line 15: Can you use `.reject(&:blank?)` instead of `.filter(Boolean`)? I'd suggest that the filter clause is harder to understand.

lwrubel: This is in a Stimulus controller. Is there a better JS equivalent? We use `.filter(Boolean)` in subsequent lines in this controller.

justinlittman: Ooops. Can you add a comment indicating what it is doing?

justinlittman (approved):

app/javascript/controllers/participants_controller.js
lwrubel: No prob, added a comment.

PR #2393, opened 2026-07-24
structured value mapping changes

### LINKED ISSUE #2391: Update Cocina mappings to not create invalid structuredValues
cocina-models will be requiring that a `structuredValue` has more than one element. H3 mappings can currently create a structuredValue with only one element. These are affiliations where there is an institution and no department and formats with a main type and no subtype. 

@arcadiafalcone will provide examples of objects with Cocina that will be invalid with this requirement. 

The updated mapping for H3 will need to be implemented in coordination with remediation happening in Argo, to avoid any objects becoming un-roundtrippable. HOLD the PR for this work.

peetucket:
Fixes #2391 - change how structured value maps into cocina

Note: this should be held until existing objects are remediated to prevent round trip errors

Codex generated code and tests

### DIFF
--- a/app/cocina_builders/contributor_affiliation_cocina_builder.rb
+++ b/app/cocina_builders/contributor_affiliation_cocina_builder.rb
@@ -15,11 +15,13 @@ class ContributorAffiliationCocinaBuilder
   attr_reader :department, :institution, :uri
 
   def call
+    return institution_params if department.blank?
+
     {
       structuredValue: [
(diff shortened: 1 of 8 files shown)


justinlittman (approved):

peetucket:
per https://github.com/sul-dlss/dor-services-app/issues/6047#issuecomment-5108159179, this is mergable

PR #2403, opened 2026-08-05
Get object_updated_at from events if possible

### LINKED ISSUE #2398: Last modified date in table updates to "July 23, 2026" when item is viewed.
This happening for objects in a test collection on stage as well as to items in the production system.

To reproduce:

1. Go to a deposit list for a collection
2. Click on the title of an item to view it
3. Return to the deposit list by clicking on the collection name in the breadcrumbs OR by clicking the browser back button 
4. Refresh the page
5. The modified date for the recently viewed item will update to "July 23, 2026."

The updated dates also display on the tables on the dashboard and appear to be the dates being used to list items in the Administrator's recent activity table, as they now appear at the top (though this table does not display dates).

https://github.com/user-attachments/assets/828bdb80-2fc0-4a2d-a778-ecc8bfe2e7a1

aaron-collier:
Fixes #2398

### DIFF
--- a/app/model_synchronizers/work_model_synchronizer.rb
+++ b/app/model_synchronizers/work_model_synchronizer.rb
@@ -35,8 +35,17 @@ class WorkModelSynchronizer
       collection:,
       version:
     }.tap do |params|
-      params[:object_updated_at] = cocina_object.try(:modified)
+      params[:object_updated_at] = most_recent_event_timestamp || cocina_object.try(:modified)
     end.compact
+  end
+
+  def most_recent_event_timestamp
+    return nil unless work.druid
+
+    events = Sdr::Event.list(druid: work.druid)
+    return nil if events.blank?
+
+    events.first.timestamp
   end
 
   def title
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -116,7 +116,7 @@ GEM
       capistrano (~> 3.1)
     capistrano-maintenance (1.2.1)
       capistrano (>= 3.0)
-    capistrano-one_time_key (0.2.0)
+    capistrano-one_time_key (0.3.0)
       capistrano (~> 3.0)
     capistrano-passenger (0.2.1)
       capistrano (~> 3.0)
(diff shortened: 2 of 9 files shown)


app/model_synchronizers/work_model_synchronizer.rb
```diff
@@ -35,10 +35,19 @@ def update_params
...
     }.tap do |params|
-      params[:object_updated_at] = cocina_object.try(:modified)
+      params[:object_updated_at] = most_recent_event_timestamp || cocina_object.try(:modified)
     end.compact
   end
 
+  def most_recent_event_timestamp
+    return nil unless work.druid
+
+    events = Sdr::Event.list(druid: work.druid)
```
mjgiarlo, line 45: If the druid here is not found by DSA, it looks like `Sdr::Event#list` will return `nil`. Should the check on line 46 be for `#blank?` just in case? Or do we want the exception here if we're dealing with an unknown druid?

app/model_synchronizers/work_model_synchronizer.rb
```diff
@@ -35,10 +35,19 @@ def update_params
...
+    return nil if events.empty?
+
+    events.map { |event| Time.zone.parse(event.timestamp) }.max
```
mjgiarlo: Looking at the associated controller in DSA, it appears events are returning in timestamp (descending) order. Do you think we could safely use `events.first.timestamp` here instead of calling two iterator methods?

app/model_synchronizers/work_model_synchronizer.rb
aaron-collier: Ah, good catch, I think `blank?` is probably the correct approach here, updating.

app/model_synchronizers/work_model_synchronizer.rb
aaron-collier: I would be OK with that too, this just ensures that we don't have to rely on that being the case in the event that it changes.

justinlittman: Agree with avoiding the unnecessary sorting: https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/events_controller.rb#L6

Other things in this app will break if the order changes, so not too worried.

aaron-collier: Done.

mjgiarlo (approved):

PR #2408, opened 2026-08-07
Rake task for calling WorkModelSynchronizer on all non-draft works

aaron-collier:
Basic script used to resynch work data with cocina win necessary.

### DIFF
--- a/lib/tasks/resync.rake
+++ b/lib/tasks/resync.rake
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 namespace :resync do
-  desc 'Synchronize all with their cocina objects'
+  desc 'Synchronize all works with their cocina objects'
   task works: :environment do
     Work.where.not(druid: nil).find_each do |work|
       cocina_object = Sdr::Repository.find(druid: work.druid)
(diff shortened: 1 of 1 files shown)


lib/tasks/rsynch.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+namespace :resynch do
```
justinlittman: `resync`

lib/tasks/rsynch.rake
```diff
@@ -0,0 +1,17 @@
...
+  desc 'Synchronize all non-draft works with their cocina objects'
+  task works: :environment do
+    Work.where.not(druid: nil).find_each do |work|
+      version_status = Sdr::Repository.status(druid: work.druid)
+      next if version_status.draft?
```
justinlittman: Why non-draft?

aaron-collier: It may not matter, for some reason I thought that came up in discussion... I can drop that.

lib/tasks/resync.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :resync do
+  desc 'Synchronize all with their cocina objects'
```
mjgiarlo:
```suggestion
  desc 'Synchronize all works with their cocina objects'
```

justinlittman (approved):

mjgiarlo (approved):

